PR #91, opened 2019-04-04
New BC table type for H diffusion

RemiTheWarrior:
Implemented a new type of boundary condition named `"table"`. Here's how to use it:

In the parameters dict:

```python
"boundary_conditions": [
            {
                "surface": 1,
                "value":
                        [
                            [0, 0],
                            [5, 2],
                            [8, 6],
                            [9, 4],
                            [12, 18],
                            [11, 18]
                        ],
                "component": 0,
                "type": "dc"
            },
            {
                "surface": 2,
                "type": "table",
                "value":
                        [
                            [0, 1],
                            [1, 2],
                            [3, 0],
                            [5, 4],
                            [8, 6],
                            [10, 6]
                        ],
            }
    ]
```
The key `"value"`  can only be a 2D table and data is set as `[{time}, {value}]`.

Minor changes:
- Raise an error if type `key` is unknown

### DIFF
--- a/Main/FESTIM.py
+++ b/Main/FESTIM.py
@@ -1,11 +1,22 @@
 from fenics import *
 from dolfin import *
 import numpy as np
+from scipy.interpolate import interp1d
 import sympy as sp
 import csv
 import sys
 import os
 import argparse
+
+
+class ExpressionFromInterpolatedData(UserExpression):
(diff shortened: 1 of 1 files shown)


RemiTheWarrior:
@phugoide with this you'll be able to set BCs from experimental data for instance

Issue #493, opened 2022-06-29
Time display in `Simulation` should adapt to the log level


RemDelaporteMathurin:
When the loglevel attribute of a simulation is 20 (INFO), the time is not displayed properly.

![image](https://user-images.githubusercontent.com/40028739/176479250-1637d0aa-c9d7-4779-9631-b7d9c7997d63.png)

It could be easily fix by modifiying this logic:

https://github.com/RemDelaporteMathurin/FESTIM/blob/2793e7ab8efa6d1928c7106a88791c416bb3d304/FESTIM/generic_simulation.py#L347-L350

Something like:
```python
 if self.t != self.settings.final_time and self.loglevel == 40: 
     print(msg, end="\r") 
 else: 
     print(msg)
```

PR #524, opened 2022-09-28
Adaptation of time display according to log level

Lili-Gigi:
To answer to issue #493, here is my solution

### DIFF
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -343,7 +343,7 @@ class Simulation:
         msg = "{:.1f} %        ".format(simulation_percentage)
         msg += "{:.1e} s".format(simulation_time)
         msg += "    Ellapsed time so far: {:.1f} s".format(elapsed_time)
-        if self.t != self.settings.final_time and self.loglevel == 40:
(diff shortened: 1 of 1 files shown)


festim/generic_simulation.py
```diff
@@ -343,7 +343,7 @@ def display_time(self):
         msg = "{:.1f} %        ".format(simulation_percentage)
         msg += "{:.1e} s".format(simulation_time)
         msg += "    Ellapsed time so far: {:.1f} s".format(elapsed_time)
-        if self.t != self.settings.final_time:
+        if self.t != self.settings.final_time and self.loglevel == 40:
```
RemDelaporteMathurin, line 346: The tests are failing because there's a typo here:
```suggestion
        if self.t != self.settings.final_time and self.log_level == 40:
```

RemDelaporteMathurin (changes requested):
Thank you so much @Lili-Gigi for contributing to FESTIM! I think the tests fail because of this typo.

RemDelaporteMathurin:
Tested it locally and can attest it fixes #493 !

Thanks @Lili-Gigi for contributing!

PR #525, opened 2022-09-28
Solver parameter options for extrinsic traps

jhdark:
## Proposed changes

When using a temporal evolution of trap density for extrinsic traps, issues can be presented in the solver for the extrinsic traps as the default absolute and relative tolerances can be too strict. This change allows users to access these parameters to adjust them as necessary.

an example would look like this:

```python
my_trap = F.ExtrinsicTrap(
    k_0=1,
    E_k=1,
    p_0=1,
    E_p=1,
    materials=1,
    phi_0=1,
    n_amax=1,
    n_bmax=1,
    eta_a=1,
    eta_b=1,
    f_a=1,
    f_b=1,
    absolute_tolerance=1,
    relative_tolerance=2,
    maximum_iterations=1,
    linear_solver="mumps")
```

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments


festim/concentration/traps/extrinsic_trap.py
```diff
@@ -48,6 +66,10 @@ def __init__(
         eta_b,
         f_a,
         f_b,
+        absolute_tolerance=1e0,
+        relative_tolerance=1e-10,
+        maximum_iterations=30,
+        linear_solver=None,
```
RemDelaporteMathurin, line 72:
```suggestion
        **kwargs,
```

festim/concentration/traps/extrinsic_trap.py
```diff
@@ -57,6 +79,10 @@ def __init__(
             p_0,
             E_p,
             materials,
+            absolute_tolerance,
+            relative_tolerance,
+            maximum_iterations,
+            linear_solver,
```
RemDelaporteMathurin, line 85:
```suggestion
            **kwargs,
```

festim/concentration/traps/neutron_induced_trap.py
```diff
@@ -28,13 +28,34 @@ class NeutronInducedTrap(ExtrinsicTrapBase):
...
+        phi,
+        K,
+        n_max,
+        A_0,
+        E_A,
+        absolute_tolerance=1e0,
+        relative_tolerance=1e-10,
+        maximum_iterations=30,
+        linear_solver=None,
+        id=None,
```
RemDelaporteMathurin, line 47:
```suggestion
        id=None,
        **kwargs
```

festim/concentration/traps/neutron_induced_trap.py
```diff
@@ -28,13 +28,34 @@ class NeutronInducedTrap(ExtrinsicTrapBase):
...
         super().__init__(
             k_0,
             E_k,
             p_0,
             E_p,
             materials,
+            absolute_tolerance,
+            relative_tolerance,
+            maximum_iterations,
+            linear_solver,
```
RemDelaporteMathurin, line 58:
```suggestion
```

RemDelaporteMathurin: In addition to this, `**kwargs` should be added at the end of the arguments definition.

festim/concentration/traps/traps.py
```diff
@@ -72,7 +72,16 @@ def define_variational_problem_extrinsic_traps(self, dx, dt, T):
     def solve_extrinsic_traps(self):
         for trap in self.traps:
             if isinstance(trap, festim.ExtrinsicTrapBase):
-                f.solve(trap.form_density == 0, trap.density[0], [])
+                du_t = f.TrialFunction(trap.density[0].function_space())
+                J_t = f.derivative(trap.form_density, trap.density[0], du_t)
+                problem = f.NonlinearVariationalProblem(
+                    trap.form_density, trap.density[0], [], J_t)
```
RemDelaporteMathurin, line 78: I wonder if the Jacobian needs to be explicitly defined here. Wouldn't `problem = f.NonlinearVariationalProblem(trap.form_density, trap.density[0], [])` be enough?

RemDelaporteMathurin (changes requested):
Thanks for this PR. It adds flexibility on the ExtrinsicTrap solvers  which weren't exposed to users before!

I think we can refactore some of it by using `**kwargs` since these new parameters need only to be defined at the `ExtrinsicTrapBase` level.

Also, can you add something in the docstrings for documentation?

festim/concentration/traps/traps.py
jhdark: Yeah thats true its not likely needed, but I'll do some quick testing to see if it makes a difference

RemDelaporteMathurin: I think the default behaviour of fenics is to do exactly what lines 75-76 do. So it would just be saving 2 lines :-)

festim/concentration/traps/extrinsic_trap.py
```diff
@@ -48,6 +66,7 @@ def __init__(
         eta_b,
         f_a,
         f_b,
+        **kwargs,
```
RemDelaporteMathurin, line 69: **kwargs should be the last in the definition list

festim/concentration/traps/extrinsic_trap.py
```diff
@@ -57,6 +76,7 @@ def __init__(
             p_0,
             E_p,
             materials,
+            **kwargs,
```
RemDelaporteMathurin, line 79: **kwargs should be the last in the definition list

RemDelaporteMathurin (approved):
Thanks for this added flexibility!

PR #527, opened 2022-10-08
moving to newer setuptools_scm

shimwell:
## Proposed changes
Just noticed you are using setuptools_scm_git_archive which is now included in setuptools_scm version 7 and above so I thought you might be interesting in removing a build dependency

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# version file
+**_version.py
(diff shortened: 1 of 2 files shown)


RemDelaporteMathurin:
Thanks @shimwell for keeping a eye out for these things! I must admit I don't tend to update the python packaging scripts that often.

I tested a local pip install with this new version and can confirm it works! merging this now!

PR #529, opened 2022-10-10
moved to src package strut

shimwell:
## Proposed changes

Just wondering if you are interested in moving to a src layout which has become the  [recommended](https://packaging.python.org/en/latest/tutorials/packaging-projects/) layout
```
├─ src
│  └─ festim
│     ├─ __init__.py
│     └─ ...
├─ tests
│  └─ ...
└─ pyproject.toml
```

There are several reasons which are well described elsewhere e.g. [[1]](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure), 

I've been updating a few packages to use this structure but would be keen to see what you think @RemDelaporteMathurin 

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted


RemDelaporteMathurin:
@shimwell thanks for the suggestion. I am not very strongly oppiniated on the matter tbh.
One of the main advantages would be that it "forces" the users to install the package in order to run the tests which is good (while being inconvenient! 😃 )

I wonder how we could fix the tests to make it work with this new structure.

shimwell:
The tests are failing because the CI is testing the folder and not installing the package.

This is part of the reason for moving to the src folder layout.

Adding a ```pip install .``` to the ci scripts should fix that.

RemDelaporteMathurin:
> There could be parts of the folder structure that don't get packaged up when making the package, so testing the folder can be a bit risky

I think I'm having this issue with [HTM](https://github.com/RemDelaporteMathurin/h-transport-materials) as the package actually contains data like csv files. I would be happy to have your opinion on this too

pyproject.toml
```diff
@@ -1,10 +1,44 @@
...
+]
+license = {file = "LICENSE"}
+description = "Finite element simulations of hydrogen transport"
+readme = "README.md"
+requires-python = ">=3.7"
+classifiers = [
+    "Natural Language :: English",
+    "Topic :: Scientific/Engineering",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.6",
```
RemDelaporteMathurin, line 19: Above there's `requires-python = ">=3.7"`

shimwell: ah yes sorry about this, it might have to raise the minimal required python a few versions

shimwell:
ah I see that the docker image used has an older version of pip (19.3.1) and python (3.6) installed. Some of these more modern packaging methods need newer versions of python and newer versions of pip.

I tried upgrading pip in the CI but it would only go up to 21.3.1 when the current latest version is 22.2.2, I guess this is related to python being the now depreciated 3.6 on the image.

Might need a bit of an update to the docker image to get this working

RemDelaporteMathurin:
Ah yes this old pip is a bit of a pain. This FEniCS docker image isn't updated anymore....

> Might need a bit of an update to the docker image to get this working

There is a conda install for FEniCS. Maybe we should use it instead...

shimwell:
Tried switching the ci to conda but I think more might be needed to install fenics

RemDelaporteMathurin:
That's odd. It seems the C++ part of fenics is not installed. Are we sure the tests are using the conda env?

RemDelaporteMathurin:
Closing this for now. Feel free to reopen

PR #549, opened 2023-07-19
Smaller README

RemDelaporteMathurin:
* Take 2 * 

@jhdark what do you think about having the readme even shorter?


jhdark (approved):
I guess this way we remove all duplication of code

Issue #513, opened 2022-08-11
Feature request: class for DissociationFlux


RemDelaporteMathurin:
We already have a convenient class for `RecombinationFlux.`

We could also use a class for `DissociationFlux.`

The structure would be very similar to the [recombination flux](https://github.com/RemDelaporteMathurin/FESTIM/blob/main/festim/boundary_conditions/fluxes/recombination_flux.py).

The expression of the flux would be:

$$ \mathrm{flux} = K_\mathrm{diss} P $$

where $P$ is the partial pressure of H in Pa, and $K_\mathrm{diss}$ is the dissociation coefficient

PR #557, opened 2023-07-26
New feature: Dissociation flux

jhdark:
## Proposed changes

Fix for issue  #513 

This new feature would allow for a dissociation flux BC

The expression of the flux would be:

$$ \mathrm{flux} = K_\mathrm{diss} P $$

where $P$ is the partial pressure of H in Pa, and $K_\mathrm{diss}$ is the dissociation coefficient

@RemDelaporteMathurin  do we need to integrate this into the ImplantationDirichlet class? Amending the equation to:

$$ c = \frac{\phi \cdot R_{p}}{D} + \sqrt{\frac{\phi + K_\mathrm{diss}  P}{Kr}}  $$

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments


RemDelaporteMathurin:
> do we need to integrate this into the ImplantationDirichlet class? Amending the equation to:

Your equation here is not homogeneous. Check out the units.

festim/boundary_conditions/fluxes/dissociation_flux.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class DissociationFlux(FluxBC):
+    """
+    FluxBC subclass for hydrogen dissociation flux.
+    -D(T) * grad(c) * n = -Kd(T) * P
+
+    Args:
+        Kd_0 (float or sp.Expr): dissociation coefficient pre-exponential
+            factor (m^(2)/s)
```
RemDelaporteMathurin, line 13: Wrong units

festim/boundary_conditions/fluxes/dissociation_flux.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def __init__(self, Kd_0, E_Kd, P, surfaces) -> None:
+        self.Kd_0 = Kd_0
+        self.E_Kd = E_Kd
+        self.P = P
+        super().__init__(surfaces=surfaces, field=0)
+
+    def create_form(self, T):
+        Kd_0_expr = f.Expression(sp.printing.ccode(self.Kd_0), t=0, degree=1)
+        E_Kd_expr = f.Expression(sp.printing.ccode(self.E_Kd), t=0, degree=1)
```
RemDelaporteMathurin, line 28: Note:

having Expression objects for everything comes at a cost. these objects are more costly than Constant() objects for instance.
Especially since we keep "updating" them at each timestep even though they don't evolve with time.
In the future, it'd be nice to have some kind of tests to check: is it a constant? is it only a function of space? is it only a function of time? and use the most appropriate fenics object then.

Does that make sense?

festim/boundary_conditions/fluxes/dissociation_flux.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    FluxBC subclass for hydrogen dissociation flux.
+    -D(T) * grad(c) * n = -Kd(T) * P
+
+    Args:
+        Kd_0 (float or sp.Expr): dissociation coefficient pre-exponential
+            factor (m^(2)/s)
+        E_Kd (float or sp.Expr): dissociation coefficient activation
+            energy (eV)
+        P (float): partial pressure of H (Pa)
```
RemDelaporteMathurin, line 16: It is way more likely that users will want to have a time dependent pressure than a time dependent `Kd_0` don't you think?

`P` should also accept a `sp.Expr` object.

RemDelaporteMathurin (changes requested):

festim/boundary_conditions/fluxes/dissociation_flux.py
jhdark: Yeah I get what you mean, but would we need to make this test in every class that creates a form?

festim/boundary_conditions/fluxes/dissociation_flux.py
jhdark: m-2 s-1 Pa-1?

festim/boundary_conditions/fluxes/dissociation_flux.py
RemDelaporteMathurin: yes

festim/boundary_conditions/fluxes/dissociation_flux.py
```diff
@@ -0,0 +1,33 @@
...
+        super().__init__(surfaces=surfaces, field=0)
+
+    def create_form(self, T):
+        Kd_0_expr = f.Expression(sp.printing.ccode(self.Kd_0), t=0, degree=1)
+        E_Kd_expr = f.Expression(sp.printing.ccode(self.E_Kd), t=0, degree=1)
+        P_expr = f.Expression(sp.printing.ccode(self.P), t=0, degree=1)
+
+        Kd = Kd_0_expr * f.exp(-E_Kd_expr / k_B / T)
+        self.form = -Kd * P_expr
+        self.sub_expressions = [Kd_0_expr, E_Kd_expr]
```
RemDelaporteMathurin, line 33: If you don't add `P_expr` to the sub_expressions then they won't get updated in time 
```suggestion
        self.sub_expressions = [Kd_0_expr, E_Kd_expr, P_expr]
```

RemDelaporteMathurin (changes requested):

festim/boundary_conditions/dirichlets/dc_imp.py
```diff
@@ -57,6 +82,9 @@ def create_expression(self, T):
             E_D=self.E_D,
             Kr_0=self.Kr_0,
             E_Kr=self.E_Kr,
+            Kd_0=self.Kd_0,
+            E_Kd=self.E_Kd,
+            P=self.P,
```
RemDelaporteMathurin, line 87: Should pressure be an Expression object?

festim/boundary_conditions/fluxes/dissociation_flux.py
```diff
@@ -0,0 +1,33 @@
+from festim import FluxBC, k_B
+import fenics as f
+import sympy as sp
+
+
+class DissociationFlux(FluxBC):
+    """
+    FluxBC subclass for hydrogen dissociation flux.
+    -D(T) * grad(c) * n = -Kd(T) * P
```
RemDelaporteMathurin, line 9:
```suggestion
    -D(T) * grad(c) * n = Kd(T) * P
```

festim/boundary_conditions/fluxes/dissociation_flux.py
```diff
@@ -0,0 +1,33 @@
...
+        self.P = P
+        super().__init__(surfaces=surfaces, field=0)
+
+    def create_form(self, T):
+        Kd_0_expr = f.Expression(sp.printing.ccode(self.Kd_0), t=0, degree=1)
+        E_Kd_expr = f.Expression(sp.printing.ccode(self.E_Kd), t=0, degree=1)
+        P_expr = f.Expression(sp.printing.ccode(self.P), t=0, degree=1)
+
+        Kd = Kd_0_expr * f.exp(-E_Kd_expr / k_B / T)
+        self.form = -Kd * P_expr
```
RemDelaporteMathurin, line 32:
```suggestion
        self.form = Kd * P_expr
```

Be careful with the sign

festim/boundary_conditions/dirichlets/dc_imp.py
jhdark: It should be indeed

festim/boundary_conditions/dirichlets/dc_imp.py
```diff
@@ -32,21 +36,47 @@ class ImplantationDirichlet(DirichletBC):
...
+        self.E_Kd = E_Kd
+        self.P = P
 
     def create_expression(self, T):
         phi = f.Expression(sp.printing.ccode(self.phi), t=0, degree=1)
         R_p = f.Expression(sp.printing.ccode(self.R_p), t=0, degree=1)
-        sub_expressions = [phi, R_p]
+        if self.P is not None:
+            P = f.Expression(sp.printing.ccode(self.P), t=0, degree=1)
+            sub_expressions = [phi, R_p, P]
```
RemDelaporteMathurin, line 76: What you can do here instead of creating the sub_expressions list in both cases is to first create it with phi and Rp outside of the if statement and then if P is not None, append P to the list.

RemDelaporteMathurin (changes requested):

festim/boundary_conditions/dirichlets/dc_imp.py
jhdark: True

RemDelaporteMathurin:
@ehodille what do you think of the formulation for c as a function of implantation flux and partial pressure?

RemDelaporteMathurin:
@jhdark I believe [this section](https://festim.readthedocs.io/en/latest/userguide/boundary_conditions.html#plasma-implantation-approximation) could be modified

RemDelaporteMathurin:
No Sievert's law doesn't apply here as it implies the diffusion is much slower than the surface fluxes. You have to have the recombination and dissociation coefficients for handling a pressure 

If phi=0 then you obtain Sievert's law

docs/source/userguide/boundary_conditions.rst
```diff
@@ -120,9 +131,9 @@ A plasma implantation can be approximated by:
 
 .. math::
 
-    c_\mathrm{m} = \frac{\varphi_\mathrm{imp} \, R_p}{D} + \sqrt{\frac{\varphi_\mathrm{imp}}{Kr}}
+    c_\mathrm{m} = \frac{\varphi_\mathrm{imp} \, R_p}{D} + \sqrt{\frac{\varphi_\mathrm{imp} + Kd \cdot P}{Kr}}
 
-Where :math:`\varphi_\mathrm{imp}` is the implanted flux, :math:`R_p` is the implantation depth, :math:`D` is the diffusion coefficient and :math:`Kr` is the recombination coefficient.
+Where :math:`\varphi_\mathrm{imp}` is the implanted flux, :math:`R_p` is the implantation depth, :math:`D` is the diffusion coefficient, :math:`Kd` is the dissociation coefficient and :math:`Kr` is the recombination coefficient.
```
RemDelaporteMathurin, line 136:
```suggestion
Where :math:`\varphi_\mathrm{imp}` is the implanted flux, :math:`R_p` is the implantation depth, :math:`D` is the diffusion coefficient, :math:`Kd` is the dissociation coefficient, :math:`P` is the partial pressure of hydrogen,  and :math:`Kr` is the recombination coefficient.

When :math:`P = 0`:

.. math::
    c_\mathrm{m} = \frac{\varphi_\mathrm{imp} \, R_p}{D} + \sqrt{\frac{\varphi_\mathrm{imp}}{Kr}}

Moreover, assuming an instantaneous recombination :math:`Kr = \infty`:

.. math::
    c_\mathrm{m} = \frac{\varphi_\mathrm{imp} \, R_p}{D}
```

Issue #563, opened 2023-08-09
[BUG] TrapDensityXDMF projects value over whole domain


jhdark:
**Describe the bug**
Both @ehodille  and I ran into this issue.
When using the TrapDensityXDMF export, the correct density is exported, however, it is projected over the whole domain, thus in a multi-material system this doesn't work.

An example can be seen here where a 2mm domain is comprised of two materials with a trap each, but for both materials, the trap density extends over the domain.

![trap_densities](https://github.com/RemDelaporteMathurin/FESTIM/assets/65899899/96265ca4-d1c0-416b-a37f-f330d5eefb49)

**To Reproduce**
```python
import numpy as np
import festim as F
import fenics as f

my_model = F.Simulation(log_level=20)
vertices = np.linspace(0, 2e-06, num=100)
my_model.mesh = F.MeshFromVertices(vertices=vertices)
mat_1 = F.Material(
    id=1,
    D_0=1,
    E_D=0,
    S_0=1,
    E_S=0,
    borders=[0, 1e-06]
)
mat_2 = F.Material(
    id=2,
    D_0=1,
    E_D=0,
    S_0=1,
    E_S=0,
    borders=[1e-06, 2e-06]
)
my_model.materials = F.Materials([mat_1, mat_2])
trap_mat_1 = F.Trap(
    k_0=1,
    E_k=0,
    p_0=1,
    E_p=0,
    density=15,
    materials=mat_1,
)
trap_mat_2 = F.Trap(
    k_0=1,
    E_k=0,
    p_0=1,
    E_p=0,
    density=10,
    materials=mat_2,
)
my_model.traps = F.Traps(
    [
        trap_mat_1,
        trap_mat_2,
    ]
)
my_model.T = F.Temperature(value=100)
my_model.boundary_conditions = [
    F.DirichletBC(surfaces=1, value=1, field=0),
    F.DirichletBC(surfaces=2, value=0, field=0),
]
folder_results = "Results/mwe/"
my_model.exports = F.Exports(
    [
        F.TrapDensityXDMF(trap_mat_1, folder=folder_results, label="trap_density_mat_1", checkpoint=False),
        F.TrapDensityXDMF(trap_mat_2, folder=folder_results, label="trap_density_mat_2", checkpoint=False),
    ]
)
my_model.settings = F.Settings(
    transient=False,
    absolute_tolerance=1e10,
    relative_tolerance=1e-10,
)
my_model.initialise()
my_model.run()
```

[...]

PR #564, opened 2023-08-10
Trap density xdmf fix

jhdark:
## Proposed changes

fix for issue #563 

This would allow users to export the TrapDensityXDMF in a multi-material system 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/exports/exports.py
+++ b/festim/exports/exports.py
@@ -14,7 +14,7 @@ class Exports:
         """writes to file
 
         Args:
-            label_to_function (str, int): label of function
+            label_to_function (dict): dictionary of labels mapped to solutions
             dt (festim.Stepsize): the model's stepsize
             dx (fenics.Measure): the measure for dx
         """
--- a/festim/exports/exports.py
+++ b/festim/exports/exports.py
@@ -10,7 +10,7 @@ class Exports:
         self.final_time = None
         self.nb_iterations = 0
 
-    def write(self, label_to_function, dt, dx, log_level):
+    def write(self, label_to_function, dt, dx):
         """writes to file
 
         Args:
@@ -49,7 +49,7 @@ class Exports:
                             )
                     export.function = label_to_function[export.field]
                     if isinstance(export, festim.TrapDensityXDMF):
-                        export.write(self.t, dx, log_level)
+                        export.write(self.t, dx)
                     else:
                         export.write(self.t)
                     export.append = True
--- a/festim/exports/exports.py
+++ b/festim/exports/exports.py
@@ -78,4 +78,3 @@ class Exports:
                 export.data = [export.make_header()]
                 export.assign_measures_to_quantities(dx, ds)
                 export.assign_properties_to_quantities(materials)
-
--- a/festim/exports/exports.py
+++ b/festim/exports/exports.py
@@ -11,6 +11,13 @@ class Exports:
         self.nb_iterations = 0
 
     def write(self, label_to_function, dt, dx):
+        """writes to file
+
+        Args:
+            label_to_function (str, int): label of function
+            dt (festim.Stepsize): the model's stepsize
+            dx (fenics.Measure): the measure for dx
+        """
         for export in self.exports:
             if isinstance(export, festim.DerivedQuantities):
                 # compute derived quantities
--- a/festim/exports/exports.py
+++ b/festim/exports/exports.py
@@ -10,7 +10,7 @@ class Exports:
         self.final_time = None
         self.nb_iterations = 0
 
-    def write(self, label_to_function, dt):
+    def write(self, label_to_function, dt, dx):
         for export in self.exports:
             if isinstance(export, festim.DerivedQuantities):
                 # compute derived quantities
@@ -41,7 +41,10 @@ class Exports:
                                 label_to_function["retention"], self.V_DG1
                             )
                     export.function = label_to_function[export.field]
-                    export.write(self.t)
+                    if isinstance(export, festim.TrapDensityXDMF):
+                        export.write(self.t, dx)
+                    else:
+                        export.write(self.t)
                     export.append = True
 
             elif isinstance(export, festim.TXTExport):
@@ -68,3 +71,4 @@ class Exports:
                 export.data = [export.make_header()]
                 export.assign_measures_to_quantities(dx, ds)
                 export.assign_properties_to_quantities(materials)
+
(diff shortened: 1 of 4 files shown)


festim/exports/exports.py
```diff
@@ -10,7 +10,7 @@ def __init__(self, exports=[]) -> None:
         self.final_time = None
         self.nb_iterations = 0
 
-    def write(self, label_to_function, dt):
+    def write(self, label_to_function, dt, dx):
```
RemDelaporteMathurin, line 13: Can we take the opportunity to add docstrings here?

festim/exports/trap_density_xdmf.py
```diff
@@ -16,13 +16,22 @@ def __init__(self, trap, **kwargs) -> None:
...
-        density_as_function = f.project(self.trap.density[0], functionspace)
-        self.function = density_as_function
+        u = f.Function(functionspace)
+        v = f.TestFunction(functionspace)   
+        F = f.inner(u, v) * dx
+
+        for mat in self.trap.materials:
+            F -= f.inner(self.trap.density[0], v) * dx(mat.id)
+        
+        f.solve(F == 0, u, bcs=[])
```
RemDelaporteMathurin, line 34: Could we have some log and print statements?
Like if the log level is below 40 then print "solving variational problem for trap density export".

festim/generic_simulation.py
```diff
@@ -363,7 +363,7 @@ def run_post_processing(self):
         self.update_post_processing_solutions()
 
         self.exports.t = self.t
-        self.exports.write(self.label_to_function, self.dt)
+        self.exports.write(self.label_to_function, self.dt, self.mesh.dx)
```
RemDelaporteMathurin, line 366: Should `mesh.dx` be an attribute of `exports` instead?

RemDelaporteMathurin (changes requested):

festim/generic_simulation.py
jhdark: This might be better as we might need a way to check in the TrapDensityXDMF the number of subdomains to make the variational formulation properly

RemDelaporteMathurin: I don't see how having the dx measurement an attribute of exports vs an argument of exports.write makes you more able to check the number of subdomains... this suggestion was more to refactor the code

festim/exports/trap_density_xdmf.py
```diff
@@ -16,13 +16,27 @@ def __init__(self, trap, **kwargs) -> None:
 
         self.trap = trap
 
-    def write(self, t):
+    def write(self, t, dx, log_level):
```
RemDelaporteMathurin, line 19: I wouldn't have it as an argument.

Imagine if we want to have this everywhere then you'd have to pass log_level literally everywhere.

Maybe let's forget about this for this PR.

What may be a solution is to have log_level an attribute of FESTIM.

Instead of:

```python
my_model.log_level = 20
```

we'd have

```python
import festim

festim.log_level = 20
```

That way, log_level can be accessed from anywhere.

festim/generic_simulation.py
```diff
@@ -363,7 +363,9 @@ def run_post_processing(self):
         self.update_post_processing_solutions()
 
         self.exports.t = self.t
-        self.exports.write(self.label_to_function, self.dt)
+        self.exports.write(
+            self.label_to_function, self.dt, self.mesh.dx, self.log_level
```
RemDelaporteMathurin, line 367: no log_level in the arguments

test/unit/test_exports/test_trap_density_xdmf_export.py
```diff
@@ -74,7 +78,7 @@ def test_trap_density_xdmf_export_write(tmpdir):
     my_export.function = Function(V_vector).sub(1)
 
     # run
-    my_export.write(t=1)
+    my_export.write(t=1, dx=dx, log_level=20)
```
RemDelaporteMathurin, line 81: no log_level

festim/exports/trap_density_xdmf.py
```diff
@@ -16,13 +16,27 @@ def __init__(self, trap, **kwargs) -> None:
...
         functionspace = self.function.function_space().collapse()
-        density_as_function = f.project(self.trap.density[0], functionspace)
-        self.function = density_as_function
+        u = f.Function(functionspace)
+        v = f.TestFunction(functionspace)
+        F = f.inner(u, v) * dx
+
+        for mat in self.trap.materials:
+            if isinstance(mat, str):
+                F -= f.inner(self.trap.density[0], v) * dx
```
RemDelaporteMathurin, line 33: This cannot work.

If you do this, then this is going back to the old behaviour.

Imagine a test case where the materials of a trap are [mat1, "mat_id2"] where mat1 is a festim.Material.

Then your formulation is something like

u = trap_density + trap_density in mat1 and trap_density in mat2

festim/exports/trap_density_xdmf.py
jhdark: This would be a nice solution, i'll remove it from the function

test/unit/test_exports/test_trap_density_xdmf_export.py
```diff
@@ -2,6 +2,7 @@
 from fenics import *
 import sympy as sp
 from pathlib import Path
+import numpy as np
 
 
 def test_trap_density_xdmf_export_intergration_with_simultion(tmpdir):
```
RemDelaporteMathurin, line 8: This test only tests the feature with a monomaterial so it doesn't prove the new feature works as expected.
I would create a 1D domain with 2 subdomains and put a trap only in one of the domains.

test/unit/test_exports/test_trap_density_xdmf_export.py
```diff
@@ -129,12 +130,16 @@ def test_trap_density_xdmf_export_traps_materials_mixed(tmpdir):
...
-    V = FunctionSpace(my_model.mesh.mesh, "CG", 1)
-
-    density_expected = interpolate(festim.as_expression(density_expr), V)
+    V = FunctionSpace(my_model.mesh.mesh, "DG", 1)
+    volume_markers = my_model.mesh.volume_markers
+    dx = Measure("dx", domain=my_model.mesh.mesh, subdomain_data=volume_markers)
+    density_expected = Function(V)
+    v = TestFunction(V)
+    F = inner(density_expected, v) * dx
+    F -= inner(festim.as_expression(density_expr), v) * dx(1)
```
RemDelaporteMathurin, line 139: no need for `as_expression` here
```suggestion
    F -= inner(density_expr, v) * dx(1)
```

PR #580, opened 2023-10-05
Tiny timestep

gabriele-ferrero:
## Proposed changes

When starting to solve a model, some initial stepsize can be required. Sometimes the adaptive timesteps becomes very small at the beginning of a simulation. In these cases, the message output in festim is 0.0 s. With this change, we can see tiny timesteps in the scientific format (eg 1E-9 s).
## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Black formatted
- [X ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -351,7 +351,6 @@ class Simulation:
     def display_time(self):
         """Displays the current time"""
         simulation_percentage = round(self.t / self.settings.final_time * 100, 2)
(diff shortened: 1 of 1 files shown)


RemDelaporteMathurin (approved):
Thank you very much @gabriele-ferrero for this fix!

PR #581, opened 2023-10-05
Fix dissociation BC for festim

gabriele-ferrero:
## Proposed changes

The create form function in DissociationFlux takes 2 arguments, but 3 were given, so it doesn't work properly when solving a model. by adding a solute argument, even if unused, the issue is solved. The test for dissociation flux

## Types of changes

- [X ] Bugfix (non-breaking change which fixes an issue)
- [ ] Code refactoring
- [ ] New tests

## Checklist

- [X ] Black formatted

### DIFF
--- a/test/unit/test_boundary_conditions.py
+++ b/test/unit/test_boundary_conditions.py
@@ -548,4 +548,4 @@ def test_dissoc_flux():
     T = fenics.Expression(sp.printing.ccode(expr), degree=1, t=0)
 
     my_BC = festim.DissociationFlux(surfaces=[0], Kd_0=expr, E_Kd=expr, P=1)
-    my_BC.create_form(T)
+    my_BC.create_form(T,None)
--- a/festim/boundary_conditions/fluxes/dissociation_flux.py
+++ b/festim/boundary_conditions/fluxes/dissociation_flux.py
@@ -23,7 +23,7 @@ class DissociationFlux(FluxBC):
         self.P = P
         super().__init__(surfaces=surfaces, field=0)
 
-    def create_form(self, T):
+    def create_form(self, T, solute):
         Kd_0_expr = f.Expression(sp.printing.ccode(self.Kd_0), t=0, degree=1)
         E_Kd_expr = f.Expression(sp.printing.ccode(self.E_Kd), t=0, degree=1)
         P_expr = f.Expression(sp.printing.ccode(self.P), t=0, degree=1)


RemDelaporteMathurin (changes requested):
Thanks @gabriele-ferrero 

Do you think you can add a small test that catches the buggy behaviour?

Something like


```python
import festim as F

sim = F.Simulation()

sim.boundary_conditions = [F.DissociationFlux(.....)]
sim.T = 500
sim.mesh = F.MeshFromVertices([0, 1, 2, 3])
sim.materials = F.Material(id=1, D_0=1, E_D=0)
sim.settings = F.Settings(transient=False, ...)
# maybe it needs other things too

sim.initialise()
sim.run()
```

gabriele-ferrero:
The new test fails with this error message "AttributeError: 'int' object has no attribute 'sources'". even when i define sim.sources=[] . Do you have any suggestion on why this is happening?

Test for reference: 
def test_dissociation_flux_solve():
    sim = festim.Simulation()
    sim.mesh = festim.MeshFromVertices([0, 1, 2, 3])
    sim.materials = festim.Material(id=1, D_0=1, E_D=0)
    sim.T = 500
    sim.boundary_conditions = [
        festim.DissociationFlux(surfaces=[1], Kd_0=1, E_Kd=0, P=1e4)
    ]
    sim.settings = festim.Settings(
        transient=False, absolute_tolerance=1e8, relative_tolerance=1e-8
    )
    sim.dt = None
    sim.exports = []
    sim.initialise()
    sim.run()

gabriele-ferrero:
Wait a second to merge. I want to write a better test

test/unit/test_boundary_conditions.py
```diff
@@ -543,9 +543,26 @@ def test_string_for_field_in_dirichletbc():
     bc.create_dirichletbc(V, fenics.Constant(1), surface_marker)
 
 
+def test_dissociation_flux_solve():
```
RemDelaporteMathurin, line 546:
```suggestion
def test_dissociation_flux_solve():
    """Test to catch bug DissociationFlux see #581
    """
```

test/unit/test_boundary_conditions.py
```diff
@@ -543,9 +543,28 @@ def test_string_for_field_in_dirichletbc():
...
+    """Test to catch bug DissociationFlux see #581
+    """
```
RemDelaporteMathurin, line 548:
```suggestion
    """Test to catch bug DissociationFlux see #581"""
```

RemDelaporteMathurin (approved):
Thanks for this @gabriele-ferrero ! 🎉

PR #584, opened 2023-10-05
HydrogenTransportProblem class, Mesh and 1Dmesh class

jhdark:
## Proposed changes

Proposing a new alternative to general simulation having a HydrogenTransportProblem class. Started work on meshing options having support for a 1D mesh. From this the function spaces and subdomains are defined. To ensure it works one can run the following mwe script:

```python
import festim as F
import numpy as np
from dolfinx.io import XDMFFile
from dolfinx import fem
from mpi4py import MPI

my_model = F.HydrogenTransportProblem()

my_mesh = F.Mesh1D(vertices=np.linspace(0, 1, num=100))

my_model.geometry = F.Geometry(mesh=my_mesh)

my_model.initialise()

u = fem.Function(my_model.geometry.function_space)
u.interpolate(lambda x: 1 + x[0] ** 2)

test_xdmf = XDMFFile(MPI.COMM_WORLD, "test_output.xdmf", "w")
test_xdmf.write_mesh(my_model.geometry.mesh.mesh)
test_xdmf.write_function(u)
test_xdmf.close()
```

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments


festim/geometry/mesh.py
```diff
@@ -0,0 +1,57 @@
...
+        ds (dolfinx.Measure):
+    """
+
+    def __init__(
+        self,
+        mesh=None,
+        volume_markers=None,
+        surface_markers=None,
+        subdomains=[],
+        **kwargs
```
RemDelaporteMathurin, line 21: What are the `kwargs` for here? doesn't seem to be used anywhere else in the `__init__`

festim/geometry/mesh.py
```diff
@@ -0,0 +1,57 @@
...
+        if self.mesh is not None:
+            # create cell to facet connectivity
+            self.mesh.topology.create_connectivity(
+                self.mesh.topology.dim, self.mesh.topology.dim - 1
+            )
+
+            # create facet to cell connectivity
+            self.mesh.topology.create_connectivity(
+                self.mesh.topology.dim - 1, self.mesh.topology.dim
+            )
```
RemDelaporteMathurin, line 47: I feel like this should be in a setter method

festim/geometry/mesh.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def define_markers(self, function_space):
```
RemDelaporteMathurin, line 49: I don't understand this:

- Why a method for a one liner that calls to another method of self?
- `festim.Mesh` doesn't have a `define_surface_markers` it is a `festim.Mesh1D` method
I don't see a case where we would do the following. I think that's because this method shouldn't be here. 
Especially since the markers are arguments of this class

festim/geometry/mesh.py
```diff
@@ -0,0 +1,57 @@
...
+        self.surface_markers = self.define_surface_markers(function_space)
+
+    def define_measures(self):
+        """Creates the ufl.Measure objects for self.ds"""
+
+        self.ds = ufl.Measure(
+            "ds", domain=self.mesh, subdomain_data=self.surface_markers
+        )
```
RemDelaporteMathurin, line 57: Why not dx?
Why not call this in the `__init__`?

festim/geometry/mesh_1d.py
```diff
@@ -0,0 +1,72 @@
...
+from festim import Mesh
+
+
+class Mesh1D(Mesh):
+    """
+    1D Mesh
+    Attributes:
+        vertices (list): the mesh x-coordinates
+        size (float): the size of the 1D mesh
+        V (dolfinx.fem.FunctionSpace): the function space of the simulation
```
RemDelaporteMathurin, line 14: One mesh can be used in several simulations, therefore can have several function spaces, meaning the function space is not inherent to the mesh: it is not an attribute of the mesh (although you need a mesh to make a FunctionSpace)

festim/geometry/mesh_1d.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    def __init__(self, vertices, **kwargs) -> None:
+        """Inits Mesh1D
+
+        Args:
+            vertices (list): the mesh x-coordinates
```
RemDelaporteMathurin, line 21:
```suggestion
            vertices (list): the mesh x-coordinates (m)
```

festim/geometry/mesh_1d.py
```diff
@@ -0,0 +1,72 @@
...
+
+        self.start = min(vertices)
+        self.size = max(vertices)
+        self.V = None
+
+        mesh = self.generate_mesh()
+        super().__init__(mesh=mesh, **kwargs)
+
+    def generate_mesh(self):
+        """Generates a 1D mesh"""
```
RemDelaporteMathurin, line 34: Can you write full docstrings pls?

festim/geometry/mesh_1d.py
```diff
@@ -0,0 +1,72 @@
...
+        cell = ufl.Cell(shape, geometric_dimension=gdim)
+        domain = ufl.Mesh(ufl.VectorElement("Lagrange", cell, degree))
+
+        mesh_points = np.reshape(self.vertices, (len(self.vertices), 1))
+        indexes = np.arange(self.vertices.shape[0])
+        cells = np.stack((indexes[:-1], indexes[1:]), axis=-1)
+
+        return mesh.create_mesh(MPI.COMM_WORLD, cells, mesh_points, domain)
+
+    def define_surface_markers(self, function_space):
```
RemDelaporteMathurin, line 45: I understand now the function_space thing.
So let me rephrase what I said earlier:

One mesh and a set of subdomains can be applicable to several different models.
HOWEVER, the mesh tags are only relevant for a given simulation since they are linked to the function space (and because the function space is unique for the simulation)

I think we therefore need to change the way we do things. 

We can keep this method cause it's still Geometry related, but just not store non-geometry-unique attributes in Geometry (so mesthtags, measures)

festim/hydrogen_transport_problem.py
```diff
@@ -0,0 +1,42 @@
...
+        boundary_conditions=[],
+        solver_parameters=None,
+        exports=[],
+    ) -> None:
+        self.geometry = geometry
+        self.species = species
+        self.temperature = temperature
+        self.boundary_conditions = boundary_conditions
+        self.solver_parameters = solver_parameters
+        self.exports = exports
```
RemDelaporteMathurin, line 33: Similar to comments above, I have realised that measures are attributes of the `Model` not `Geometry.`

RemDelaporteMathurin (changes requested):
First round of comment, I think we have to rethink what is an attribute of what.
The main comment is that ds and dx are not unique to the Geometry but unique to the model you are running. This is because the function space is unique to the model, not the geometry.

festim/geometry/geometry.py
```diff
@@ -0,0 +1,23 @@
...
+        """
+        Args:
+
+        """
+        self.mesh = mesh
+        self.subdomains = subdomains
+
+    def define_function_space(self):
+        elements = ufl.FiniteElement("CG", self.mesh.mesh.ufl_cell(), 1)
+        return fem.FunctionSpace(self.mesh.mesh, elements)
```
RemDelaporteMathurin, line 23: Again, this will only return one type of `FunctionSpace` only.
I don't think this is a method of `Geometry` but rather a method of `HydrogenTransportProblem.`

Which makes me think that maybe we don't need the `Geometry` class at all in the end since it's now only holding two attributes together (subdomains and mesh) and doesn't do anything with it

festim/geometry/mesh.py
RemDelaporteMathurin: Please address this

festim/hydrogen_transport_problem.py
```diff
@@ -32,11 +32,16 @@ def __init__(
         self.solver_parameters = solver_parameters
         self.exports = exports
 
+        self.dx = None
+        self.ds = None
+        self.function_spaces = []
```
RemDelaporteMathurin, line 37: 1 model = 1 function space

I don't see why you'd have several.

Proof is, in `self.functions_spaces` is only appended to in `initialise(),` which is a method only called once per simulation.

`self.function_spaces` shouldn't be a list.

festim/hydrogen_transport_problem.py
```diff
@@ -32,11 +32,16 @@ def __init__(
...
+
     def initialise(self):
         """Initialise the model. Creates suitable function
         spaces, the subdomains...
         """
 
-        # geometry
-        self.geometry.define_function_space()
-        self.geometry.define_subdomains()
+        self.function_spaces.append(self.geometry.define_function_space())
```
RemDelaporteMathurin, line 44: See my other comments, `define_function_space` shouldn't be a method of geometry but a method of `HydrogenTransportProblem`

```suggestion
        self.function_space = self.define_function_space()
```

festim/hydrogen_transport_problem.py
```diff
@@ -32,11 +32,16 @@ def __init__(
...
+        self.geometry.subdomains = self.geometry.mesh.define_measures(
+            self.function_spaces[0]
+        )
```
RemDelaporteMathurin, line 47: I don't understand. 

`Geometry` has an attribute subdomains of type list of `Subdomain` objects.
Here, you are assigning in this attribute, whatever is returned by `Mesh.define_measures()` which is a `fenicsx.ds` object

There is something wrong here

RemDelaporteMathurin (changes requested):

festim/hydrogen_transport_problem.py
```diff
@@ -0,0 +1,59 @@
...
+        spaces, the subdomains...
+        """
+
+        self.define_function_space()
+        (
+            self.dx,
+            self.ds,
+            self.facet_tags,
+            self.volume_tags,
+        ) = self.mesh.create_measures_and_tags(self.function_space)
```
RemDelaporteMathurin, line 55: Maybe we just need the dx and ds and not the tags here.

festim/mesh/mesh.py
```diff
@@ -0,0 +1,34 @@
...
+            self.mesh.topology.create_connectivity(
+                self.mesh.topology.dim - 1, self.mesh.topology.dim
+            )
+
+    def create_measures_and_tags(self, function_space):
+        """Creates the ufl.Measure objects for self.ds"""
+        facet_tags, volume_tags = self.create_meshtags(function_space)
+        dx = ufl.Measure("dx", domain=self.mesh, subdomain_data=volume_tags)
+        ds = ufl.Measure("ds", domain=self.mesh, subdomain_data=facet_tags)
+        return dx, ds, facet_tags, volume_tags
```
RemDelaporteMathurin, line 34: Maybe we don't need this to also return the tags but simply the dx and ds

RemDelaporteMathurin (approved):
Looks good to me, maybe the tags don't need to be stored as an attribute of HTransportProblem

RemDelaporteMathurin (approved):

PR #591, opened 2023-10-10
Permeation test case modification

jhdark:
## Proposed changes

Made some small changes to the test case to ensure the model works. i.e re-calling the `create_solver()` function to add the boundary conditions, defined after initialising, to the formulation and using the ds created in the HydrogenTransportProblem class. Furthermore, the test is compared against an analytical solution.

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/test/test_permeation_problem.py
+++ b/test/test_permeation_problem.py
@@ -8,13 +8,7 @@ from dolfinx.fem import (
     form,
     assemble_scalar,
 )
-from ufl import (
-    dot,
-    grad,
-    exp,
-    FacetNormal,
-    ds,
-)
+from ufl import dot, grad, exp, FacetNormal, ds, Measure
 from dolfinx import log
 import numpy as np
 import tqdm.autonotebook
@@ -76,13 +70,12 @@ def test_permeation_problem():
     )
     bc_outgas = dirichletbc(Constant(my_mesh.mesh, PETSc.ScalarType(0)), right_dofs, V)
     my_model.boundary_conditions = [bc_sieverts, bc_outgas]
-
-    final_time = 50
-
-    # log.set_log_level(log.LogLevel.INFO)
+    my_model.create_solver()
 
     mobile_xdmf = XDMFFile(MPI.COMM_WORLD, "mobile_concentration.xdmf", "w")
-    mobile_xdmf.write_mesh(my_mesh.mesh)
+    mobile_xdmf.write_mesh(my_model.mesh.mesh)
+
+    final_time = 50
 
     flux_values = []
     times = []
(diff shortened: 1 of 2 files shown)


test/test_permeation_problem.py
```diff
@@ -76,13 +70,12 @@ def siverts_law(T, S_0, E_S, pressure):
     )
     bc_outgas = dirichletbc(Constant(my_mesh.mesh, PETSc.ScalarType(0)), right_dofs, V)
     my_model.boundary_conditions = [bc_sieverts, bc_outgas]
-
-    final_time = 50
-
-    # log.set_log_level(log.LogLevel.INFO)
+    my_model.create_solver()
```
RemDelaporteMathurin, line 73: if this is what is needed for the test to work, then I'd like the call to `.create_solver()` to be removed from `.initialise()` for now.

test/test_permeation_problem.py
```diff
@@ -96,23 +89,43 @@ def siverts_law(T, S_0, E_S, pressure):
...
-
-        mobile_xdmf.write_function(u, t)
-
         # update previous solution
         mobile_H.prev_solution.x.array[:] = u.x.array[:]
 
     mobile_xdmf.close()
 
+    # analytical solution
+    S = 4.02e21 * exp(-1.04 / F.k_B / float(temperature))
```
RemDelaporteMathurin, line 106: The pre-exp and activation energy are duplicated (defined here and on line 67
```python
surface_conc = siverts_law(T=temperature, S_0=4.02e21, E_S=1.04, pressure=100)
```

Same goes for the temperature and mesh length. 

Can you store all these in variables to avoid code duplication?

RemDelaporteMathurin (changes requested):
Looks like a needed addition to this test.

Could you add docstrings to the test?

test/test_permeation_problem.py
```diff
@@ -96,23 +91,39 @@ def siverts_law(T, S_0, E_S, pressure):
...
+    summation = np.sum(
+        (-1) ** n_array
+        * np.exp(-((np.pi * n_array) ** 2) * float(my_model.D) / L**2 * times),
+        axis=0,
+    )
+    analytical_flux = P_up**0.5 * permeability / L * (2 * summation + 1)
+
+    flux_values = np.array(np.abs(flux_values))
+    normalised_analytical_flux = analytical_flux / analytical_flux.max()
+    normalised_compute_flux = flux_values / flux_values.max()
```
RemDelaporteMathurin, line 120: No we don't want to normalise anything! We may miss some buggy behaviour. Imagine the normalised curves match but not the absolute ones

test/test_permeation_problem.py
```diff
@@ -96,23 +91,39 @@ def siverts_law(T, S_0, E_S, pressure):
...
+
+    diff = np.abs(normalised_analytical_flux - normalised_compute_flux)
+    error = diff.mean()
+
+    assert error < 0.001
```
RemDelaporteMathurin, line 125: This is the absolute mean difference. We may want to go for the relative difference instead.

Imagine a case where the computed flux is zero at all times, and the analytical flux is always below 0.001 H/m2/s

Then the absolute difference will be below 0.001 (because 0.001 - 0 = 0.001) and yet the code doesn't work!

RemDelaporteMathurin (changes requested):
Thanks for addressing the comments, just a few things to change in the test

RemDelaporteMathurin:
Confirmed docs builds are failing due to a change outside of festim.

I'll try fixing it in another PR, ignore for now

PR #601, opened 2023-10-13
Check borders

jhdark:
## Proposed changes

A small feature to test the borders of the volume subdomains. Additional tests have been added to ensure it works. The current implementation only works in the case that only VolumeSubdomain1D are used, This will need to be accounted for in the future. 

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -138,6 +138,8 @@ class HydrogenTransportProblem:
                 )
                 tags_volumes[entities] = sub_dom.id
 
+        self.check_borders(self.mesh)
+
         # dofs and tags need to be in np.in32 format for meshtags
         dofs_facets = np.array(dofs_facets, dtype=np.int32)
         tags_facets = np.array(tags_facets, dtype=np.int32)
@@ -157,6 +159,39 @@ class HydrogenTransportProblem:
         self.dx = Measure(
             "dx", domain=self.mesh.mesh, subdomain_data=self.volume_meshtags
         )
+
+    def check_borders(self, mesh):
+        """Checks that the borders of the subdomain are within the domain
+
+        Args:
+            mesh (festim.Mesh): the mesh of the model
+
+        Raises:
+            Value error: if borders outside the domain
+        """
+        # check that subdomains are connected
+        all_borders = []
+        for vol in self.volume_subdomains:
+            for border in vol.borders:
+                all_borders.append(border)
+        if len(all_borders) == 0:
+            raise ValueError("No volume subdomains defined")
+        sorted_borders = np.sort(all_borders).reshape(int(len(all_borders) / 2), 2)
+        for i in range(0, len(sorted_borders) - 1):
+            if sorted_borders[i][1] != sorted_borders[i + 1][0]:
+                raise ValueError("Subdomain borders don't match to each other")
+
+        # check volume subdomain is defined
+        # TODO this possible by default
+        if len(all_borders) == 0:
+            raise ValueError("No volume subdomains defined")
+
+        # check that subdomains are within the domain
+        if (
+            sorted_borders[0][0] != mesh.vertices[0]
+            or sorted_borders[-1][-1] != mesh.vertices[-1]
+        ):
+            raise ValueError("borders dont match domain borders")
 
     def assign_functions_to_species(self):
         """Creates for each species the solution, prev solution and test function"""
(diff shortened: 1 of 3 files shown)


festim/hydrogen_transport_problem.py
```diff
@@ -158,6 +160,37 @@ def define_markers_and_measures(self):
             "dx", domain=self.mesh.mesh, subdomain_data=self.volume_meshtags
         )
 
+    def check_borders(self, mesh):
```
RemDelaporteMathurin, line 163: should this be a method of `Mesh1D` ?

we would pass all the subdomains as argument.

Something like

```python
# inside Mesh1D
def check_borders(self, list_of_subdomains1D):
    ....

# inside Problem

def define_markers_and_measures(self):
    ....
    if isinstance(self.mesh, F.Mesh1D):
        self.mesh.check_borders(self.subdomains)
```

festim/hydrogen_transport_problem.py
```diff
@@ -158,6 +160,37 @@ def define_markers_and_measures(self):
...
+            mesh (festim.Mesh): the mesh of the model
+
+        Raises:
+            Value error: if borders outside the domain
+        """
+        # check that subdomains are connected
+        all_borders = []
+        for vol in self.volume_subdomains:
+            for border in vol.borders:
+                all_borders.append(border)
```
RemDelaporteMathurin, line 176: We can make this a one liner if we want to
```suggestion
        all_borders = [border for border in vol.borders for vol in self.volume_subdomains]
```

festim/hydrogen_transport_problem.py
```diff
@@ -158,6 +160,37 @@ def define_markers_and_measures(self):
...
+        sorted_borders = np.sort(all_borders).reshape(int(len(all_borders) / 2), 2)
+        for i in range(0, len(sorted_borders) - 1):
+            if sorted_borders[i][1] != sorted_borders[i + 1][0]:
+                raise ValueError("Subdomain borders don't match to each other")
```
RemDelaporteMathurin, line 180:
```suggestion
        sorted_borders = np.sort(all_borders)
        for start, end in zip(sorted_borders[1:-2:2], sorted_borders[2:-1:2]):
            if start != end:
                raise ValueError("Subdomain borders don't match to each other")
```

festim/hydrogen_transport_problem.py
```diff
@@ -158,6 +160,37 @@ def define_markers_and_measures(self):
...
+        # check volume subdomain is defined
+        # TODO this possible by default
+        if len(all_borders) == 0:
+            raise ValueError("No volume subdomains defined")
+
+        # check that subdomains are within the domain
+        if (
+            sorted_borders[0][0] != mesh.vertices[0]
+            or sorted_borders[-1][-1] != mesh.vertices[-1]
+        ):
```
RemDelaporteMathurin, line 191:
```suggestion
        if (
            sorted_borders[0] != mesh.vertices[0]
            or sorted_borders[-1] != mesh.vertices[-1]
        ):
```

RemDelaporteMathurin (changes requested):
A couple of comments:
- refactoring suggestions
- maybe `check_borders` should be a method of `Mesh1D` and takes as argument a list of `Subdomain1D`

festim/mesh/mesh_1d.py
```diff
@@ -1,4 +1,4 @@
-from dolfinx import mesh
+from dolfinx import mesh as dolfin_mesh
```
RemDelaporteMathurin, line 1:
```suggestion
import dolfinx.mesh
```

festim/mesh/mesh_1d.py
```diff
@@ -32,4 +32,32 @@ def generate_mesh(self):
         indexes = np.arange(self.vertices.shape[0])
         cells = np.stack((indexes[:-1], indexes[1:]), axis=-1)
 
-        return mesh.create_mesh(MPI.COMM_WORLD, cells, mesh_points, domain)
+        return dolfin_mesh.create_mesh(MPI.COMM_WORLD, cells, mesh_points, domain)
```
RemDelaporteMathurin, line 35:
```suggestion
        return dolfinx.mesh.create_mesh(MPI.COMM_WORLD, cells, mesh_points, domain)
```

RemDelaporteMathurin (changes requested):
Better name space

PR #607, opened 2023-10-16
benchmarking script

jhdark:
## Proposed changes

Performance benchmarking is a useful way to keep track of how updates to the code affect performance.

A benchmark script is proposed comparing the performance of a simulation of gas permeation in pure fenicsx and in festim. It can be run using;
```python3 test/benchmark.py```
If the performance of festim is more than 20% worse than the fenics script on average, the test will fail rasing an error:
```ValueError: festim is {diff:.1f}% slower than fenics, current acceptble threshold of 20%```

More sophisticated methods of continuous testing can be implemented in the future: https://github.com/ionelmc/pytest-benchmark.  

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

## Further comments

### CODE REVIEW

File: benchmark.py [hunk 1]
```diff
@@ -0,0 +1,203 @@
...
+
+    festim_times = []
+    for i in range(repetitions):
+        start = time.time()
+        test_permeation_problem()
+        festim_times.append(time.time() - start)
+    festim_time = np.mean(festim_times)
+
+    diff = (np.abs(fenics_time - festim_time) / ((fenics_time + festim_time) / 2)) * 100
+    if diff > 20:
```
RemDelaporteMathurin:
```suggestion
    if diff > 10:
```

benchmark.py
```diff
@@ -0,0 +1,203 @@
...
+        raise ValueError(f"festim is {diff:.1f}% slower than fenics")
```
RemDelaporteMathurin: It'd be nice to print it nonetheless
```suggestion
        raise ValueError(f"festim is {diff:.1f}% slower than fenics")
    else:
        print(f"festim is {diff:.1f}% slower than fenics")
```

benchmark.py
```diff
@@ -0,0 +1,203 @@
...
+
+    relative_error = relative_error[
+        np.where(analytical_flux > 0.01 * np.max(analytical_flux))
+    ]
+    error = relative_error.mean()
+
+    assert error < 0.01
+
+
+def test_festim_vs_fenics_permeation_benchmark():
```
RemDelaporteMathurin:
```suggestion
def test_festim_vs_fenics_permeation_benchmark():
    """Runs a problem with pure fenicsx and the same problem with FESTIM and
    raise ValueError if difference is too high"""
```

RemDelaporteMathurin (changes requested):
- move the benchmark script to test
- change threshold

benchmark.py
jhdark: unfortunately, festim is already 10-12% slower as is, maybe 15 instead then?

test/benchmark.py
```diff
@@ -0,0 +1,202 @@
...
+
+    start = time.time()
+    test_permeation_problem(mesh_size=20001)
+    festim_time = time.time() - start
+
+    diff = (fenics_time - festim_time) / fenics_time
+    if diff < -0.1:
+        raise ValueError(
+            f"festim is {np.abs(diff):.1%} slower than fenics, current acceptable threshold of 10%"
+        )
```
RemDelaporteMathurin, line 196:
```suggestion
    threshold = -0.1
    if diff < threshold:
        raise ValueError(
            f"festim is {np.abs(diff):.1%} slower than fenics, current acceptable threshold of {threshold:.1%}"
        )
```

RemDelaporteMathurin (changes requested):
One small refactoring change and then good to merge!

Issue #598, opened 2023-10-12
Species and reactions


RemDelaporteMathurin:
This is a proposal for a species & reaction architecture for the fenicsx rewrite.

The main idea is to define explicitely the species (unknowns of the problem) and reactions ( A + B <-> C).
We would make use of the `Species` class and implement a `Reaction` class that takes reactants, product, and rate constants as arguments.

## Basic case

A very simple example would be to do:

```python
my_model = F.HTransportProblem()

cm = F.Species("cm")
ct = F.Species("ct")
empty_trap = F.Species("empty_trap")
my_model.species = [cm, ct, empty_trap]

my_model.reactions = [
    F.Reaction(
        reactant1=cm, reactant2=empty_trap, product=[ct], k_0=1, E_k=0.1, p_0=2, E_p=0.2
    ),
]
```
In the example above, `cm` reacts with `empty_trap` to create `ct`.
A reaction is created and passed to the model.

The system of equations would be:

$$\frac{\partial c_{m}}{\partial t} = \nabla \cdot D \nabla c_{m} - k c_{m} c_{empty} + p c_{t}$$

$$\frac{\partial c_{empty}}{\partial t} = -k c_{m} c_{empty} + p c_{t}$$

$$\frac{\partial c_{t}}{\partial t} = k c_{m} c_{empty} - p c_{t}$$

This opens up the possibility to have diffusing trapped concentrations and/or diffusing empty traps (#376).

## Implicit species

Now what if we don't want to explicitly define the empty trap but rather define it as $c_{empty} = n - c_t$ where $n=10$?
Then we have to come with a new class `ImplicitSpecies` (we can find better naming...) that would, under the hood, do the c_empty = n - c_t thing.

```python
my_model = F.HTransportProblem()

cm = F.Species("cm")
ct = F.Species("ct")

empty_trap = F.ImplicitSpecies("empty_trap", n=10, others=[ct])

my_model.species = [cm, ct]

my_model.reactions = [
    F.Reaction(
        reactant1=cm, reactant2=empty_trap, product=[ct], k_0=1, E_k=0.1, p_0=2, E_p=0.2
    ),
]
```

The system of equations is now:

$$ \frac{\partial c_{m}}{\partial t} = \nabla \cdot D \nabla c_{m} - k c_{m} (n - c_{t}) + p c_{t}$$

$$ \frac{\partial c_{t}}{\partial t} = k c_{m} (n - c_{t}) - p c_{t}$$

[...]

PR #614, opened 2023-10-19
Reactions

RemDelaporteMathurin:
## Proposed changes

This is a first step towards #598 

This PR implements a `Reaction` class and an `ImplicitSpecies` class.

For now the `Reaction` class just has a method called `reaction_term` that returns k * c1 * c2 - p * c3

This form will be used in the variational formulation later.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/species.py
+++ b/festim/species.py
@@ -1,4 +1,4 @@
-from typing import List, Union
+from typing import List, Union, Callable
 
 
 class Species:
@@ -31,6 +31,12 @@ class Species:
         self.solution = None
         self.prev_solution = None
         self.test_function = None
+
+    def __repr__(self) -> str:
+        return f"Species({self.name})"
+
+    def __str__(self) -> str:
+        return f"{self.name}"
 
     @property
     def concentration(self):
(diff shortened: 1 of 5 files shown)


festim/species.py
```diff
@@ -10,6 +13,7 @@ class Species:
         solution (dolfinx.fem.Function or ...): the solution for the current timestep
         prev_solution (dolfinx.fem.Function or ...): the solution for the previous timestep
         test_function (ufl.Argument or ...): the testfunction associated with this species
+        concentration (dolfinx.fem.Function): the concentration of the species
```
jhdark, line 16: does the species need a concentration attribute too?

jhdark (commented):
Looks good otherwise

festim/species.py
```diff
@@ -1,3 +1,6 @@
+from typing import List
```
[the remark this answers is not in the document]
RemDelaporteMathurin: Yes I used it for type hinting
https://github.com/RemDelaporteMathurin/FESTIM/blob/ec2d15ea9bd7764a0aef22a979ed9b54cf39f89c/festim/species.py#L92

festim/species.py
RemDelaporteMathurin: This can be a bit redundant since for `Species` the concentration is `solution`. Although it provides the users a more straightforward way to get the concentration.

But it mostly to slightly simplify the implementation of `Reaction.reaction_term()`:

https://github.com/RemDelaporteMathurin/FESTIM/blob/ec2d15ea9bd7764a0aef22a979ed9b54cf39f89c/festim/reaction.py#L80-L81

PR #615, opened 2023-10-19
Fix ufl cell vs basix cell

RemDelaporteMathurin:
## Proposed changes

This is a small PR to anticipate the future deprecation of `ufl.FiniteElement` and uses basix instead.
Also modifies the permeation test to avoid dividing by zero.

Overall reduces the number of warnings in the tests.


jhdark:
Could update this in the test/benchmark.py file too

jhdark (approved):
LGTM

PR #619, opened 2023-10-26
Update time dependent bcs

jhdark:
## Proposed changes

Currently, all boundary conditions are updated at each time step to ensure the value is updated in case it's dependent on time. With this change, only boundary conditions with values dependent on time will be updated at each time step. This should also improve performance.

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments


RemDelaporteMathurin:
Let's wait until #620 is merged before looking at this

RemDelaporteMathurin (approved):
This looks good to me! I know it's hard to test that the bc is updated only once.
Although we could potentially do it by timing the operation, but maybe a manual test is enough for now.

Have you tested it manually by having print statements?

jhdark:
> This looks good to me! I know it's hard to test that the bc is updated only once. Although we could potentially do it by timing the operation, but maybe a manual test is enough for now.
> 
> Have you tested it manually by having print statements?

Yep, I'll merge it then

PR #620, opened 2023-10-30
Time dependent temperature

jhdark:
## Proposed changes

This update would allow users to define temperature using a callable meaning the temperature can be a function of space or time. When a time_dependent temperature function is given, an attribute temperature_time_dependent is then True, which will then update the temperature function at each time step.

Usage:
```python
my_model.temperature = lambda t: 1.0 + t
```
or 
```python
def temperature_function(t):
    return 1.0 + t


my_model.temperature = temperature_function
```

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments


festim/hydrogen_transport_problem.py
```diff
@@ -99,17 +99,23 @@ def __init__(
         self.formulation = None
         self.volume_subdomains = []
         self.bc_forms = []
+        self.temperature_fenics_value = None
+        self.temperature_time_dependent = False
 
     @property
-    def temperature(self):
-        return self._temperature
+    def temperature_fenics_value(self):
```
RemDelaporteMathurin, line 106: Do we want to find something shorter like `temperature_fenics` or `temperature_value`?

festim/hydrogen_transport_problem.py
RemDelaporteMathurin, line 1: Shouldn't we have a setter/getter for `temperature` too? At least to do some type checking?

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +129,50 @@ def initialise(self):
...
+                    kwargs["t"] = self.t
+                    self.temperature_time_dependent = True
+                if "x" in arguments:
+                    kwargs["x"] = x
+
+                # store the expression of the boundary condition
+                # to update the value_fenics later
+                self.temperature_expr = fem.Expression(
+                    self.temperature(**kwargs),
+                    self.function_space.element.interpolation_points(),
```
RemDelaporteMathurin, line 172: `self.function_space.element` can either be CG1 or a `MixedElement`.
For some reason this seems to work, but it's not straightforward as `self.function_space` is for hydrogen species.

We want temperature to always be CG1.

Can we create a CG1 element here?

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +129,50 @@ def initialise(self):
...
+                self.temperature_fenics_value = fem.Function(self.function_space)
+                kwargs = {}
+                if "t" in arguments:
+                    kwargs["t"] = self.t
+                    self.temperature_time_dependent = True
+                if "x" in arguments:
+                    kwargs["x"] = x
+
+                # store the expression of the boundary condition
+                # to update the value_fenics later
```
RemDelaporteMathurin, line 169: `value_fenics`? I suppose that's `temperature_fenics_value` right?

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +129,50 @@ def initialise(self):
...
 
+        self.define_temperature()
         self.define_boundary_conditions()
         self.create_formulation()
         self.create_solver()
         self.defing_export_writers()
 
+    def define_temperature(self):
+        """If temperature value is given only dependent on t, create a
+        fem.Constant else create an expression to be updated and a function"""
```
RemDelaporteMathurin, line 140: This method does a lot of things. I would be a bit more descriptive here. Something along these lines. What do you think?

```suggestion
        """Sets the value of .temperature_fenics_value.
        The type depends on self.temperature.
        If self.temperature is a function on t only, create a
        fem.Constant. Else create an dolfinx.fem.Expression
        (stored in self.temperature_expr) to be updated, a
        dolfinx.fem.Function object is created from the Expression
        (stored in self.temperature_fenics_value)"""
```

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +129,50 @@ def initialise(self):
...
+        # if temperature is a float or int, create a fem.Constant
+        if isinstance(self.temperature, (float, int)):
+            self._temperature_fenics_value = F.as_fenics_constant(
+                self.temperature, self.mesh.mesh
+            )
+        elif isinstance(self.temperature, fem.Constant):
```
RemDelaporteMathurin, line 146: I think we also want to treat the case where it's given as a `fem.Function`
```suggestion
        elif isinstance(self.temperature, (fem.Constant, fem.Function)):
```

festim/hydrogen_transport_problem.py
```diff
@@ -424,3 +467,17 @@ def run(self):
             flux_values = [flux_values_1, flux_values_2]
 
         return times, flux_values
+
+    def update_time_dependent_values(self, t):
+        """Updates the time dependent values of the model"""
```
RemDelaporteMathurin, line 472:
```suggestion
        """
        Updates the time dependent values of the model
        like temperature, boundary conditions, sources, etc.
        """
```

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +129,50 @@ def initialise(self):
...
+                kwargs = {}
+                if "t" in arguments:
+                    kwargs["t"] = self.t
+                    self.temperature_time_dependent = True
+                if "x" in arguments:
+                    kwargs["x"] = x
+
+                # store the expression of the boundary condition
+                # to update the value_fenics later
+                self.temperature_expr = fem.Expression(
```
RemDelaporteMathurin, line 170: Add this attribute to docstrings

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +129,50 @@ def initialise(self):
         self.t = fem.Constant(self.mesh.mesh, 0.0)
         self.dt = self.settings.stepsize.get_dt(self.mesh.mesh)
 
+        self.define_temperature()
         self.define_boundary_conditions()
         self.create_formulation()
         self.create_solver()
         self.defing_export_writers()
 
+    def define_temperature(self):
```
RemDelaporteMathurin, line 138: Could we add a test at the beginning of this method that checks if `self.temperature` is `None` and raises an error if that's the case?

test/test_temperature.py
```diff
@@ -3,22 +3,127 @@
...
+        if callable(value):
+            my_model.initialise()
+        else:
+            with pytest.raises(TypeError):
+                my_model.initialise()
     else:
-        my_model.temperature = value
-        assert isinstance(my_model.temperature, fem.Constant)
+        my_model.initialise()
+        assert isinstance(my_model.temperature_fenics_value, fem.Constant)
```
RemDelaporteMathurin, line 37: I see there's quite the logic in this test. Let's minimise it by splitting the test into two: one that tests against valid inputs, another that tests against invalid inputs

RemDelaporteMathurin: Also, related to another comment I had above, if we keep the setter for `HydrogenTransportProblem.temperature`, then I don't think we need to modify all this

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +129,50 @@ def initialise(self):
...
+
+        # if temperature is callable, process accordingly
+        elif callable(self.temperature):
+            arguments = self.temperature.__code__.co_varnames
+            if "t" in arguments and "x" not in arguments and "T" not in arguments:
+                # only t is an argument
+                self.temperature_fenics_value = F.as_fenics_constant(
+                    mesh=self.mesh.mesh, value=self.temperature(t=float(self.t))
+                )
+                self.temperature_time_dependent = True
```
RemDelaporteMathurin, line 157: I feel like `temperature_time_dependent` should be a property of the class.

Something like:

```python
@property
def temperature_time_dependent(self):
    if self.temperature is None:
        return False
    if callable(self.temperature):
        arguments = self.temperature.__code__.co_varnames
        return "t" in arguments:
    else:
        return False

File: test/test_temperature.py [hunk 11]
```diff
@@ -3,22 +3,127 @@
...
+    my_model.initialise()
+
+    if isinstance(value, (float, int)):
+        assert not my_model.temperature_time_dependent
+    else:
+        arguments = value.__code__.co_varnames
+        if "t" in arguments:
+            assert my_model.temperature_time_dependent
+        else:
+            assert not my_model.temperature_time_dependent
```
RemDelaporteMathurin, line 79: This should be split into two tests, one for positive inputs another for negative inputs. The amount of logic in tests needs to be minimised to keep readability.

Also, as per the comment above, if `temperature_time_dependent` is a property of `HydrogenTransportProblem,` then this test will be simplified a lot.

test/test_temperature.py
```diff
@@ -3,22 +3,127 @@
...
+    subdomain = F.SurfaceSubdomain1D(1, x=4)
+    vol_subdomain = F.VolumeSubdomain1D(1, borders=[0, 4], material=dummy_mat)
+
+    my_model = F.HydrogenTransportProblem(
+        mesh=test_mesh,
+        subdomains=[vol_subdomain, subdomain],
+    )
+    my_model.species = [F.Species("H")]
+    my_bc = F.DirichletBC(subdomain, 1.0, "H")
+    my_model.boundary_conditions = [my_bc]
```
RemDelaporteMathurin, line 103: Not needed for this test

test/test_temperature.py
```diff
@@ -3,22 +3,127 @@
...
+
+    my_model.temperature = value
+
+    my_model.settings = F.Settings(atol=1, rtol=0.1, final_time=3)
+    my_model.settings.stepsize = F.Stepsize(initial_value=1)
+
+    # RUN
+    my_model.initialise()
+    my_model.run()
```
RemDelaporteMathurin, line 112: This test checks that the temperature is correct at the last time step only.
It would be better to check that the temperature is updated after each call of `iterate`.

BUT, we don't have an `iterate` method. So hold onto this until we have one (I'm on it.)

test/test_temperature.py
```diff
@@ -3,22 +3,127 @@
...
+        elif "x" in arguments:
+            expected_value = value(x=np.array([subdomain.x]))
+            computed_value = my_model.temperature_fenics_value.vector.array[-1]
+        elif "t" in arguments:
+            expected_value = value(t=3.0)
+            computed_value = float(my_model.temperature_fenics_value)
+
+    if isinstance(expected_value, Conditional):
+        expected_value = float(expected_value)
+    assert np.isclose(computed_value, expected_value)
```
RemDelaporteMathurin, line 129: Too much logic, hard to understand.
Instead have the arguments of the test function be a couple of values `T_function, expected_values`

where `T_function` is the current input, and `expected_values` is a list of three values for t=1, t=2, t=3.

When we have an `iterate` method, then it's as simple as:

```python
# TEST
for i in range(3):
    my_model.iterate()
    computed_value = ...  # simple test if computed value is Constant or Function
    computed_value = my_model.temperature_fenics_value.vector.array[-1]
    # or
    computed_value = float(my_model.temperature_fenics_value)
    assert np.isclose(computed_value, expected_value[i])
```

RemDelaporteMathurin (changes requested):
Looks great. First round of comments to improve things

test/test_temperature.py
RemDelaporteMathurin: See #621

test/test_temperature.py
RemDelaporteMathurin: See #621

festim/hydrogen_transport_problem.py
```diff
@@ -22,7 +22,8 @@ class HydrogenTransportProblem:
         mesh (festim.Mesh): the mesh of the model
         subdomains (list of festim.Subdomain): the subdomains of the model
         species (list of festim.Species): the species of the model
-        temperature (float or fem.Constant): the temperature of the model
+        temperature (float, int, fem.Constant, fem.Function or callable): the
+            temperature of the model
```
RemDelaporteMathurin, line 26:
```suggestion
        temperature (float, int, fem.Constant, fem.Function or callable): the
            temperature of the model (K)
```

festim/hydrogen_transport_problem.py
```diff
@@ -33,7 +34,8 @@ class HydrogenTransportProblem:
         mesh (festim.Mesh): the mesh of the model
         subdomains (list of festim.Subdomain): the subdomains of the model
         species (list of festim.Species): the species of the model
-        temperature (fem.Constant): the temperature of the model
+        temperature (float, int, fem.Constant, fem.Function or callable): the
+            temperature of the model
```
RemDelaporteMathurin, line 38:
```suggestion
        temperature (float, int, fem.Constant, fem.Function or callable): the
            temperature of the model (K)
```

festim/hydrogen_transport_problem.py
```diff
@@ -47,7 +49,14 @@ class HydrogenTransportProblem:
...
         solver (dolfinx.nls.newton.NewtonSolver): the solver of the model
-        multispecies (bool): True if the model has more than one species
+        multispecies (bool): True if the model has more than one species.
+        temperature_fenics (fem.Constant or fem.Function): the
+            temperature of the model as a fenics object (fem.Constant or
+            fem.Function).
+        temperature_expr (fem.Expression): the expression of the temperature
+            that is used to update the temperature_fenics
+        temperature_time_dependent (bool): True if the temperature is time
+
```
RemDelaporteMathurin, line 59:
```suggestion
        temperature_time_dependent (bool): True if the temperature is time
            dependent

```

festim/hydrogen_transport_problem.py
```diff
@@ -108,8 +118,37 @@ def temperature(self):
...
+    @property
+    def temperature_fenics(self):
+        return self._temperature_fenics
+
+    @temperature_fenics.setter
+    def temperature_fenics(self, value):
+        if value is None:
+            self._temperature_fenics = value
+            return
+        self._temperature_fenics = value
```
RemDelaporteMathurin, line 139: Ok. I see that in this case, having a setter is kinda useless

The point of having a setter is to do some type checking.

Here, we want to make sure that `temperature_fenics` is either a `Constant` or a `Function` and raise an error otherwise

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +162,66 @@ def initialise(self):
...
         self.create_formulation()
         self.create_solver()
         self.defing_export_writers()
 
+    def define_temperature(self):
+        """Sets the value of temperature_fenics_value. The type depends on
+        self.temperature. If self.temperature is a function on t only, create
+        a fem.Constant. Else, create an dolfinx.fem.Expression (stored in
+        self.temperature_expr) to be updated, a dolfinx.fem.Function object
+        is created from the Expression (stored in self.temperature_fenics_value)
```
RemDelaporteMathurin, line 176: add the error raise to the docstrings

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +162,66 @@ def initialise(self):
...
+        """
+        # check if temperature is None
+        if self.temperature is None:
+            raise TypeError("Temperature needs to be defined")
```
RemDelaporteMathurin, line 180:
```suggestion
            raise ValueError("the temperature attribute needs to be defined")
```

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +162,66 @@ def initialise(self):
...
+                # only t is an argument
+                self.temperature_fenics = F.as_fenics_constant(
+                    mesh=self.mesh.mesh, value=self.temperature(t=float(self.t))
+                )
+            else:
+                x = ufl.SpatialCoordinate(self.mesh.mesh)
+                CG_temperature = basix.ufl.element(
+                    basix.ElementFamily.P,
+                    self.mesh.mesh.basix_cell(),
+                    1,
```
RemDelaporteMathurin, line 204: What is `1` here? looks like a magic number

festim/hydrogen_transport_problem.py
```diff
@@ -123,11 +162,66 @@ def initialise(self):
...
+                if "t" in arguments:
+                    kwargs["t"] = self.t
+                if "x" in arguments:
+                    kwargs["x"] = x
+
+                # store the expression of the temperature
+                # to update the temperature_fenics later
+                self.temperature_expr = fem.Expression(
+                    self.temperature(**kwargs),
+                    function_space_temperature.element.interpolation_points(),
```
RemDelaporteMathurin, line 221: Do you have to create a functionspace or can you simply use `CG_temperature?`

festim/hydrogen_transport_problem.py
```diff
@@ -431,3 +523,13 @@ def iterate(
 
         # update previous solution
         self.u_n.x.array[:] = self.u.x.array[:]
+
+    def update_time_dependent_values(self, t):
```
RemDelaporteMathurin, line 527: Why passing `t` as an argument since it's an attribute of the class?

test/test_temperature.py
RemDelaporteMathurin, line 1: These tests should live in test_h_transport_problem.py

test/test_temperature.py
```diff
@@ -3,22 +3,116 @@
...
+        desc="Solving H transport problem",
+        total=my_model.settings.final_time,
+        unit_scale=True,
+    )
+    # TODO get rid of these when post processing is implemented
+    my_model.flux_values, my_model.times = [], []
+
+    # TEST
+    for i in range(3):
+        my_model.iterate()
```
RemDelaporteMathurin, line 98: I think you could test this behaviour by just calling `update_time_dependent_values()`

RemDelaporteMathurin (changes requested):

festim/hydrogen_transport_problem.py
jhdark: It seems we need to have a functionspace for this one

festim/hydrogen_transport_problem.py
jhdark: It was to clean the script a bit, seeing as the time will be used in all the updating variables, but can just use the attribute otherwise

jhdark: `float(self.t)` yes, I can change it

test/test_h_transport_problem.py
```diff
@@ -1,11 +1,100 @@
...
+        (lambda x, t: ufl.conditional(ufl.lt(t, 1.0), 100.0 + x[0], 0.0), True),
+    ],
+)
+def test_time_dependent_temperature_attribute(input, expected_value):
+    """Test that the temperature_time_dependent attribute is correctly set"""
+
+    my_model = F.HydrogenTransportProblem()
+    my_model.temperature = input
+
+    assert my_model.temperature_time_dependent == expected_value
```
RemDelaporteMathurin, line 51: Just a comment here, I think by convention comparing booleans with `is` instead of `==` is preferred.

`assert my_model.temperature_time_dependent is expected_value`

test/test_h_transport_problem.py
```diff
@@ -1,11 +1,100 @@
...
+
+
+def test_define_temperature_value_error_raised():
+    """Test that a ValueError is rasied when the temperature is None"""
```
RemDelaporteMathurin, line 55:
```suggestion
    """Test that a ValueError is raised when the temperature is None"""
```

One day we'll get them all

RemDelaporteMathurin (approved):
Happy to merge this!

PR #621, opened 2023-10-31
Iterate method

RemDelaporteMathurin:
As discussed in #620 , this adds an `.iterate` method to `HydrogenTransportProblem` to refactor the `run` method a bit.

Temporarily, I added a `skip_post_processing` boolean flag to the method (defaulting to `False`) to simplify the test since most of what we have in the post-processing atm will go away.

I also added a test for `iterate`. It doesn't test everything but tried to keep it small.

> Notes:
* we would need to add unit tests for the methods of `HydrogenTransportProblem`
* I added a bunch of new undocumented attributes for communication between `run` and `iterate`. These will be removed as soon as we have a proper post-processing step

### DIFF
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -379,7 +379,9 @@ class HydrogenTransportProblem:
 
         return self.times, self.flux_values
 
-    def iterate(self):
+    def iterate(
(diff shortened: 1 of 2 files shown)


jhdark (approved):
Looks good to me

Issue #626, opened 2023-11-01
[BUG] temperature cannot be ufl.conditional


RemDelaporteMathurin:
**Describe the bug**
When trying to set the temperature as a `ufl.conditional` it raises a `TypeError`.

```
Traceback (most recent call last):
  File "/home/remidm/festim/mwe.py", line 48, in <module>
    my_model.initialise()
  File "/home/remidm/festim/festim/hydrogen_transport_problem.py", line 184, in initialise
    self.define_temperature()
  File "/home/remidm/festim/festim/hydrogen_transport_problem.py", line 216, in define_temperature
    self.temperature_fenics = F.as_fenics_constant(
                              ^^^^^^^^^^^^^^^^^^^^^
  File "/home/remidm/festim/festim/helpers.py", line 22, in as_fenics_constant
    raise TypeError(
TypeError: Value must be a float, an int or a dolfinx.Constant, not <class 'ufl.conditional.Conditional'>
```

**To Reproduce**
Run:

```python
import festim as F
import numpy as np

L = 3e-04
vertices = np.linspace(0, L, num=1000)

my_mesh = F.Mesh1D(vertices)

my_model = F.HydrogenTransportProblem()
my_model.mesh = my_mesh

my_mat = F.Material(D_0=1.9e-7, E_D=0.2, name="my_mat")
my_subdomain = F.VolumeSubdomain1D(id=1, borders=[0, L], material=my_mat)
left_surface = F.SurfaceSubdomain1D(id=1, x=0)
right_surface = F.SurfaceSubdomain1D(id=2, x=L)
my_model.subdomains = [
    my_subdomain,
    left_surface,
    right_surface,
]

my_model.species = [F.Species("H")]


from ufl import conditional, lt

implantation_time = 100
implantation_temp = 300
temperature_ramp = 5  # K/s


def temp_function(t):
    ramping_temperature = implantation_temp + temperature_ramp * (t - implantation_time)
    return conditional(lt(t, implantation_time), implantation_temp, ramping_temperature)


my_model.temperature = temp_function

my_model.settings = F.Settings(
    atol=1e10,
    rtol=1e-10,
    max_iterations=30,
    final_time=200,
)

my_model.settings.stepsize = F.Stepsize(initial_value=1 / 20)

my_model.initialise()

```

RemDelaporteMathurin:
same problem for using `ufl.conditional` as the value of a `DirichletBC`.

I think we just need to adapt the logic a bit

[...]

PR #627, opened 2023-11-02
Better error message when users use conditional for functions of t only

RemDelaporteMathurin:
## Proposed changes

Linked to #626 

This PR adds a type check when converting temperature and DirichletBC values to a dolfinx object.
If the value is a function of `t` only, and the user used `ufl.conditional` ( _ie_ the type of the returned object is not a float or int) then an error is raised.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -3,6 +3,7 @@ from dolfinx.nls.petsc import NewtonSolver
 from dolfinx.io import XDMFFile
 import basix
 import ufl
+from ufl.core.operator import Operator
 from mpi4py import MPI
 from dolfinx.fem import Function, form, assemble_scalar
 from dolfinx.mesh import meshtags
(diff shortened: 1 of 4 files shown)


festim/hydrogen_transport_problem.py
```diff
@@ -196,6 +196,10 @@ def define_temperature(self):
         elif callable(self.temperature):
             arguments = self.temperature.__code__.co_varnames
             if "t" in arguments and "x" not in arguments and "T" not in arguments:
```
jhdark, line 198: should take the chance to remove this too
```suggestion
            if "t" in arguments and "x" not in arguments
```

RemDelaporteMathurin: Why? if `T` is in the arguments then it has to be a `Function` not a `Constant`

jhdark: This snippet is for defining the temperature, it was accidentally left in from a previous PR

Issue #623, opened 2023-11-01
Single output file for `TXTExport`


KulaginVladimir:
For long-term transient simulations with a small timestep, FESTIM produces a lot of files, when data is exported with `TXTExport`. Won't it be more practical to have one output file? 

Possibly, smt like that can be usefull?
```python
def is_it_first_time_to_export(self, current_time, nb_iteration):
    if self.times is None:
        return True if nb_iteration == 0 else False
    else:
        return True if np.isclose(self.times[0], current_time) else False
def write(self, current_time, nb_iteration, steady):
    # create a DG1 functionspace
    V_DG1 = f.FunctionSpace(self.function.function_space().mesh(), "DG", 1)
    solution = f.project(self.function, V_DG1)
    if self.is_it_time_to_export(current_time):
        if steady:
            filename = "{}/{}_steady.txt".format(self.folder, self.label)
            header = 'x,t=steady'
        else:
            filename = "{}/{}_transient.txt".format(self.folder, self.label)
            header = 'x,t={}s'.format(current_time)
        x = f.interpolate(f.Expression("x[0]", degree=1), V_DG1)
        # if the directory doesn't exist
        # create it
        dirname = os.path.dirname(filename)
        if not os.path.exists(dirname):
            os.makedirs(dirname, exist_ok=True)
        # if steady or it is the first time to export
        # write data
        # else append new column to the existing file    
        if steady or self.is_it_first_time_to_export(current_time, nb_iteration):
            np.savetxt(filename, np.transpose([x.vector()[:], solution.vector()[:]]), header = header, delimiter = ',', comments='')
        else:
            # Update the header
            old_file = open(filename)
            header = old_file.readline().split('\n')[0] + ',t={}s'.format(current_time)
            old_file.close()
            # Append new column
            np.savetxt(filename, np.column_stack([np.loadtxt(filename, delimiter = ',', skiprows = 1), np.transpose(solution.vector()[:])]), header=header, delimiter = ',', comments='')
```

RemDelaporteMathurin:
Hi @KulaginVladimir 

I must admit I'm not a big user of TXT exports, I usually write data to XDMF and then visualise it in Paraview. Have you ever used them?

Another alternative would be to only export the TXT at certain timesteps.

Also, for other exports (`XDMFExport`, `DerivedQuantities`) we have the option to say "write this export every N iterations". I don't remember if we have this for TXTExport or not, but that's surely something worth adding!

[...]

PR #628, opened 2023-11-02
Export data with TXTExport into single file

KulaginVladimir:
## Proposed changes

This PR modifies `TXTExport` according to #623.

For steady-state simulations, it produces 'FileLabel_steady.txt' with header: 'x,t=steady' in it. For transient simulations, it produces 'FileLabel_transient.txt' with header: 'x,t=Time[0]s,t=Time[1]s,..' where T corresponds either to `times` passed to `TXTExport` or all timesteps, if `times is None`.

I also modified previous tests to catch the updated `TXTExport`  and added some new.

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/exports/txt_export.py
+++ b/festim/exports/txt_export.py
@@ -49,7 +49,7 @@ class TXTExport(festim.Export):
         if self.times is None:
             return nb_iteration == 0
         else:
-            return True if np.isclose(self.times[0], current_time) else False
+            return np.isclose(self.times[0], current_time)
 
     def write(self, current_time, nb_iteration, steady):
         # create a DG1 functionspace
--- a/festim/exports/txt_export.py
+++ b/festim/exports/txt_export.py
@@ -33,6 +33,7 @@ class TXTExport(festim.Export):
     @property
     def filename(self):
         return f"{self.folder}/{self.label}.txt"
+
     def is_it_time_to_export(self, current_time):
         if self.times is None:
             return True
--- a/festim/exports/txt_export.py
+++ b/festim/exports/txt_export.py
@@ -46,22 +46,42 @@ class TXTExport(festim.Export):
                 return time
         return None
 
-    def write(self, current_time, steady):
+    def is_it_first_time_to_export(self, current_time, nb_iteration):
+        if self.times is None:
+            return True if nb_iteration == 0 else False
+        else:
+            return True if np.isclose(self.times[0], current_time) else False
+
+    def write(self, current_time, nb_iteration, steady):
         # create a DG1 functionspace
         V_DG1 = f.FunctionSpace(self.function.function_space().mesh(), "DG", 1)
 
         solution = f.project(self.function, V_DG1)
         if self.is_it_time_to_export(current_time):
-            filename = "{}/{}_{}s.txt".format(self.folder, self.label, current_time)
             if steady:
                 filename = "{}/{}_steady.txt".format(self.folder, self.label)
+                header = 'x,t=steady'
+            else:
+                filename = "{}/{}_transient.txt".format(self.folder, self.label)
+                header = 'x,t={}s'.format(current_time)
             x = f.interpolate(f.Expression("x[0]", degree=1), V_DG1)
             # if the directory doesn't exist
             # create it
             dirname = os.path.dirname(filename)
             if not os.path.exists(dirname):
                 os.makedirs(dirname, exist_ok=True)
-            np.savetxt(filename, np.transpose([x.vector()[:], solution.vector()[:]]))
+            # if steady or it is the first time to export
+            # write data
+            # else append new column to the existing file    
+            if steady or self.is_it_first_time_to_export(current_time, nb_iteration):
+                np.savetxt(filename, np.transpose([x.vector()[:], solution.vector()[:]]), header = header, delimiter = ',', comments='')
+            else:
+                # Update the header
+                old_file = open(filename)
+                header = old_file.readline().split('\n')[0] + ',t={}s'.format(current_time)
+                old_file.close()
+                # Append new column
+                np.savetxt(filename, np.column_stack([np.loadtxt(filename, delimiter = ',', skiprows = 1), np.transpose(solution.vector()[:])]), header=header, delimiter = ',', comments='')
 
 
 class TXTExports:
--- a/festim/exports/txt_export.py
+++ b/festim/exports/txt_export.py
@@ -33,7 +33,6 @@ class TXTExport(festim.Export):
     @property
     def filename(self):
         return f"{self.folder}/{self.label}.txt"
-        
     def is_it_time_to_export(self, current_time):
         if self.times is None:
             return True
--- a/festim/exports/txt_export.py
+++ b/festim/exports/txt_export.py
@@ -35,7 +35,6 @@ class TXTExport(festim.Export):
         for time in self.times:
             if np.isclose(time, current_time):
                 return True
-
         return False
 
     def when_is_next_time(self, current_time):
@@ -60,10 +59,10 @@ class TXTExport(festim.Export):
         if self.is_it_time_to_export(current_time):
             if steady:
                 filename = "{}/{}_steady.txt".format(self.folder, self.label)
-                header = 'x,t=steady'
+                header = "x,t=steady"
             else:
                 filename = "{}/{}_transient.txt".format(self.folder, self.label)
-                header = 'x,t={}s'.format(current_time)
+                header = "x,t={}s".format(current_time)
             x = f.interpolate(f.Expression("x[0]", degree=1), V_DG1)
             # if the directory doesn't exist
             # create it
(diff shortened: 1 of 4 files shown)


festim/exports/txt_export.py
```diff
@@ -46,22 +45,61 @@ def when_is_next_time(self, current_time):
                 return time
         return None
 
-    def write(self, current_time, steady):
+    def is_it_first_time_to_export(self, current_time, nb_iteration):
+        if self.times is None:
+            return True if nb_iteration == 0 else False
```
jhdark, line 50:
```suggestion
            return nb_iteration == 0 else False
```

festim/exports/txt_export.py
```diff
@@ -46,22 +45,61 @@ def when_is_next_time(self, current_time):
...
+        else:
+            return True if np.isclose(self.times[0], current_time) else False
```
jhdark, line 52:
```suggestion
            return np.isclose(self.times[0], current_time) else False
```

festim/exports/txt_export.py
```diff
@@ -46,22 +45,61 @@ def when_is_next_time(self, current_time):
                 return time
         return None
 
-    def write(self, current_time, steady):
+    def is_it_first_time_to_export(self, current_time, nb_iteration):
```
jhdark, line 48: Could do with some doc strings for this method

festim/exports/txt_export.py
```diff
@@ -46,22 +45,61 @@ def when_is_next_time(self, current_time):
...
+        if self.times is None:
+            return True if nb_iteration == 0 else False
+        else:
+            return True if np.isclose(self.times[0], current_time) else False
+
+    def write(self, current_time, nb_iteration, steady):
```
jhdark, line 54: Could do with some doc strings here

festim/exports/txt_export.py
```diff
@@ -46,22 +45,61 @@ def when_is_next_time(self, current_time):
...
+            # if steady or it is the first time to export
+            # write data
+            # else append new column to the existing file
+            if steady or self.is_it_first_time_to_export(current_time, nb_iteration):
+                np.savetxt(
+                    filename,
+                    np.transpose([x.vector()[:], solution.vector()[:]]),
+                    header=header,
+                    delimiter=",",
+                    comments="",
```
jhdark, line 81: not sure if this attribute is needed unless you had another use in mind for it?

festim/exports/txt_export.py
```diff
@@ -46,22 +45,61 @@ def when_is_next_time(self, current_time):
...
+                    filename,
+                    np.column_stack(
+                        [
+                            np.loadtxt(filename, delimiter=",", skiprows=1),
+                            np.transpose(solution.vector()[:]),
+                        ]
+                    ),
+                    header=header,
+                    delimiter=",",
+                    comments="",
```
jhdark, line 101: see above

test/system/test_misc.py
```diff
@@ -121,6 +121,38 @@ def test_wrong_value_for_bc_field(field):
...
+    my_model.T = F.Temperature(500)
+    my_model.dt = F.Stepsize(0.1)
+
+    my_export = F.TXTExport(
+        "solute", label="mobile_conc", times=[0.2, 0.5], folder=tmp_path
+    )
+    my_model.exports = [my_export]
+
+    my_model.initialise()
+    my_model.run()
```
jhdark, line 142: Is a way to test the export without having to make a system test? but I'm not sure on this

test/unit/test_exports/test_txt_export.py
```diff
@@ -105,3 +123,33 @@ def test_there_is_a_next_time(self, my_export):
     def test_last(self, my_export):
         assert my_export.when_is_next_time(3) is None
         assert my_export.when_is_next_time(4) is None
+
+
+class TestIsItFirstTimeToExport:
```
jhdark, line 128: Could do with some doc strings here

jhdark (changes requested):
Hey @KulaginVladimir  Congrats on your first PR 🚀. It looks great so far, just some minor refactoring things from me. I think @RemDelaporteMathurin has some comments to make too, just to potentially simplify things. Otherwise welcome! Thanks for the contribution, should be able to get this through soon without too much modification

festim/exports/txt_export.py
RemDelaporteMathurin: In fact, just 
```suggestion
            return nb_iteration == 0
```

festim/exports/txt_export.py
RemDelaporteMathurin:
```suggestion
            return np.isclose(self.times[0], current_time)
```

festim/exports/txt_export.py
RemDelaporteMathurin: That's partly on us but yes I agree!

festim/exports/txt_export.py
```diff
@@ -46,22 +45,61 @@ def when_is_next_time(self, current_time):
...
         V_DG1 = f.FunctionSpace(self.function.function_space().mesh(), "DG", 1)
 
         solution = f.project(self.function, V_DG1)
         if self.is_it_time_to_export(current_time):
-            filename = "{}/{}_{}s.txt".format(self.folder, self.label, current_time)
             if steady:
                 filename = "{}/{}_steady.txt".format(self.folder, self.label)
+                header = "x,t=steady"
+            else:
+                filename = "{}/{}_transient.txt".format(self.folder, self.label)
```
RemDelaporteMathurin, line 64: If we are now writing one file for all timesteps, then I don't think we need two different filenames.

Instead, we could have an attribute `.filename` defined in the `__init__` method:

```python
self.filename = f"{self.folder}/{self.label}.txt"
```

festim/exports/txt_export.py
RemDelaporteMathurin, line 48: I don't think we need a separate method for that. This requires that we pass another argument to the write method.

Instead we could have a `._first_time` attribute as a boolean, initialised as `True` to simplify things.

I'll make a PR to your branch to show what I'm talking about

festim/exports/txt_export.py
RemDelaporteMathurin: I think this is needed because otherwise the default value for coments is `"#"` and it adds this symbol to the header.
Correct @KulaginVladimir ?

test/system/test_misc.py
RemDelaporteMathurin: There must be a way to just call `write()` but I don't know if it's worth it, this does the trick

test/unit/test_exports/test_txt_export.py
RemDelaporteMathurin: Kinda hard to document a test class, you can document the test functions however.

RemDelaporteMathurin (changes requested):
Hi @KulaginVladimir thanks for this PR! First of many I hope! 🎉 

This is a very good quality of life feature!

I have a few comments, the main one being refactoring this by replacing the `is_first_time_to_export` method by a class attribute.
I'll make another PR to your branch that you can then merge to add the changes here ( _ie_ no need to close this PR and make a new one).

festim/exports/txt_export.py
KulaginVladimir: @RemDelaporteMathurin, right. I think `comments` should certainly be turned off when new columns are added to the output file

festim/exports/txt_export.py
KulaginVladimir: > Instead we could have a ._first_time attribute as a boolean, initialised as True to simplify things.

Got it. It surely makes things better!

RemDelaporteMathurin: See [this PR](https://github.com/KulaginVladimir/FESTIM/pull/1) to your fork

festim/exports/txt_export.py
KulaginVladimir: Maybe, just use a variable `filename` in `write()`? Otherwise, `test_create_folder` in `test_txt_export` fails

RemDelaporteMathurin: Well in theory it doesn't make to redefine it every time we write

KulaginVladimir: ok, is `test_create_folder` needed? I think, creation of passed folder is checked in `test_file_exists`

RemDelaporteMathurin: Yes we need to keep these tests. Maybe we can adapt them?

RemDelaporteMathurin: I think I understand the reason why `test_file_exists` fails, it's because as `filename` is created at the instantiation of the class, if you modify the `folder` attribute afterwards it doesn't modify `filename.`

The solution to this is to have it as a `property` of the class.

To add a `property` attribute, simply add:
```python
    def __init__(self, field, label, folder, times=None) -> None:
        super().__init__(field=field)
        if times:
            self.times = sorted(times)
        else:
            self.times = times
        self.label = label
        self.folder = folder
        self._first_time = True

    @property
    def filename(self):
        return f"{self.folder}/{self.label}.txt"
```

test/system/test_misc.py
```diff
@@ -121,6 +121,38 @@ def test_wrong_value_for_bc_field(field):
...
+
+    assert os.path.exists(
+        "{}/{}.txt".format(my_export.folder, my_export.label)
+    )
```
RemDelaporteMathurin, line 146:
```suggestion
    assert os.path.exists("{}/{}.txt".format(my_export.folder, my_export.label))
```

pep8 formatting

festim/exports/txt_export.py
```diff
@@ -28,14 +28,18 @@ def __init__(self, field, label, folder, times=None) -> None:
             self.times = times
         self.label = label
         self.folder = folder
+        self._first_time = True
 
+    @property
+    def filename(self):
+        return f"{self.folder}/{self.label}.txt"
+
```
RemDelaporteMathurin, line 36: Trailing whitespace
```suggestion

```

RemDelaporteMathurin (changes requested):
Yay this looks good to me! There are just a few formatting errors like whitespaces and line-length, etc.

I've suggested a couple but there are more (see the [linting check failing](https://github.com/RemDelaporteMathurin/FESTIM/actions/runs/6746146871/job/18340264044?pr=628))

It is super easy to format your code, just run:

```
python -m black .
```

This will run the black formatter on all the files and make sure they are correctly formatted.
If you don't have black installed, you can install it with pip.

RemDelaporteMathurin (approved):
Looks good to me! Congrats @KulaginVladimir on your first PR! 🎉 🚀 

Feel free to join the [FESTIM slack channel](https://join.slack.com/t/festimworkspace/shared_invite/zt-246hw8d6o-htWASLsbdosUo_2nRKCf9g) where we discuss about development, but also user support.

On another note, tell me if I'm wrong but by the look of your commit messages, it looks like you made the code modifications directly in github.com?
If that's so, for next time, I think it would be easier for you to clone the repository locally, do the changes, then commit and push the changes to your remote repository on github.com
It will give you more control of what you are doing and it's good practice in general.

If you are still interested in contributing to FESTIM, don't hesitate to take a look at [issues tagged with "Good first issues"](https://github.com/RemDelaporteMathurin/FESTIM/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22). These are issues that can be easily fixed and that we usually leave open for newcomers!

RemDelaporteMathurin:
@jhdark I'll leave this open in case you have additional things to say. The docs are failing due to a build issue, it has been fixed in another PR #630

jhdark (approved):
Looks great to me

PR #629, opened 2023-11-02
Surface Quantities: Surface Flux

jhdark:
## Proposed changes

With this change users will now be able to export the surface flux at a given surface subdomain at each time step using:
```python
my_species = F.Species("A")
my_subdomain = F.SurfaceSubdomain1D(id=1, x=1)

my_export = F.SurfaceFlux(
        filename="surface_flux.csv",
        field=my_species,
        surface_subdomain=my_subdomain,
    )
```
With this implementation, the filename is optional, however, the export data will still be accessible after the simulation. Additionally, the time of the simulation is also saved at each iteration:

```python
simulation_times = my_export .t
flux_values = my_export .data
```

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments


festim/hydrogen_transport_problem.py
```diff
@@ -268,8 +324,12 @@ def define_function_space(self):
 
         self.function_space = fem.FunctionSpace(self.mesh.mesh, element)
 
-        self.u = Function(self.function_space)
-        self.u_n = Function(self.function_space)
+        # create global DG function spaces of degree 0 and 1
+        self.V_DG_0 = fem.FunctionSpace(self.mesh.mesh, ("DG", 0))
+        self.V_DG_1 = fem.FunctionSpace(self.mesh.mesh, ("DG", 1))
```
RemDelaporteMathurin, line 329: Needs adding to docstrings

festim/hydrogen_transport_problem.py
```diff
@@ -483,48 +545,18 @@ def run(self):
...
-                D_2 = self.subdomains[0].material.get_diffusion_coefficient(
-                    self.mesh.mesh, self.temperature_fenics, self.species[1]
-                )
-                surface_flux_1 = form(D_1 * dot(grad(cm_1), self.mesh.n) * self.ds(2))
-                surface_flux_2 = form(D_2 * dot(grad(cm_2), self.mesh.n) * self.ds(2))
-                flux_1 = assemble_scalar(surface_flux_1)
-                flux_2 = assemble_scalar(surface_flux_2)
-                self.flux_values_1.append(flux_1)
-                self.flux_values_2.append(flux_2)
-                self.times.append(float(self.t))
```
RemDelaporteMathurin, line 523: 🙏🙏🙏

festim/exports/surface_quantity.py
```diff
@@ -0,0 +1,73 @@
...
+            raise TypeError("field must be of type festim.Species")
+
+        self._field = value
+
+    @property
+    def write_to_file(self):
+        if self.filename is None:
+            return False
+        else:
+            return True
```
RemDelaporteMathurin, line 68: Tell me if I'm wrong but this seems to be used only in the __init__ correct?
Is it worth having a property for this or can we get away with just `if self.filename`?

Happy to keep it if we think it's worth it

festim/hydrogen_transport_problem.py
```diff
@@ -537,5 +569,35 @@ def update_time_dependent_values(self):
...
+            # update global D if temperature time dependent or internal
+            # variables time dependent
+            species_not_updated = self.species.copy()  # make a copy of the species
+            for export in self.exports:
+                if isinstance(export, F.SurfaceFlux):
+                    # if the D of the species has not been updated yet
+                    if export.field in species_not_updated:
+                        export.D.interpolate(export.D_expr)
+                        species_not_updated.remove(export.field)
+
```
RemDelaporteMathurin, line 581: Should this go in `post_processing` ?

festim/exports/surface_flux.py
```diff
@@ -0,0 +1,53 @@
...
+class SurfaceFlux(F.SurfaceQuantity):
+    """Exports surface flux at a given subdomain
+
+    Args:
+        field (festim.Species): species for which the surface flux is computed
+        surface_subdomain (festim.SurfaceSubdomain1D): surface subdomain
+        filename (str, optional): name of the file to which the surface flux is exported
+
+    Attributes:
+        field (festim.Species): species for which the surface flux is computed
```
RemDelaporteMathurin, line 16: Note to ourselves: one day we will want to compute HeatFluxes

festim/exports/surface_flux.py
```diff
@@ -0,0 +1,53 @@
...
+        t (list): list of time values
+        data (list): list of surface flux values
+        title (str): title of the exported data
+        write_to_file (bool): True if the data is exported to a file
+    """
+
+    def __init__(
+        self,
+        field,
+        surface_subdomain: int or F.SurfaceSubdomain1D,
```
RemDelaporteMathurin, line 28: we may want to just name this argument `surface`

RemDelaporteMathurin, line 28: Here this argument cannot be an `int` since the `id` attribute is used for setting `self.title` just below.

festim/exports/surface_flux.py
```diff
@@ -0,0 +1,53 @@
...
+        self.data = []
+
+        self.title = "Flux surface {}: {}".format(
+            self.surface_subdomain.id, self.field.name
+        )
+
+        if self.write_to_file:
+            with open(self.filename, mode="w", newline="") as file:
+                writer = csv.writer(file)
+                writer.writerow(["Time", f"{self.title}"])
```
RemDelaporteMathurin, line 43: Do we want to specify units?

RemDelaporteMathurin (changes requested):
Haven't looked at the tests yet but here are a few comments

festim/hydrogen_transport_problem.py
RemDelaporteMathurin, line 1: I think you want to also remove these lines:
https://github.com/RemDelaporteMathurin/FESTIM/blob/b7898a84a47b03424e9a2dc0b80c6c512ceb702a/festim/hydrogen_transport_problem.py#L543-L544

RemDelaporteMathurin: In fact you would remove self.flux_values altogether right?

jhdark: yessir

festim/exports/surface_quantity.py
jhdark: Yeah you're right, not needed really

festim/exports/surface_flux.py
jhdark: it would be easier to understand

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,139 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+    )
+
+    my_model.define_function_spaces()
+    my_model.define_markers_and_measures()
+    my_model.define_temperature()
+    my_model.initialise_exports()
+
+    Ds = []
+    for export in my_model.exports:
+        Ds.append(export.D)
```
RemDelaporteMathurin, line 343:
```suggestion
    Ds = [export.D for export in my_model_exports]
```

test/test_surface_quantity.py
```diff
@@ -0,0 +1,128 @@
...
+        field=my_species,
+        surface=dummy_surface,
+    )
+    my_export.D = D
+
+    # RUN
+    my_export.compute(n=my_mesh.n, ds=ds)
+
+    # TEST
+    expected_value = -D * 4 * dummy_surface.x
```
RemDelaporteMathurin, line 50:
```suggestion
    # flux = -D grad(c)_ \cdot n = -D 4 x
    expected_value = -D * 4 * dummy_surface.x
```

test/test_surface_quantity.py
```diff
@@ -0,0 +1,128 @@
...
+    my_export.value = 2.0
+
+    for i in range(10):
+        my_export.write(i)
+
+    computed_value = len(np.genfromtxt(my_export.filename, delimiter=","))
+
+    expected_value = 10
+
+    assert computed_value == expected_value
```
RemDelaporteMathurin, line 115: Instead of just checking the final length, you could check the length at each step.

```suggestion
        file_length = len(np.genfromtxt(my_export.filename, delimiter=","))

        expected_length = i + 1

        assert file_length == expected_length
```

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+    my_model.exports = [F.SurfaceFlux(field="J", surface=surf)]
+    my_model.define_function_spaces()
+
+    # TEST
+    with pytest.raises(ValueError, match="Species J not found in list of species"):
+        my_model.initialise_exports()
+
+
+def test_define_D_global_different_temperatures():
+    """Test that the D_global attribute is correctly defined when the temperature
```
RemDelaporteMathurin, line 232: there is no " `D_global` attribute"

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+            F.VolumeSubdomain1D(id=2, borders=[2, 4], material=my_mat),
+        ],
+        species=[H],
+        temperature=lambda x: 100.0 * x[0] + 50,
+        exports=[
+            F.SurfaceFlux(
+                field=H,
+                surface=surf,
+            ),
+        ],
```
RemDelaporteMathurin, line 252: I think we don't need to define exports for this test.
my_model.define_D_global doesn't look for exports

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+    D_analytical_left = D_0 * np.exp(-E_D / (F.k_B * 50))
+    D_analytical_right = D_0 * np.exp(-E_D / (F.k_B * 450))
+
+    expected_values = [D_analytical_left, D_analytical_right]
+
+    assert np.isclose(computed_values, expected_values).all()
+
+
+def test_define_D_global_different_materials():
+    """Test that the D_global attribute is correctly defined when the material
```
RemDelaporteMathurin, line 272: Not an attribute

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+            F.VolumeSubdomain1D(id=2, borders=[2, 4], material=my_mat_R),
+        ],
+        species=[F.Species("H"), F.Species("D")],
+        temperature=500,
+        exports=[
+            F.SurfaceFlux(
+                field=H,
+                surface=surf,
+            ),
+        ],
```
RemDelaporteMathurin, line 294: Same comment as above, don't need exports here I think

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+    surf = F.SurfaceSubdomain1D(id=1, x=0)
+    H = F.Species("H")
+
+    my_model = F.HydrogenTransportProblem(
+        mesh=F.Mesh1D(np.linspace(0, 4, num=101)),
+        subdomains=[
+            F.VolumeSubdomain1D(id=1, borders=[0, 2], material=my_mat_L),
+            F.VolumeSubdomain1D(id=2, borders=[2, 4], material=my_mat_R),
+        ],
+        species=[F.Species("H"), F.Species("D")],
```
RemDelaporteMathurin, line 287: 1) Why two species?
2) You seem to define twice F.Species("H"), here and on line 281 just above

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+    my_model.define_function_spaces()
+    my_model.define_markers_and_measures()
+    my_model.define_temperature()
+
+    D_computed, D_expr = my_model.define_D_global(H)
+
+    computed_values = [D_computed.x.array[0], D_computed.x.array[-1]]
+
+    D_analytical_left = D_0_left * np.exp(-E_D_left / (F.k_B * my_model.temperature))
+    D_analytical_right = D_0_right * np.exp(-E_D_right / (F.k_B * my_model.temperature))
```
RemDelaporteMathurin, line 306: `analytical`? why?
I think you mean expected here

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+
+    expected_values = [D_analytical_left, D_analytical_right]
+
+    assert np.isclose(computed_values, expected_values).all()
+
+
+def test_initialise_exports_multiple_exports_same_species():
+    """Test that the D attribute is the same for multiple exports of the same species,
+    and that D_global is only created once per species"""
```
RemDelaporteMathurin, line 315: I don't see where you test this

RemDelaporteMathurin: You seem to be testing that the first `D` has the same value as the second `D` but this could very well mean they are two different objects that have the same value. You want to check that they are the same object.

In that case, a simple `==` comparison should do.

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+    )
+
+    my_model.define_function_spaces()
+    my_model.define_markers_and_measures()
+    my_model.define_temperature()
+    my_model.initialise_exports()
+
+    Ds = [export.D for export in my_model.exports]
+
+    assert np.isclose(Ds[0].x.array[0], Ds[1].x.array[0])
```
RemDelaporteMathurin, line 349: Here I think that if you want to check that `Ds[0]` is the same _object_ as `Ds[1]` then you want:

`Ds[0] == Ds[1]`

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+
+
+def test_define_D_global_multispecies():
+    """Test that the D_global attribute is correctly defined when there are multiple
```
RemDelaporteMathurin, line 353: not an attribute

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+        exports=[
+            F.SurfaceFlux(
+                field=A,
+                surface=surf,
+            ),
+            F.SurfaceFlux(
+                field=A,
+                surface=surf,
+            ),
+        ],
```
RemDelaporteMathurin, line 382: No need for exports here

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+
+    D_A_computed, D_A_expr = my_model.define_D_global(A)
+    D_B_computed, D_B_expr = my_model.define_D_global(B)
+
+    computed_values = [D_A_computed.x.array[-1], D_B_computed.x.array[-1]]
+
+    D_analytical_A = D_0_A * np.exp(-E_D_A / (F.k_B * my_model.temperature))
+    D_analytical_B = D_0_B * np.exp(-E_D_B / (F.k_B * my_model.temperature))
+
+    expected_values = [D_analytical_A, D_analytical_B]
```
RemDelaporteMathurin, line 397: Again I don't think that "analytical" is the correct word here

test/test_h_transport_problem.py
```diff
@@ -207,3 +207,240 @@ def test_update_time_dependent_values_temperature(T_function, expected_values):
...
+
+    assert np.isclose(computed_values, expected_values).all()
+
+
+def test_post_processing_update_D_global():
+    """Test that the D_global attribute is updated at each time
```
RemDelaporteMathurin, line 403: Not an attribute

RemDelaporteMathurin (changes requested):

festim/exports/surface_flux.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+        self.data.append(self.value)
+
+    def initialise_export(self):
+        self.title = "Flux surface {}: {}".format(self.surface.id, self.field.name)
+
+        if self.filename is not None:
+            with open(self.filename, mode="w", newline="") as file:
+                writer = csv.writer(file)
+                writer.writerow(["t(s)", f"{self.title}"])
```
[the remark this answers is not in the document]
jhdark: I did have this originally, but if we do this, we would have to add another find species from string test if the field is given as a string, to get the name of the species for the title. Whereas like this we can take advantage of the check done in L238 in HydrogenTransportProblem.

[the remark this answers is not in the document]
RemDelaporteMathurin: Well in that case it can be done at the first call of `write` with just a check that the file doesn't exist.

festim/exports/surface_quantity.py
```diff
@@ -21,6 +21,9 @@ def __init__(self, field, surface, filename: str = None) -> None:
         self.surface = surface
         self.filename = filename
 
+        self.t = []
+        self.data = []
```
RemDelaporteMathurin, line 25: these attributes should be documented

test/test_surface_quantity.py
RemDelaporteMathurin, line 1: Is there any test in here that checks if a file is created when calling .write?

jhdark: There is the test `test_writer()` which makes sure a new row is written to the output file at each time the .write method is called. Is that enough?

RemDelaporteMathurin (approved):

PR #635, opened 2023-11-07
Tests for reactions

jhdark:
Added some tests for this new PR


jhdark:
I think volume should be an argument of reaction, then once the source term PR goes through, we can adapt this to accept a subdomain id int or a list of volume subdomain ids. What do you think @RemDelaporteMathurin

PR #636, opened 2023-11-08
support sources

jhdark:
## Proposed changes

Sources!

With this update, users will be able to define volumetric source terms. These terms can be constant throughout a volume subdomain or be dependent on space, time or temperature.

Example usage:
```python
my_volume = F.VolumeSubdomain1D(id=1, borders=[0, 1], material="my_material")
H = F.Species("H)

F.Source(value=1e15, volume=my_volume, species=H)
```

For the moment it only accepts a F.VolumeSubdomain as the volume, but perhaps we can create a similar method to `find_species_from_name` in `Species` and create `find_volume_from_id`. This way we could accept the integer id here and perhaps in other places that may use a volume subdomain.  What do you think @RemDelaporteMathurin .

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -167,7 +167,7 @@ class HydrogenTransportProblem:
 
         self.define_temperature()
         self.define_boundary_conditions()
-        self.define_sources()
+        self.create_source_values_fenics()
         self.create_formulation()
         self.create_solver()
         self.initialise_exports()
@@ -481,19 +481,32 @@ class HydrogenTransportProblem:
 
         return form
 
-    def define_sources(self):
-        """Create a fenics object value for each source of the model"""
+    def create_source_values_fenics(self):
+        """For each source in the model, check if the species and volumes are lists,
+        create one if not, check if each element in species and volumes are
+        festim.Species and festim.VolumeSubdomain1D, then create the value fenics
+        attribute.
+        """
         for source in self.sources:
+            # iterate through species, find species if name of species is given
+            if not isinstance(source.species, list):
+                source.species = [source.species]
             for idx, spe in enumerate(source.species):
                 if isinstance(spe, str):
                     # if name of species is given then replace with species object
                     source.species[idx] = F.find_species_from_name(spe, self.species)
+
+            # iterate through volumes, find volume if id of volume is given
+            if not isinstance(source.volume, list):
+                source.volume = [source.volume]
             for idx, vol in enumerate(source.volume):
                 if isinstance(vol, int):
                     # if name of species is given then replace with species object
                     source.volume[idx] = F.find_volume_from_id(
                         source.volume, self.volume_subdomains
                     )
+
+            # create value_fenics for all F.Source objects
             if isinstance(source, F.Source):
                 function_space_value = None
                 if callable(source.value):
@@ -505,7 +518,7 @@ class HydrogenTransportProblem:
                             0
                         ].collapsed_function_space
 
-                source.create_value(
+                source.create_value_fenics(
                     mesh=self.mesh.mesh,
                     temperature=self.temperature_fenics,
                     function_space=function_space_value,
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -530,7 +530,9 @@ class HydrogenTransportProblem:
         # add sources
         for source in self.sources:
             self.formulation -= (
-                source.value_fenics * source.species.test_function * self.dx(vol.id)
+                source.value_fenics
+                * source.species.test_function
+                * self.dx(source.volume.id)
             )
 
             # add fluxes
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -167,6 +167,7 @@ class HydrogenTransportProblem:
 
         self.define_temperature()
         self.define_boundary_conditions()
+        self.define_sources()
         self.create_formulation()
         self.create_solver()
         self.initialise_exports()
@@ -480,6 +481,29 @@ class HydrogenTransportProblem:
 
         return form
 
+    def define_sources(self):
+        """Defines the sources of the model"""
+        for source in self.sources:
+            if isinstance(source.species, str):
+                # if name of species is given then replace with species object
+                source.species = F.find_species_from_name(source.species, self.species)
+            if isinstance(source, F.Source):
+                function_space_value = None
+
+                if callable(source.value):
+                    # if bc.value is a callable then need to provide a functionspace
+                    if not self.multispecies:
+                        function_space_value = source.species.sub_function_space
+                    else:
+                        function_space_value = source.species.collapsed_function_space
+
+                source.create_value(
+                    mesh=self.mesh.mesh,
+                    temperature=self.temperature_fenics,
+                    function_space=function_space_value,
+                    t=self.t,
+                )
+
     def create_formulation(self):
         """Creates the formulation of the model"""
         if len(self.sources) > 1:
@@ -503,11 +527,10 @@ class HydrogenTransportProblem:
                 self.formulation += ((u - u_n) / self.dt) * v * self.dx(vol.id)
 
                 # add sources
-                # TODO implement this
-                # for source in self.sources:
-                #     # f = Constant(my_mesh.mesh, (PETSc.ScalarType(0)))
-                #     if source.species == spe:
-                #         formulation += source * v * self.dx
+                for source in self.sources:
+                    if source.species == spe:
+                        self.formulation -= source.value_fenics * v * self.dx
+
                 # add fluxes
                 # TODO implement this
                 # for bc in self.boundary_conditions:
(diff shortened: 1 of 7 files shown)


festim/source.py
```diff
@@ -0,0 +1,148 @@
...
+        volume (festim.VolumeSubdomain1D): the volume subdomain where the source
+            is applied
+        value (float, int, fem.Constant or callable): the value of the soure
+        species (str): the name of the species to which the source is applied
+
+    Attributes:
+        volume (festim.VolumeSubdomain1D): the volume subdomain where the source
+            is applied
+        value (float, int, fem.Constant or callable): the value of the soure
+        species (str): the name of the species to which the source is applied
```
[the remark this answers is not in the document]
RemDelaporteMathurin: Also, could users provide several species?

festim/hydrogen_transport_problem.py
```diff
@@ -480,6 +481,37 @@ def create_dirichletbc_form(self, bc):
...
+                source.create_value(
+                    mesh=self.mesh.mesh,
+                    temperature=self.temperature_fenics,
+                    function_space=function_space_value,
+                    t=self.t,
+                )
+
     def create_formulation(self):
         """Creates the formulation of the model"""
         if len(self.sources) > 1:
```
RemDelaporteMathurin, line 517: I think this should go

festim/hydrogen_transport_problem.py
```diff
@@ -503,11 +535,14 @@ def create_formulation(self):
...
-                #     # f = Constant(my_mesh.mesh, (PETSc.ScalarType(0)))
-                #     if source.species == spe:
-                #         formulation += source * v * self.dx
+                for source in self.sources:
+                    for source_spe in source.species:
+                        for source_vol in source.volume:
+                            if source_spe == spe and source_vol == vol:
+                                self.formulation -= (
+                                    source.value_fenics * v * self.dx(vol.id)
+                                )
```
RemDelaporteMathurin, line 542: This can go outsie the main species loop and just iterate through the the sources. Something like:

```python
for source in self.sources:
    for source_spe in source.species:
        v = source_spe.test_function
        for source_vol in source.volume:
            self.formulation -= (
                source.value_fenics * v * self.dx(source_vol.id)
            )
```

festim/source.py
```diff
@@ -0,0 +1,198 @@
...
+
+    @volume.setter
+    def volume(self, value):
+        # check that volume is festim.VolumeSubdomain1D or list of festim.VolumeSubdomain1D
+        if not isinstance(value, (F.VolumeSubdomain1D, int)) and not isinstance(
+            value, list
+        ):
+            raise TypeError(
+                "volume must be of type festim.VolumeSubdomain1D or int or a list of festim.VolumeSubdomain1D or int"
+            )
```
RemDelaporteMathurin, line 59: I'd have all this after the `isinstance(value, list)` check and just have an elif statement

festim/source.py
```diff
@@ -0,0 +1,198 @@
...
+        # check that all elements of list are festim.VolumeSubdomain1D
+        if isinstance(value, list):
+            for element in value:
+                if not isinstance(element, (F.VolumeSubdomain1D, int)):
+                    raise TypeError(
+                        "volume must be of type festim.VolumeSubdomain1D or int or a list of festim.VolumeSubdomain1D or int"
+                    )
+        # if volume is festim.VolumeSubdomain1D, convert to list
+        if not isinstance(value, list):
```
RemDelaporteMathurin, line 68:
```suggestion
        else:
```

festim/source.py
```diff
@@ -0,0 +1,198 @@
...
+            value = [value]
```
RemDelaporteMathurin, line 69: I'm not sure if we should do this in the setter as it modifies the user input quite a lot... Maybe we can do this in HTransportProblem instead

festim/source.py
```diff
@@ -0,0 +1,198 @@
...
+        # check that all elements of list are festim.Species
+        if isinstance(value, list):
+            for element in value:
+                if not isinstance(element, (F.Species, str)):
+                    raise TypeError(
+                        "species must be of type festim.Species or str or a list of festim.Species or str"
+                    )
+        # if species is festim.Species, convert to list
+        if not isinstance(value, list):
+            value = [value]
```
RemDelaporteMathurin, line 93: Similar to comment above

festim/hydrogen_transport_problem.py
```diff
@@ -480,10 +481,39 @@ def create_dirichletbc_form(self, bc):
 
         return form
 
+    def define_sources(self):
+        """Create a fenics object value for each source of the model"""
+        for source in self.sources:
+            for idx, spe in enumerate(source.species):
```
RemDelaporteMathurin, line 487: The approach taken here is to always have the source.species (same for volume) as a list. However this is modifying the user input quite a lot.

Instead, I think it would be better to do this modification here.

Something like:

```python
if not isinstance(source.species, list):
    source_species = [source.species]
else:
    source_species = source.species

for idx, spe in enumerate(species):
    ....
```

Does htat make sense?

Same is applicable to volume

festim/hydrogen_transport_problem.py
```diff
@@ -480,10 +481,39 @@ def create_dirichletbc_form(self, bc):
 
         return form
 
+    def define_sources(self):
```
RemDelaporteMathurin, line 484: I wouldn't call this `define_sources` as the sources are already "defined". What this method does is create the fenics value object for the source.

Maybe also true for other methods like `define_boundary_conditions`. What do you think?

RemDelaporteMathurin (changes requested):

festim/hydrogen_transport_problem.py
```diff
@@ -480,10 +481,52 @@ def create_dirichletbc_form(self, bc):
...
+                    source.species[idx] = F.find_species_from_name(spe, self.species)
+
+            # iterate through volumes, find volume if id of volume is given
+            if not isinstance(source.volume, list):
+                source.volume = [source.volume]
+            for idx, vol in enumerate(source.volume):
+                if isinstance(vol, int):
+                    # if name of species is given then replace with species object
+                    source.volume[idx] = F.find_volume_from_id(
+                        source.volume, self.volume_subdomains
```
RemDelaporteMathurin, line 506: source.volume is a list here, so I don't expect this to work,

It would rather be:

```suggestion
                    source.volume[idx] = F.find_volume_from_id(
                        source.volume[idx], self.volume_subdomains
```

festim/source.py
RemDelaporteMathurin: @jhdark this hasn't been resolved

festim/hydrogen_transport_problem.py
```diff
@@ -480,13 +486,56 @@ def create_dirichletbc_form(self, bc):
...
+    def create_source_values_fenics(self):
+        """For each source in the model, check if the species and volumes are lists,
+        create one if not, check if each element in species and volumes are
+        festim.Species and festim.VolumeSubdomain1D, then create the value fenics
+        attribute.
+        """
+        for source in self.sources:
+            # iterate through species, find species if name of species is given
+            if not isinstance(source.species, list):
+                source.species = [source.species]
```
RemDelaporteMathurin, line 498: No, the idea is to _not_ modify the attribute .species!

That's why in my suggestion I created another local variable `source_species` so that the attribute is left untouched

jhdark: What about creating new attributes species_festim and volume_festim and appending the F.Species and F.Volumesubdomain objects to them? Because we will need to access them later on when building the formulation

RemDelaporteMathurin: I think for now we can do this same thing when building the formulation. Let's try an minimise the number of attributes overall

festim/hydrogen_transport_problem.py
```diff
@@ -480,13 +486,32 @@ def create_dirichletbc_form(self, bc):
...
+        for source in self.sources:
+            # create value_fenics for all F.Source objects
+            if isinstance(source, F.Source):
+                function_space_value = None
+                if callable(source.value):
+                    # if bc.value is a callable then need to provide a functionspace
+                    if not self.multispecies:
+                        function_space_value = source.species.sub_function_space
+                    else:
+                        function_space_value = source.species.collapsed_function_space
```
RemDelaporteMathurin, line 500: Since the species is an attribute of the source, could this go inside `source.create_value_fenics()` ?

festim/hydrogen_transport_problem.py
```diff
@@ -502,18 +527,18 @@ def create_formulation(self):
...
-                # add fluxes
-                # TODO implement this
-                # for bc in self.boundary_conditions:
-                #     pass
-                #     if bc.species == spe and bc.type != "dirichlet":
-                #         formulation += bc * v * self.ds
+        # add sources
+        for source in self.sources:
+            self.formulation -= (
+                source.value_fenics * source.species.test_function * self.dx(vol.id)
```
RemDelaporteMathurin, line 533: should it not be `self.dx(source.volume.id)` instead?

jhdark: indeed

festim/hydrogen_transport_problem.py
jhdark: Yeah this is true

jhdark: Actually no, we would still need to test if the model is multispecies or not. Same as for Dirichlet BC

RemDelaporteMathurin (commented):
@jhdark let's merge this and work on MMS asap

PR #641, opened 2023-11-14
Mesh from XDMF files

jhdark:
## Proposed changes

With this change, users can use meshes made in external software as long as they are converted to XDMF, this can be done using [Meshio](https://github.com/nschloe/meshio). @RemDelaporteMathurin perhaps we could add example conversion scripts to the docs.

Usage:
```python 
mesh=F.MeshFromXDMF(
    volume_file="volumes_file.xdmf",
    facet_file="facets_file.xdmf",
)
```

Additional note, the class takes 2 further arguments: `mesh_name` and `meshtags_name`, these are here as the default name for meshes created in dolfinx is "mesh" and "mesh_tags" for meshtags, where as meshes created by meshio puts everything in something called "Grid". 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -172,6 +172,10 @@ class HydrogenTransportProblem:
         return [s for s in self.subdomains if isinstance(s, F.VolumeSubdomain)]
 
     @property
+    def surface_subdomains(self):
+        return [s for s in self.subdomains if isinstance(s, F.SurfaceSubdomain)]
+
+    @property
     def species(self):
         return self._species
 
@@ -403,7 +407,8 @@ class HydrogenTransportProblem:
 
         elif isinstance(self.mesh, F.Mesh1D):
             self.facet_meshtags, self.volume_meshtags = self.mesh.define_meshtags(
-                subdomains=self.subdomains,
+                surface_subdomains=self.surface_subdomains,
+                volume_subdomains=self.volume_subdomains,
             )
         elif isinstance(self.mesh, F.Mesh):
             # FIXME # refer to issue #647
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -172,10 +172,6 @@ class HydrogenTransportProblem:
         return [s for s in self.subdomains if isinstance(s, F.VolumeSubdomain)]
 
     @property
-    def surface_subdomains(self):
-        return [s for s in self.subdomains if isinstance(s, F.SurfaceSubdomain)]
-
-    @property
     def species(self):
         return self._species
 
@@ -408,7 +404,6 @@ class HydrogenTransportProblem:
         elif isinstance(self.mesh, F.Mesh1D):
             self.facet_meshtags, self.volume_meshtags = self.mesh.define_meshtags(
                 subdomains=self.subdomains,
-                volume_subdomains=self.volume_subdomains,
             )
         elif isinstance(self.mesh, F.Mesh):
             # FIXME # refer to issue #647
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -40,10 +40,8 @@ class HydrogenTransportProblem:
         ds (dolfinx.fem.ds): the surface measure of the model
         function_space (dolfinx.fem.FunctionSpace): the function space of the
             model
-        facet_meshtags (dolfinx.mesh.MeshTags): the facet tags of the model
+        facet_meshtags (dolfinx.mesh.MeshTags): the facet meshtags of the model
         volume_meshtags (dolfinx.mesh.MeshTags): the volume meshtags of the
-            model
-        surface_meshtags (dolfinx.mesh.MeshTags): the surface meshtags of the
             model
         formulation (ufl.form.Form): the formulation of the model
         solver (dolfinx.nls.newton.NewtonSolver): the solver of the model
@@ -59,6 +57,10 @@ class HydrogenTransportProblem:
             over domain
         V_DG_1 (dolfinx.fem.FunctionSpace): A DG function space of degree 1
             over domain
+        volume_subdomains (list of festim.VolumeSubdomain): the volume subdomains
+            of the model
+        surface_subdomains (list of festim.SurfaceSubdomain): the surface subdomains
+            of the model
 
 
     Usage:
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -405,40 +405,22 @@ class HydrogenTransportProblem:
             self.facet_meshtags = self.mesh.define_surface_meshtags()
             self.volume_meshtags = self.mesh.define_volume_meshtags()
 
-        else:
-            facet_indices, tags_facets = [], []
+        elif isinstance(self.mesh, F.Mesh1D):
+            self.facet_meshtags, self.volume_meshtags = self.mesh.define_meshtags(
+                subdomains=self.subdomains,
+                volume_subdomains=self.volume_subdomains,
+            )
+        elif isinstance(self.mesh, F.Mesh):
+            # FIXME # refer to issue #647
+            facet_indices = np.array([], dtype=np.int32)
+            facet_tags = np.array([], dtype=np.int32)
+            self.facet_meshtags = meshtags(
+                self.mesh.mesh, self.mesh.fdim, facet_indices, facet_tags
+            )
 
-            # find all cells in domain and mark them as 0
             num_cells = self.mesh.mesh.topology.index_map(self.mesh.vdim).size_local
             mesh_cell_indices = np.arange(num_cells, dtype=np.int32)
-            tags_volumes = np.full(num_cells, 0, dtype=np.int32)
-
-            for sub_dom in self.subdomains:
-                if isinstance(sub_dom, F.SurfaceSubdomain1D):
-                    facet_index = sub_dom.locate_boundary_facet_indices(
-                        self.mesh.mesh, self.mesh.fdim
-                    )
-                    facet_indices.append(facet_index)
-                    tags_facets.append(sub_dom.id)
-                if isinstance(sub_dom, F.VolumeSubdomain1D):
-                    # find all cells in subdomain and mark them as sub_dom.id
-                    entities = sub_dom.locate_subdomain_entities(
-                        self.mesh.mesh, self.mesh.vdim
-                    )
-                    tags_volumes[entities] = sub_dom.id
-
-            # check if all borders are defined
-            if isinstance(self.mesh, F.Mesh1D):
-                self.mesh.check_borders(self.volume_subdomains)
-
-            # dofs and tags need to be in np.in32 format for meshtags
-            facet_indices = np.array(facet_indices, dtype=np.int32)
-            tags_facets = np.array(tags_facets, dtype=np.int32)
-
-            # define mesh tags
-            self.facet_meshtags = meshtags(
-                self.mesh.mesh, self.mesh.fdim, facet_indices, tags_facets
-            )
+            tags_volumes = np.full(num_cells, 1, dtype=np.int32)
             self.volume_meshtags = meshtags(
                 self.mesh.mesh, self.mesh.vdim, mesh_cell_indices, tags_volumes
             )
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -109,8 +109,6 @@ class HydrogenTransportProblem:
         self.formulation = None
         self.bc_forms = []
         self.temperature_fenics = None
-        # self.surface_subdomains = []
-        # self.volume_subdomains = []
 
     @property
     def temperature(self):
@@ -160,19 +158,11 @@ class HydrogenTransportProblem:
 
     @property
     def volume_subdomains(self):
-        values = []
-        for subdom in self.subdomains:
-            if isinstance(subdom, F.VolumeSubdomain):
-                values.append(subdom)
-        return values
+        return [s for s in self.subdomains if isinstance(s, F.VolumeSubdomain)]
 
     @property
     def surface_subdomains(self):
-        values = []
-        for subdom in self.subdomains:
-            if isinstance(subdom, F.SurfaceSubdomain):
-                values.append(subdom)
-        return values
+        return [s for s in self.subdomains if isinstance(s, F.SurfaceSubdomain)]
 
     def initialise(self):
         self.define_function_spaces()
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -175,6 +175,7 @@ class HydrogenTransportProblem:
     def surface_subdomains(self):
         return [s for s in self.subdomains if isinstance(s, F.SurfaceSubdomain)]
 
+    @property
     def species(self):
         return self._species
 
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -41,7 +41,9 @@ class HydrogenTransportProblem:
         function_space (dolfinx.fem.FunctionSpace): the function space of the
             model
         facet_meshtags (dolfinx.mesh.MeshTags): the facet tags of the model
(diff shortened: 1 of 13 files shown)


festim/hydrogen_transport_problem.py
```diff
@@ -107,9 +107,10 @@ def __init__(
         self.facet_meshtags = None
         self.volume_meshtags = None
         self.formulation = None
-        self.volume_subdomains = []
         self.bc_forms = []
         self.temperature_fenics = None
+        # self.surface_subdomains = []
+        # self.volume_subdomains = []
```
RemDelaporteMathurin, line 113:
```suggestion
```

festim/hydrogen_transport_problem.py
```diff
@@ -157,6 +158,22 @@ def temperature_time_dependent(self):
...
+                values.append(subdom)
+        return values
+
+    @property
+    def surface_subdomains(self):
+        values = []
+        for subdom in self.subdomains:
+            if isinstance(subdom, F.SurfaceSubdomain):
+                values.append(subdom)
+        return values
```
RemDelaporteMathurin, line 175: Documentation

festim/hydrogen_transport_problem.py
```diff
@@ -157,6 +158,22 @@ def temperature_time_dependent(self):
     def multispecies(self):
         return len(self.species) > 1
 
+    @property
+    def volume_subdomains(self):
+        values = []
+        for subdom in self.subdomains:
+            if isinstance(subdom, F.VolumeSubdomain):
+                values.append(subdom)
+        return values
```
RemDelaporteMathurin, line 167:
```suggestion
        return [s for s in self.subdomains if isinstance(s, F.VolumeSubdomain)]
```

We can use list comprehension to make this more compact

festim/hydrogen_transport_problem.py
RemDelaporteMathurin, line 175:
```suggestion
        return [s for s in self.subdomains if isinstance(s, F.SurfaceSubdomain)]
```

festim/hydrogen_transport_problem.py
```diff
@@ -368,48 +385,52 @@ def assign_functions_to_species(self):
     def define_markers_and_measures(self):
         """Defines the markers and measures of the model"""
```
RemDelaporteMathurin, line 386: Maybe we can add more details to these docstrings like the fact that we set the facet_meshtags and volume_meshtags attributes as well as dx and ds

festim/mesh/mesh_from_xdmf.py
```diff
@@ -0,0 +1,62 @@
...
+    Attributes:
+        volume_file (str): path to the volume file
+        facet_file (str): path to the facet file
+        mesh_name (str, optional): name of the mesh in the XDMF file. Defaults to "Grid".
+        meshtags_name (str, optional): name of the meshtags in the XDMF file. Defaults to "Grid".
+        mesh (fenics.mesh.Mesh): the mesh
+    """
+
+    def __init__(
+        self, volume_file, facet_file, mesh_name="Grid", meshtags_name="Grid"
```
RemDelaporteMathurin, line 25: I thought you wanted it to default to the fenicsx flags and not the meshio ones?

festim/mesh/mesh_from_xdmf.py
```diff
@@ -0,0 +1,62 @@
...
+        facet_file (str): path to the facet file
+        mesh_name (str, optional): name of the mesh in the XDMF file. Defaults to "Grid".
+        meshtags_name (str, optional): name of the meshtags in the XDMF file. Defaults to "Grid".
+
+    Attributes:
+        volume_file (str): path to the volume file
+        facet_file (str): path to the facet file
+        mesh_name (str, optional): name of the mesh in the XDMF file. Defaults to "Grid".
+        meshtags_name (str, optional): name of the meshtags in the XDMF file. Defaults to "Grid".
+        mesh (fenics.mesh.Mesh): the mesh
```
RemDelaporteMathurin, line 21:
```suggestion
        mesh (fenics.mesh.Mesh): the fenics mesh
```

festim/mesh/mesh_from_xdmf.py
```diff
@@ -0,0 +1,62 @@
...
+        volume_file (str): path to the volume file
+        facet_file (str): path to the facet file
+        mesh_name (str, optional): name of the mesh in the XDMF file. Defaults to "Grid".
+        meshtags_name (str, optional): name of the meshtags in the XDMF file. Defaults to "Grid".
+
+    Attributes:
+        volume_file (str): path to the volume file
+        facet_file (str): path to the facet file
+        mesh_name (str, optional): name of the mesh in the XDMF file. Defaults to "Grid".
+        meshtags_name (str, optional): name of the meshtags in the XDMF file. Defaults to "Grid".
```
RemDelaporteMathurin, line 20: An attribute cannot be optional, and doesn't have default

festim/mesh/mesh_from_xdmf.py
```diff
@@ -0,0 +1,62 @@
...
+        self.facet_file = facet_file
+        self.mesh_name = mesh_name
+        self.meshtags_name = meshtags_name
+
+        volumes_file = XDMFFile(MPI.COMM_WORLD, self.volume_file, "r")
+        mesh = volumes_file.read_mesh(name=f"{self.mesh_name}")
+
+        super().__init__(mesh=mesh)
+
+    def define_surface_markers(self):
```
RemDelaporteMathurin, line 37: This is the name we have in FESTIM1, can we be consistent and use "meshtags" instead of "markers"?

festim/mesh/mesh_from_xdmf.py
```diff
@@ -0,0 +1,62 @@
...
+            dolfinx.MeshTags: the facet meshtags
+        """
+        facets_file = XDMFFile(MPI.COMM_WORLD, self.facet_file, "r")
+        facet_meshtags = facets_file.read_meshtags(
+            self.mesh, name=f"{self.meshtags_name}"
+        )
+
+        return facet_meshtags
+
+    def define_volume_markers(self):
```
RemDelaporteMathurin, line 50: meshtags instead of markers

festim/mesh/mesh_from_xdmf.py
RemDelaporteMathurin, line 25: Just a comment, this doesn't allow to have different meshtags names in the volume file and the facet file

test/test_mesh.py
```diff
@@ -36,3 +39,81 @@ def test_vdim_changes_when_mesh_changes():
...
+@pytest.mark.parametrize("mesh", [mesh_1D, mesh_2D, mesh_3D])
+def test_meshtags_from_xdmf(tmp_path, mesh):
+    """Test that the facet and volume meshtags are read correctly from the mesh XDMF files"""
+    # create mesh functions
+    fdim = mesh.topology.dim - 1
+    vdim = mesh.topology.dim
+
+    # create facet meshtags
+    facet_indices = []
+    for i in range(1):
```
RemDelaporteMathurin, line 53: really? 😆

test/test_mesh.py
```diff
@@ -36,3 +39,81 @@ def test_vdim_changes_when_mesh_changes():
...
+        facet_indices.append(
+            fenics_mesh.locate_entities_boundary(
+                mesh, fdim, lambda x: np.isclose(x[i], 0)
+            )
+        )
+        facet_indices.append(
+            fenics_mesh.locate_entities_boundary(
+                mesh, fdim, lambda x: np.isclose(x[i], 1)
+            )
+        )
```
RemDelaporteMathurin, line 63: can we improve the readibility a bit?
maybe by creating intermediate variables?

test/test_subdomains.py
```diff
@@ -135,3 +135,26 @@ def test_ValueError_rasied_when_volume_ids_are_not_unique():
...
+def test_ValueError_raised_when_id_not_found_in_surface_subdomains():
+    """test that a ValueError is raised when an id is not found in the list of surface subdomains"""
+
+    surfaces = [F.SurfaceSubdomain(id=1)]
+
+    with pytest.raises(ValueError, match="id 3 not found in list of surfaces"):
+        F.find_surface_from_id(3, surfaces)
+
+
+@pytest.mark.parametrize("input", [1, 2, 3, 4])
```
RemDelaporteMathurin, line 149: maybe no need for a parametrized test here as a simple for loop is enough.

Plus if would be interesting to test a case where the subdomain ids in `surfaces` are not ordered like:

```surfaces = [ surf_2, surf_1,surf_4, surf_3]```

**Review by RemDelaporteMathurin — CHANGES_REQUESTED** [review]

**Comment by RemDelaporteMathurin** [comment]
> perhaps we could add example conversion scripts to the docs.

I think what's even better is to have a whole tutorial for how to make meshes with GMSH, SALOME, Cubit, and how to convert them with meshio.

File: festim/mesh/mesh_from_xdmf.py (hunk 5 above)
Comment (jhdark) [reply festim/mesh/mesh_from_xdmf.py:25]: Given most users will be using this class with meshes generated by meshio though, is it not better to have those as default?

**Comment by jhdark** [comment]
> > perhaps we could add example conversion scripts to the docs.
> 
> I think what's even better is to have a whole tutorial for how to make meshes with GMSH, SALOME, Cubit, and how to convert them with meshio.

agreed

File: festim/hydrogen_transport_problem.py [hunk 13]
```diff
@@ -385,51 +396,56 @@ def assign_functions_to_species(self):
...
+            if isinstance(self.mesh, F.Mesh1D):
+                self.mesh.check_borders(self.volume_subdomains)
 
-        # dofs and tags need to be in np.in32 format for meshtags
-        facet_indices = np.array(facet_indices, dtype=np.int32)
-        tags_facets = np.array(tags_facets, dtype=np.int32)
+            # check volume ids are unique
+            vol_ids = [vol.id for vol in self.volume_subdomains]
+            if len(vol_ids) != len(np.unique(vol_ids)):
+                raise ValueError("Volume ids are not unique")
```
RemDelaporteMathurin, line 436: Should this also be done in the case of `MeshFromXDMF` ?

festim/hydrogen_transport_problem.py
```diff
@@ -385,51 +396,56 @@ def assign_functions_to_species(self):
             spe.prev_solution = sub_prev_solution[idx]
             spe.test_function = sub_test_functions[idx]
 
-    def define_markers_and_measures(self):
-        """Defines the markers and measures of the model"""
+    def define_meshtags_and_measures(self):
```
RemDelaporteMathurin, line 399: Note: we need to think of a way for users to provide their own fenics mesh with their own meshtags.

Here we have only two options: tags from XDMF, or 1D mesh. What about users providing a cube mesh but tagging the surfaces themselves?

festim/hydrogen_transport_problem.py
```diff
@@ -385,51 +396,56 @@ def assign_functions_to_species(self):
...
-                tags_volumes[entities] = sub_dom.id
+            # find all cells in domain and mark them as 0
+            num_cells = self.mesh.mesh.topology.index_map(self.mesh.vdim).size_local
+            mesh_cell_indices = np.arange(num_cells, dtype=np.int32)
+            tags_volumes = np.full(num_cells, 0, dtype=np.int32)
 
-        # check if all borders are defined
-        if isinstance(self.mesh, F.Mesh1D):
-            self.mesh.check_borders(self.volume_subdomains)
+            for sub_dom in self.subdomains:
```
RemDelaporteMathurin, line 415: Linked to my comment above, I think this only applies to `Mesh1D` right?
Should we make it a method of Mesh1D?

festim/hydrogen_transport_problem.py
jhdark: yep

festim/hydrogen_transport_problem.py
```diff
@@ -386,52 +397,39 @@ def assign_functions_to_species(self):
...
-                    self.mesh.mesh, self.mesh.vdim
-                )
-                tags_volumes[entities] = sub_dom.id
+        elif isinstance(self.mesh, F.Mesh1D):
+            self.facet_meshtags, self.volume_meshtags = self.mesh.define_meshtags(
+                subdomains=self.subdomains,
+                volume_subdomains=self.volume_subdomains,
+            )
+        elif isinstance(self.mesh, F.Mesh):
+            # FIXME # refer to issue #647
```
RemDelaporteMathurin, line 414:
```suggestion
            # FIXME # refer to issue #647
            # create empty mesh tags for now
```

festim/mesh/mesh_1d.py
```diff
@@ -19,8 +20,8 @@ class Mesh1D(Mesh):
     def __init__(self, vertices, **kwargs) -> None:
         self.vertices = vertices
 
-        mesh = self.generate_mesh()
-        super().__init__(mesh=mesh, **kwargs)
+        self.mesh = self.generate_mesh()
+        super().__init__(mesh=self.mesh, **kwargs)
```
RemDelaporteMathurin, line 24: the `.mesh` attribute is set in `festim.Mesh`, let's not set it here too:

```suggestion
        mesh = self.generate_mesh()
        super().__init__(mesh=mesh, **kwargs)
```

festim/hydrogen_transport_problem.py
```diff
@@ -164,6 +167,14 @@ def temperature_time_dependent(self):
...
         return len(self.species) > 1
 
+    @property
+    def volume_subdomains(self):
+        return [s for s in self.subdomains if isinstance(s, F.VolumeSubdomain)]
+
+    @property
+    def surface_subdomains(self):
+        return [s for s in self.subdomains if isinstance(s, F.SurfaceSubdomain)]
+
```
RemDelaporteMathurin, line 177: Is this even used anywhere now?

festim/mesh/mesh_1d.py
```diff
@@ -61,3 +62,48 @@ def check_borders(self, volume_subdomains):
             or sorted_borders[-1] != self.vertices[-1]
         ):
             raise ValueError("borders dont match domain borders")
+
+    def define_meshtags(self, subdomains, volume_subdomains):
```
RemDelaporteMathurin, line 66: I don't understand why we need both subdomains and volume_subdomains since one encompasses the other

RemDelaporteMathurin (changes requested):
A few more comments

festim/hydrogen_transport_problem.py
jhdark: The volume subdomains is still used in a few places, surface_subdomains no, could remove that one

festim/mesh/mesh_1d.py
jhdark: True

festim/mesh/mesh_1d.py
```diff
@@ -61,3 +62,49 @@ def check_borders(self, volume_subdomains):
             or sorted_borders[-1] != self.vertices[-1]
         ):
             raise ValueError("borders dont match domain borders")
+
+    def define_meshtags(self, subdomains):
```
RemDelaporteMathurin, line 66: It would simplify the method if we have instead of `subdomains`, `surface_subdomains` and `volume_subdomains`.

```python
        for sub_dom in subdomains:
            if isinstance(sub_dom, F.SurfaceSubdomain1D):
```
would become

```python
        for surf in surface_subdomains:
            ...
        for vol in solume_subdomains:
            ...
```

festim/hydrogen_transport_problem.py
```diff
@@ -164,6 +167,10 @@ def temperature_time_dependent(self):
     def multispecies(self):
         return len(self.species) > 1
 
+    @property
+    def volume_subdomains(self):
+        return [s for s in self.subdomains if isinstance(s, F.VolumeSubdomain)]
+
```
RemDelaporteMathurin, line 173: See my comment below, I think we can keep both `volume_subdomains` and `surface_subdomains` but instead adapt the arguments of `Mesh1D.define_meshtags`

RemDelaporteMathurin (changes requested):
Just a small refactor of `Mesh1D.define_meshtags` and we're good to go

PR #646, opened 2023-11-16
Formulation well defined

jhdark:
## Proposed changes

In the case of a steady-state simulation, the solver will break if every species does not have a definition in every volume subdomain within the model. This fix adds a term to the formulation for each species not defined with a given volume subdomain in a steady-state case.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Black formatted

## Further comments

### DIFF
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -579,12 +579,33 @@ class HydrogenTransportProblem:
                 * self.dx(source.volume.id)
             )
 
-            # add fluxes
-            # TODO implement this
-            # for bc in self.boundary_conditions:
-            #     pass
-            #     if bc.species == spe and bc.type != "dirichlet":
-            #         formulation += bc * v * self.ds
+        # add fluxes
+        # TODO implement this
+        # for bc in self.boundary_conditions:
+        #     pass
+        #     if bc.species == spe and bc.type != "dirichlet":
+        #         formulation += bc * v * self.ds
+
+        # check if each species is defined in all volumes
+        if not self.settings.transient:
+            for spe in self.species:
+                # if species mobile, already defined in diffusion term
+                if not spe.mobile:
+                    not_defined_in_volume = self.volume_subdomains.copy()
+                    for vol in not_defined_in_volume:
+                        # check reactions
+                        for reaction in self.reactions:
+                            if vol == reaction.volume:
+                                not_defined_in_volume.remove(vol)
+                        # check sources
+                        for source in self.sources:
+                            if vol == source.volume:
+                                not_defined_in_volume.remove(vol)
+                    # if species not defined in all volumes, add to formulation
+                    for vol in not_defined_in_volume:
+                        self.formulation += (
+                            spe.solution * spe.test_function * self.dx(vol.id)
+                        )
 
     def create_solver(self):
         """Creates the solver of the model"""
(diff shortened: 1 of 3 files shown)


test/test_h_transport_problem.py
```diff
@@ -617,12 +617,14 @@ def test_run_in_steady_state():
...
     my_vol = F.VolumeSubdomain1D(id=1, borders=[0, 4], material=dummy_mat)
+    H = F.Species("H")
     my_model = F.HydrogenTransportProblem(
         mesh=test_mesh,
         temperature=500,
+        sources=[F.Source(value=1.0, volume=my_vol, species=H)],
         settings=F.Settings(atol=1e-10, rtol=1e-10, transient=False),
         subdomains=[my_vol],
-        species=[F.Species("H")],
+        species=[H],
```
RemDelaporteMathurin, line 627: why?

test/test_system.py
```diff
@@ -0,0 +1,92 @@
...
+            E_k=0.39,
+            p_0=1e13,
+            E_p=0.87,
+            volume=vol_1,
+        ),
+    ]
+
+    my_model.boundary_conditions = [
+        F.DirichletBC(value=1e18, subdomain=left, species=cm)
+    ]
```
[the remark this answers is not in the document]
jhdark: Nope

test/test_h_transport_problem.py
jhdark: Was getting a bug locally, but found the issue. will change it back

festim/hydrogen_transport_problem.py
```diff
@@ -579,12 +579,30 @@ def create_formulation(self):
...
+        if not self.settings.transient:
+            for spe in self.species:
+                # if species mobile, already defined in diffusion term
+                if not spe.mobile:
+                    not_defined_in_volume = self.volume_subdomains.copy()
+                    for vol in self.volume_subdomains:
+                        # check reactions
+                        for reaction in self.reactions:
+                            if vol == reaction.volume:
+                                not_defined_in_volume.remove(vol)
```
RemDelaporteMathurin, line 599: What happened to sources?

Do we have a test for this case:

- 2 subdomains
- 2 species (one mobile, one immobile)
- 1 reaction in subdomain 1 only (involving the two species)
-  1 source term in subdomain 2 for the immobile species only

I think this should be allowed. In the current implementation, you would apply c_2 = 0 in subdomain 2 even though you have a source term there!

festim/hydrogen_transport_problem.py
```diff
@@ -579,12 +579,30 @@ def create_formulation(self):
...
+
+                    # if species not defined in all volumes, add to formulation
```
RemDelaporteMathurin, line 601: add what to formulation?
Can you be more precise here for our future selves
```suggestion
                    # add c = 0 to formulation where needed
```

festim/hydrogen_transport_problem.py
RemDelaporteMathurin: Oh thinking about this more, since this is only for steady state, I think that this case would be ill-posed anyway. Let's keep it like this!

RemDelaporteMathurin (changes requested):
Just a small update in the comment and then we're good to go!
Thanks for this! 🎉

PR #648, opened 2023-11-22
Fix README links

RemDelaporteMathurin:
following the change of ownership of the repo we need to update some links


jhdark (approved):
Looks better, thanks for this!

PR #654, opened 2023-11-28
Adaptive stepsize

RemDelaporteMathurin:
## Proposed changes

This is an attempt at adding adaptive timestepping.

I've taken a [similar approach to MOOSE](https://mooseframework.inl.gov/source/timesteppers/IterationAdaptiveDT.html) where the users give a target number of iterations and a growth factor and a cutback factor.

If the number of newton iterations is larger than the target number, the stepsize is reduced (multiplied by cutback factor). Inversely when it's lower than the target, the stepsize is increased (growth factor).

@jhdark I've noticed we have the concept of the stepsize living in different places.
First we have `HTransportProblem.settings.stepsize` AND `HTransportProblem.dt` (which is just the `fenicsx.Constant` object).
I feel like we should remove this redundancy to improve clarity. Doesn't have to be in this PR but I felt like we could kick off the discussion now

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/stepsize.py
+++ b/festim/stepsize.py
@@ -40,8 +40,9 @@ class Stepsize:
 
     @growth_factor.setter
     def growth_factor(self, value):
-        if value < 1:
-            raise ValueError("growth factor should be greater than one")
+        if value is not None:
+            if value < 1:
+                raise ValueError("growth factor should be greater than one")
 
         self._growth_factor = value
 
@@ -51,8 +52,9 @@ class Stepsize:
 
     @cutback_factor.setter
     def cutback_factor(self, value):
-        if value > 1:
-            raise ValueError("cutback factor should be smaller than one")
+        if value is not None:
+            if value > 1:
+                raise ValueError("cutback factor should be smaller than one")
 
         self._cutback_factor = value
 
--- a/festim/stepsize.py
+++ b/festim/stepsize.py
@@ -24,11 +24,15 @@ class Stepsize:
         initial_value,
     ) -> None:
         self.initial_value = initial_value
-        self.growth_factor = 1.2
-        self.cutback_factor = 0.8
-        self.target_nb_iterations = 4
+        self.growth_factor = None
+        self.cutback_factor = None
+        self.target_nb_iterations = None
 
         # TODO should this class hold the dt object used in the formulation
+
+    @property
+    def adaptive(self):
+        return self.growth_factor or self.cutback_factor or self.target_nb_iterations
 
     def modify_value(self, value, nb_iterations, t=None):
         if not self.is_adapt(t):
--- a/festim/stepsize.py
+++ b/festim/stepsize.py
@@ -7,6 +7,14 @@ class Stepsize:
 
     Args:
         initial_value (float, int): initial stepsize.
+        grwoth_factor (float, optional): factor by which the stepsize is
+            increased when adapting
+        cutback_factor (float, optional): factor by which the stepsize is
+            decreased when adapting
+        target_nb_iterations (int, optional): number of Newton iterations
+            over (resp. under) which the stepsize is increased
+            (resp. decreased)
+
 
     Attributes:
         initial_value (float, int): initial stepsize.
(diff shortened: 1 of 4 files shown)


festim/stepsize.py
```diff
@@ -10,19 +10,74 @@ class Stepsize:
 
     Attributes:
         initial_value (float, int): initial stepsize.
+        growth_factor (float): factor by which the stepsize is
+            increased when adapting
+        cutback_factor (float): factor by which the stepsize is
+            decreased when adapting
+        target_nb_iterations (int): number of Newton iterations
+            over (resp. under) which the stepsize is increased
+            (resp. decreased)
```
jhdark, line 19: Should these not be optional arguments too?

festim/stepsize.py
```diff
@@ -10,19 +10,74 @@ class Stepsize:
...
         self.initial_value = initial_value
+        self.growth_factor = None
+        self.cutback_factor = None
+        self.target_nb_iterations = None
+
+        # TODO should this class hold the dt object used in the formulation
+
+    @property
+    def adaptive(self):
+        return self.growth_factor or self.cutback_factor or self.target_nb_iterations
```
jhdark, line 35: This should also be in the documentation

festim/stepsize.py
```diff
@@ -10,19 +10,74 @@ class Stepsize:
...
+    @cutback_factor.setter
+    def cutback_factor(self, value):
+        if value is not None:
+            if value > 1:
+                raise ValueError("cutback factor should be smaller than one")
+
+        self._cutback_factor = value
+
+    def modify_value(self, value, nb_iterations, t=None):
+        if not self.is_adapt(t):
```
jhdark, line 62: If you already have a property adaptive, why not use it here? As it only needs to be modified if the stepsize is adaptive 
```suggestion
        if not self.adaptive:
```

jhdark (changes requested):
Think we can change a few small things here, but otherwise, its looking good and will be incredibly useful

festim/stepsize.py
RemDelaporteMathurin: Right I understand the confusion. I guess I forgot to show why/how we use `is_adapt(t)`

The idea is that sometimes you want to adapt the timestep at certain times only.

By default, it's always. But say we want to adapt if for t between 0, 100, and 200-1000, etc. then you can override the `is_adapt` method to reflect this behaviour

festim/stepsize.py
```diff
@@ -10,19 +10,74 @@ class Stepsize:
...
         Args:
-            mesh (dolfinx.mesh.Mesh): the domain mesh
+            t (float): the current time
+
         Returns:
-            fem.Constant: the dt value
+            bool: True if needs to adapt, False otherwise.
         """
-        return F.as_fenics_constant(self.initial_value, mesh)
+        return True
```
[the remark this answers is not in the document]
RemDelaporteMathurin: See reply above

festim/stepsize.py
```diff
@@ -10,19 +10,74 @@ class Stepsize:
...
     def __init__(
         self,
         initial_value,
     ) -> None:
         self.initial_value = initial_value
+        self.growth_factor = None
+        self.cutback_factor = None
+        self.target_nb_iterations = None
+
+        # TODO should this class hold the dt object used in the formulation
```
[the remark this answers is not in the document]
RemDelaporteMathurin: If you don't mind I'd rather merge this and then focus on it on another PR

festim/stepsize.py
```diff
@@ -7,22 +7,89 @@ class Stepsize:
 
     Args:
         initial_value (float, int): initial stepsize.
+        grwoth_factor (float, optional): factor by which the stepsize is
```
jhdark, line 10:
```suggestion
        growth_factor (float, optional): factor by which the stepsize is
```

jhdark (approved):
Looks good

PR #657, opened 2023-12-02
Trap convenience class

jhdark:
## Proposed changes

With this update users will be able to give a trap object without having to define a 2 species (immobile and implicit) and a reaction, instead just having a `F.Trap()` object instead and passing it to the new attribute of `F.HydrogenTransporModel`, `traps`. Traps needs to be given as a list like so:

```python
my_trap = F.Trap(
    name="test",
    species=my_mobile_species,
    k_0=1,
    E_k=1,
    p_0=1,
    E_p=1,
    n=1,
    volume=my_vol,
)
my_model.traps = [my_trap]
```

This convenience class should make some simpler hydrogen trapping cases easier for users

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -197,10 +197,7 @@ class HydrogenTransportProblem:
         self._species = value
 
     def initialise(self):
-        # if traps are given, create the necessary objects
-        if len(self.traps) > 0:
-            self.create_species_from_trap()
-
+        self.create_species_from_traps()
         self.define_function_spaces()
         self.define_meshtags_and_measures()
         self.assign_functions_to_species()
@@ -216,13 +213,13 @@ class HydrogenTransportProblem:
         self.create_solver()
         self.initialise_exports()
 
-    def create_species_from_trap(self):
+    def create_species_from_traps(self):
         """Generate a species and reaction per trap defined in self.traps"""
 
         for trap in self.traps:
             trap.create_species_and_reaction()
             self.species.append(trap.trapped_concentration)
-            self.reactions.append(trap.trap_reaction)
+            self.reactions.append(trap.reaction)
 
     def define_temperature(self):
         """Sets the value of temperature_fenics_value. The type depends on
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -26,6 +26,7 @@ class HydrogenTransportProblem:
             conditions of the model
         solver_parameters (dict): the solver parameters of the model
         exports (list of festim.Export): the exports of the model
+        traps (list of F.Trap): the traps of the model
 
     Attributes:
         mesh (festim.Mesh): the mesh of the model
@@ -38,6 +39,7 @@ class HydrogenTransportProblem:
             conditions of the model
         solver_parameters (dict): the solver parameters of the model
         exports (list of festim.Export): the exports of the model
+        traps (list of F.Trap): the traps of the model
         dx (dolfinx.fem.dx): the volume measure of the model
         ds (dolfinx.fem.ds): the surface measure of the model
         function_space (dolfinx.fem.FunctionSpace): the function space of the
@@ -94,6 +96,7 @@ class HydrogenTransportProblem:
         boundary_conditions=[],
         settings=None,
         exports=[],
+        traps=[],
     ) -> None:
         self.mesh = mesh
         self.subdomains = subdomains
@@ -104,6 +107,7 @@ class HydrogenTransportProblem:
         self.boundary_conditions = boundary_conditions
         self.settings = settings
         self.exports = exports
+        self.traps = traps
 
         self.dx = None
         self.ds = None
@@ -179,6 +183,10 @@ class HydrogenTransportProblem:
         self._species = value
 
     def initialise(self):
+        # smth to do with traps here
+        if len(self.traps) >= 1:
+            self.create_species_from_trap()
+
         self.define_function_spaces()
         self.define_markers_and_measures()
         self.assign_functions_to_species()
@@ -193,6 +201,14 @@ class HydrogenTransportProblem:
         self.create_formulation()
         self.create_solver()
         self.initialise_exports()
+
+    def create_species_from_trap(self):
+        """Generate a species and reaction per trap defined in self.traps"""
+
+        for trap in self.traps:
+            trap.create_species_and_reaction()
+            self.species.append(trap.trapped_concentration)
+            self.reactions.append(trap.trap_reaction)
 
     def define_temperature(self):
         """Sets the value of temperature_fenics_value. The type depends on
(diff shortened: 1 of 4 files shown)


jhdark:
@RemDelaporteMathurin for some reason when the final test I've added in `test/test_h_transport_problem` breaks everything else, but when its commented the rest of the tests pass, not sure whats going on here?

RemDelaporteMathurin:
@jhdark I figured out the issue. It is related to the fact that the default value for `reactions` in `HTransportProblem` is `[]`.

When adding a trap object (like in the test), [on this line](https://github.com/jhdark/FESTIM/blob/9caaf122c398fbc4abdad2a5196d7d95e4b94164/festim/hydrogen_transport_problem.py#L222), a reaction is added to self.reactions. However, since `self.reactions` is a list and lists in python are class instances, the (unwanted) result is that a reaction is added to the default value of `self.reactions`.

Consider this where we first create a model with a dummy reaction, then another empty model, and the reaction is carried over.

```python
import festim as F

model1 = F.HydrogenTransportProblem()
model1.reactions.append("test_reaction_as_str")


model2 = F.HydrogenTransportProblem()
print(model2.reactions)  # IT SHOULD BE EMPTY!!
```

RemDelaporteMathurin:
@gabriele-ferrero @SamueleMeschini would you mind having a look at this, at least from the user point of view? Is this implementation fine with you? The idea is to later have even more convenience by having one default mobile species created if nothing is given by the users.

festim/hydrogen_transport_problem.py
```diff
@@ -190,6 +194,10 @@ def species(self, value):
         self._species = value
 
     def initialise(self):
+        # if traps are given, create the necessary objects
+        if len(self.traps) >= 1:
```
RemDelaporteMathurin, line 198:
```suggestion
        if len(self.traps) >0:
```

festim/species.py
```diff
@@ -62,23 +63,60 @@ class Trap(Species):
 
     Args:
         name (str, optional): a name given to the trap. Defaults to None.
+        species (F.Species): the mobile species to be trapped
```
RemDelaporteMathurin, line 66: Should we call this `mobile_species` to be more explicit?

festim/species.py
```diff
@@ -62,23 +63,60 @@ class Trap(Species):
...
+        k_0 (float): the trapping rate constant pre-exponential factor.
+        E_k (float): the trapping rate constant activation energy.
+        p_0 (float): the detrapping rate constant pre-exponential factor.
+        E_p (float): the detrapping rate constant activation energy.
```
RemDelaporteMathurin, line 70: Maybe we should give units

festim/species.py
RemDelaporteMathurin, line 1: The `Trap` class would benefit from more precise documentation. Like specifying this only works for 1 mobile species, only one trapping level, etc.

Maybe even giving the equivalence with explicitly giving the species and reactions

test/test_h_transport_problem.py
```diff
@@ -645,3 +645,25 @@ def test_species_setter():
...
+        p_0=1,
+        E_p=1,
+        n=1,
+        volume=my_vol,
+    )
+    my_model.traps = [my_trap]
+    my_model.create_species_from_trap()
+
+    assert len(my_model.species) == 1
+    assert len(my_model.reactions) == 1
```
RemDelaporteMathurin, line 669: I would also test that the element in `my_model.species` and `my_model.reactions` are the correct type.

These tests would pass if we added `"coucou"` to the list

test/test_species.py
```diff
@@ -115,3 +115,29 @@ def test_implicit_species_concentration_with_no_solution():
     # test that a ValueError is raised when the second species has no solution
     with pytest.raises(ValueError):
         implicit_species.concentration
+
+
+def test_create_species_and_reaction():
+    """test that the trapped_concentration and trap_reaction attribites
```
RemDelaporteMathurin, line 121:
```suggestion
    """test that the trapped_concentration and trap_reaction attributes
```

test/test_h_transport_problem.py
RemDelaporteMathurin: I see you did it in the other test. Nonetheless, it isn't hard to add a type check here too

jhdark:
> > @gabriele-ferrero @SamueleMeschini would you mind having a look at this, at least from the user point of view? Is this implementation fine with you? The idea is to later have even more convenience by having one default mobile species created if nothing is given by the users.
> 
> It seems ok, the notes do a great job explaining the equivalence of the convinience class. A default mobile species would be even better because I think most of the users will do simple H transport.

Yeah a default mobile species will be needed, its just to have a default mobile species, this would require refactoring of a lot of other logic in `hydrogen_transport_problem`, so think we'll have to save that for another PR

gabriele-ferrero:
It seems ok to me, and I think it is a great convenience class to have in festim. I will have a further look when it will pass the checks

RemDelaporteMathurin:
> when it will pass the checks

Yes this should be fixed in #661

festim/hydrogen_transport_problem.py
```diff
@@ -208,6 +216,14 @@ def initialise(self):
         self.create_solver()
         self.initialise_exports()
 
+    def create_species_from_trap(self):
```
RemDelaporteMathurin, line 219: Should we have traps (plural) here instead?

festim/species.py
```diff
@@ -60,25 +61,104 @@ def concentration(self):
...
+        self.E_p = E_p
+        self.n = n
+        self.volume = volume
+
+    def create_species_and_reaction(self):
+        """create the immobile trapped species object and the reaction for trapping"""
+        self.trapped_concentration = F.Species(name=self.name, mobile=False)
+        trap_site = F.ImplicitSpecies(n=self.n, others=[self.trapped_concentration])
+
+        self.trap_reaction = F.Reaction(
```
RemDelaporteMathurin, line 152: I would just call this `reaction`

festim/species.py
```diff
@@ -60,25 +61,104 @@ def concentration(self):
...
+    ) -> None:
         super().__init__(name)
+        self.mobile_species = mobile_species
+        self.k_0 = k_0
+        self.E_k = E_k
+        self.p_0 = p_0
+        self.E_p = E_p
+        self.n = n
+        self.volume = volume
+
```
RemDelaporteMathurin, line 146: Let's initialise all the attributes to `None` here (reaction, trapped_concentration, etc.)

festim/hydrogen_transport_problem.py
```diff
@@ -193,6 +197,10 @@ def species(self, value):
         self._species = value
 
     def initialise(self):
+        # if traps are given, create the necessary objects
+        if len(self.traps) > 0:
```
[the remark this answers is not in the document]
jhdark: Yeah think we can get rid of this

festim/hydrogen_transport_problem.py
jhdark: sure

festim/species.py
jhdark: sure

PR #661, opened 2023-12-04
H transport initialisation refactoring

jhdark:
## Proposed changes

We habe a minor python bug when multiple instances of the same class are within the same script, and with attributes that default to empty lists, if a value is appended to this list, it becomes the default for subsequent instances like so:
```python
model_1 = F.HydrogenTransportProblem()
model_1.reactions.append(1)

model_2 = F.HydrogenTransportProblem()
print(model_2.reactions)
```
gives:
```
[1]
```

The changes presented fixes this by altering the attribute defaults in `HydrogenTransportProblem` from
```python
class HydrogenTransportProblem:
    def __init__(self, attribute=[]):
        self.attribute = attribute
```
to:
```python
class HydrogenTransportProblem:
    def __init__(self, attribute=None):
        self.attribute = attribute

        if self.attribute is None:
            self.attribute = []
```
and this seems to fix it. Found this in a handy pitfall guide: https://docs.python-guide.org/writing/gotchas/

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -100,27 +100,14 @@ class HydrogenTransportProblem:
         exports=None,
     ):
         self.mesh = mesh
-        self.subdomains = subdomains
-        self.species = species
-        self.reactions = reactions
+        self.subdomains = subdomains or []
+        self.species = species or []
+        self.reactions = reactions or []
         self.temperature = temperature
-        self.sources = sources
-        self.boundary_conditions = boundary_conditions
+        self.sources = sources or []
+        self.boundary_conditions = boundary_conditions or []
         self.settings = settings
-        self.exports = exports
-
-        if self.subdomains is None:
-            self.subdomains = []
-        if self.species is None:
-            self.species = []
-        if self.reactions is None:
-            self.reactions = []
-        if self.sources is None:
-            self.sources = []
-        if self.boundary_conditions is None:
-            self.boundary_conditions = []
-        if self.exports is None:
-            self.exports = []
+        self.exports = exports or []
 
         self.dx = None
         self.ds = None
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -90,15 +90,15 @@ class HydrogenTransportProblem:
     def __init__(
         self,
         mesh=None,
-        subdomains=[],
-        species=[],
-        reactions=[],
+        subdomains=None,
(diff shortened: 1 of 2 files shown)


festim/hydrogen_transport_problem.py
```diff
@@ -90,24 +90,24 @@ class HydrogenTransportProblem:
...
+        self.species = species or []
+        self.reactions = reactions or []
         self.temperature = temperature
-        self.sources = sources
-        self.boundary_conditions = boundary_conditions
+        self.sources = sources or []
+        self.boundary_conditions = boundary_conditions or []
         self.settings = settings
-        self.exports = exports
+        self.exports = exports or []
```
RemDelaporteMathurin, line 110: Nice hack, I didn't know of this way of doing things.
Does it work for all python versions?

Also, we may want to add a comment to say what this does for people like me who don't know this

test/test_h_transport_problem.py
```diff
@@ -645,3 +645,14 @@ def test_species_setter():
...
+
+def test_reinstantiation_of_class():
+    """Test that when an attribute defaults to empty list, when the class
+    is reinstantiated the list is not passed to the new class object"""
+
+    model_1 = F.HydrogenTransportProblem()
+    model_1.species.append(F.Species("test"))
+
+    model_2 = F.HydrogenTransportProblem()
+    assert len(model_2.species) == 0
```
RemDelaporteMathurin, line 658: This tests only for species.

You could do:

```suggestion
@pytest.mark.parametrize(
    "attribute, value",
    [
        ("species", F.Species("test")),
        ("reactions", None),
        ("sources", None),
        ("subdomains", None),
        ("boundary_conditions", None),
        ("exports", None),
    ],
)
def test_reinstantiation_of_class(attribute, value):
    """Test that when an attribute defaults to empty list, when the class
    is reinstantiated the list is not passed to the new class object"""

    model_1 = F.HydrogenTransportProblem()
    getattr(model_1, attribute).append(value)

    model_2 = F.HydrogenTransportProblem()
    assert len(getattr(model_2, attribute)) == 0
```

To be more thorough.

RemDelaporteMathurin (commented):
This looks great to me, thanks for fixing this. It is a common pitfall in python!

Just have a few comments for future proofing 🚀

RemDelaporteMathurin (approved):

Issue #659, opened 2023-12-03
Initial temperature from XDMF for transient simulations


KulaginVladimir:
According to the [discussion](https://festim.discourse.group/t/how-to-set-initial-temperature-distribution-from-xdmf/17), currently, there is no possibility to set the `initial_value` for `HeatTransferProblem` from a XDMF file.

KulaginVladimir:
@RemDelaporteMathurin. So, I came up with the following modifications for the `create_functions` method of the `HeatTransferProblem` class:

```python
        if self.transient:
            if isinstance(self.initial_value, festim.TemperatureFromXDMF):
                f.XDMFFile(self.initial_value.filename).read_checkpoint(
                    self.T, self.initial_value.label, -1
                )
                self.T_n.assign(self.T)
            else:
                ccode_T_ini = sp.printing.ccode(self.initial_value)
                self.initial_value = f.Expression(ccode_T_ini, degree=2, t=0)
                self.T_n.assign(f.interpolate(self.initial_value, V))
``` 
Thus, one could provide `F.TemperatureFromXDMF(filename = '...', label = '...')` as an input. Or, maybe, it should be done the other way?

RemDelaporteMathurin:
@KulaginVladimir I don't think we want to use the `TemperatureFromXDMF` class for this as we would like users to keep using the `InitialCondition` class.

Instead of `if isinstance(self.initial_value, festim.TemperatureFromXDMF)` we could simply have:

```python
if isinstance(self.initial_value, str):
    if self.initial_value.endswith(".xdmf"):
        ....
        # read from XDFMFile and 
```

Does that make sense?

KulaginVladimir:
@RemDelaporteMathurin
> I don't think we want to use the TemperatureFromXDMF class for this as we would like users to keep using the InitialCondition class.

Sorry, but I didn't really get it 😄. Do you suggest to pass `F.InitialCondition` directly to `F.HeatTransferProblem`? If so, the temperature field can be read from XDMF based on `value` (string), `label` and `time_step` attributes of `F.InitialCondition`, but, I suppose, it would require to check that `initial_value` of `F.HeatTransferProblem` is of `InitialCondition` class (similar to my previous comment). Maybe. I am wrong.

The `label` is reqired to read from XDMF 👇 , isn't it?
```python
if isinstance(self.initial_value, str):
    if self.initial_value.endswith(".xdmf"):
        ....
        # read from XDFMFile and 
``` 

[...]

PR #662, opened 2023-12-05
InitialCondition for HeatTransferProblem

KulaginVladimir:
## Proposed changes

According to #659, this PR attempts to enhance the functionality of initial condition for `HeatTransferProblem`. Additionally, the tests were updated to follow new changes.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/temperature/temperature_solver.py
+++ b/festim/temperature/temperature_solver.py
@@ -8,7 +8,7 @@ class HeatTransferProblem(festim.Temperature):
     Args:
         transient (bool, optional): If True, a transient simulation will
             be run. Defaults to True.
-        initial_value (sp.Add, float, optional): The initial value.
+        initial_value (festim.InitialCondition, optional): The initial value.
             Only needed if transient is True. Defaults to 0.
         absolute_tolerance (float, optional): the absolute tolerance of the newton
             solver. Defaults to 1e-03
@@ -25,7 +25,7 @@ class HeatTransferProblem(festim.Temperature):
     Attributes:
         F (fenics.Form): the variational form of the heat transfer problem
         v_T (fenics.TestFunction): the test function
-        initial_value (sp.Add, int, float): the initial value
+        initial_value (festim.InitialCondition): the initial value
         sub_expressions (list): contains time dependent fenics.Expression to
             be updated
         sources (list): contains festim.Source objects for volumetric heat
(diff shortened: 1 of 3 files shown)


RemDelaporteMathurin (commented):
Thank you very much @KulaginVladimir again for this contribution!

Just a comment for more consistency across the code base.

Also re-running the test cause apparently the upload to Codecov failed...

RemDelaporteMathurin (changes requested):
Right now that Codecov has run, we are missing tests for reading InitialCondition from XDMF files.

To do so, you can try and first export a function to XDMF then use the filename in an InitialCondition

festim/temperature/temperature_solver.py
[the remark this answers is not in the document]
KulaginVladimir: @RemDelaporteMathurin, sounds more pragmatic.

Regarding code coverage, I can try to add new test, but where it will be better to put it? Similar tests for `HeatTransferProblem` are in `test_temperature` (unit folder) and `test_solving` (heat_transfer_problem folder)

festim/temperature/temperature_solver.py
```diff
@@ -36,15 +36,15 @@ class HeatTransferProblem(festim.Temperature):
     def __init__(
         self,
         transient=True,
-        initial_value=0.0,
+        initial_condition=None,
```
KulaginVladimir, line 39: Among other thing. I'm not sure what `initial_condition` should be set as default, if `transient=True`.

festim/temperature/temperature_solver.py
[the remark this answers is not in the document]
RemDelaporteMathurin: I'd put it in test_temperature (unit folder)

festim/temperature/temperature_solver.py
KulaginVladimir: @RemDelaporteMathurin, both suggestions are included in the latest commit

festim/temperature/temperature_solver.py
```diff
@@ -75,9 +75,19 @@ def create_functions(self, materials, mesh, dt=None):
...
-            self.T_n.assign(f.interpolate(self.initial_value, V))
+            if isinstance(self.initial_condition.value, str):
+                if self.initial_condition.value.endswith(".xdmf"):
+                    with f.XDMFFile(self.initial_condition.value) as file:
+                        file.read_checkpoint(
+                            self.T,
+                            self.initial_condition.label,
+                            self.initial_condition.time_step,
+                        )
+                    self.T_n.assign(self.T)
```
RemDelaporteMathurin, line 86: I think this can be simplified by simply reading `self.T_n` 

```suggestion
                            self.T_n,
                            self.initial_condition.label,
                            self.initial_condition.time_step,
                        )
```

RemDelaporteMathurin (changes requested):
just a tiny comment and then we're good to go!

festim/temperature/temperature_solver.py
KulaginVladimir: @RemDelaporteMathurin, sure, but it will also require to change new test `test_heat_transfer_create_functions_transient`. I suppose, just removing lines 127-130 would be fine and won't affect the code coverage.

festim/temperature/temperature_solver.py
```diff
@@ -75,9 +75,18 @@ def create_functions(self, materials, mesh, dt=None):
         self.v_T = f.TestFunction(V)
 
         if self.transient:
```
RemDelaporteMathurin, line 77: I think that if `transient` is `True` and `initial_condition` is `None` , then the `else` on line 86 will be triggered. 

Maybe here we need to check that `self.initial_condition` is not `None` :

```suggestion
        if self.transient and self.initial_condition:
```

RemDelaporteMathurin (approved):
Thanks @KulaginVladimir ! This had been missing for a long long time! 🎉

PR #665, opened 2023-12-11
new vertices setter

JonathanGSDUFOUR:
## Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Black formatted

## Further comments

### DIFF
--- a/festim/mesh/mesh_1d.py
+++ b/festim/mesh/mesh_1d.py
@@ -17,11 +17,19 @@ class Mesh1D(F.Mesh):
         vertices (list): the mesh x-coordinates (m)
     """
 
-    def __init__(self, vertices, **kwargs) -> None:
+    def __init__(self, vertices, **kwargs) -> None:        
         self.vertices = vertices
 
         mesh = self.generate_mesh()
         super().__init__(mesh=mesh, **kwargs)
+    
+    @property
+    def vertices(self):
+        return self._vertices
+
+    @vertices.setter
+    def vertices(self, value):
+        self._vertices = sorted(np.unique(value))
 
     def generate_mesh(self):
         """Generates a 1D mesh"""


test/test_mesh.py
```diff
@@ -120,3 +120,10 @@ def test_meshtags_from_xdmf(tmp_path, mesh):
     assert volume_meshtags.values.all() == my_model.volume_meshtags.values.all()
     assert facet_meshtags.dim == my_model.facet_meshtags.dim
     assert facet_meshtags.values.all() == my_model.facet_meshtags.values.all()
+
+
+def test_mesh_vertices_from_list():
+    """Check that giving a list as input works"""
```
jhdark, line 126:
```suggestion
    """Check that giving vertices as a list is correctly processed"""
```

festim/mesh/mesh_1d.py
jhdark, line 1: You can update the documentation too

jhdark (commented):
Just some small changes to make

jhdark:
Thanks Jonathon for your first contribution to festim !!! Hopefully the first of many! Any other changes from you @RemDelaporteMathurin

festim/mesh/mesh_1d.py
```diff
@@ -11,10 +11,10 @@ class Mesh1D(F.Mesh):
     1D Mesh
 
     Args:
-        vertices (list): the mesh x-coordinates (m)
+        vertices (list or np.array): the mesh x-coordinates (m)
```
RemDelaporteMathurin, line 14: The correct type is `np.ndarray`

```suggestion
        vertices (list or np.ndarray): the mesh x-coordinates (m)
```

festim/mesh/mesh_1d.py
```diff
@@ -11,10 +11,10 @@ class Mesh1D(F.Mesh):
...
 
     Attributes:
-        vertices (list): the mesh x-coordinates (m)
+        vertices (np.array): the mesh x-coordinates (m)
```
RemDelaporteMathurin, line 17:
```suggestion
        vertices (np.ndarray): the mesh x-coordinates (m)
```

festim/mesh/mesh_1d.py
```diff
@@ -23,6 +23,14 @@ def __init__(self, vertices, **kwargs) -> None:
         mesh = self.generate_mesh()
         super().__init__(mesh=mesh, **kwargs)
 
+    @property
+    def vertices(self):
+        return self._vertices
+
+    @vertices.setter
+    def vertices(self, value):
+        self._vertices = np.sort(np.unique(np.array(value, dtype=np.float64)))
```
RemDelaporteMathurin, line 32: This looks a bit redundant as np.unique returns a `np.ndarray`
Why enforcing `dtype` `np.float64` ?
```suggestion
        self._vertices = np.sort(np.unique(value))
```

test/test_mesh.py
```diff
@@ -120,3 +120,10 @@ def test_meshtags_from_xdmf(tmp_path, mesh):
     assert volume_meshtags.values.all() == my_model.volume_meshtags.values.all()
     assert facet_meshtags.dim == my_model.facet_meshtags.dim
     assert facet_meshtags.values.all() == my_model.facet_meshtags.values.all()
+
+
+def test_mesh_vertices_from_list():
+    """Check that giving vertices as a list is correctly processed"""
+    my_mesh = F.Mesh1D(vertices=[1, 2, 3, 4])
+
+    assert isinstance(my_mesh.vertices, np.ndarray)
```
RemDelaporteMathurin, line 129: This only tests vertices as integers, I would test floats too

test/test_mesh.py
```diff
@@ -120,3 +120,10 @@ def test_meshtags_from_xdmf(tmp_path, mesh):
     assert volume_meshtags.values.all() == my_model.volume_meshtags.values.all()
     assert facet_meshtags.dim == my_model.facet_meshtags.dim
     assert facet_meshtags.values.all() == my_model.facet_meshtags.values.all()
+
+
+def test_mesh_vertices_from_list():
+    """Check that giving vertices as a list is correctly processed"""
```
RemDelaporteMathurin, line 126: "correctly processed" isn't very explicit.

Here, you are specifically checking that the `vertices` attributes has the correct type `np.ndarray`

RemDelaporteMathurin (changes requested):
Hi @JonathanGSDufour! Congrats for this first contribution! 🎉 

A couple of comments on my side.

Also the coverage doesn't seem to be uploaded to Codecov.

@jhdark we may have to update the uploader in the github actions workflow.

festim/mesh/mesh_1d.py
jhdark: Using this allows users to have either a list of floats or ints, as if we remove these and a list of ints is given, a RuntimeError is raised

RemDelaporteMathurin: I don't seem to have an issue. See below

```
>>> import numpy as np
>>> a = [1, 2, 3]
>>> np.unique(a)
array([1, 2, 3])
>>> np.sort(np.unique(a))
array([1, 2, 3])
```

JonathanGSDUFOUR: It seems that putting only np.sort(np.unique(value)) is not enough and retrun an error in the tests :
=========================== short test summary info ============================
FAILED test/test_mesh.py::test_mesh_vertices_from_list - RuntimeError: Unsupported mesh dtype: int64

RemDelaporteMathurin: In that case I would do something like:

```suggestion
        self._vertices = np.sort(np.unique(value)).astype(float)
```

test/test_mesh.py
```diff
@@ -120,3 +120,14 @@ def test_meshtags_from_xdmf(tmp_path, mesh):
...
+
+def test_mesh_vertices_from_list():
+    """Check that giving vertices as a list is correctly processed and ends up as a np.ndarray for the mesh"""
+    my_mesh = F.Mesh1D(vertices=[1, 2, 3, 4])
+
+    assert isinstance(my_mesh.vertices, np.ndarray)
+
+    my_mesh_2 = F.Mesh1D(vertices=[0, 0.1, 0.2, 0.3, 0.4, 0.5])
+
+    assert isinstance(my_mesh_2.vertices, np.ndarray)
```
RemDelaporteMathurin, line 133: Nice! However if the first assert statement fails, then the second won't be run.

To avoid this, we use the parametrize decorator. More details [here](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions).

Here, it would be something like:

```python
import pytest

@pytest.mark.parametrize("vertices", [[1, 2, 3, 4], [0, 0.1, 0.2, 0.3, 0.4, 0.5]])
def test_mesh_vertices_from_list(vertices):
    """Check that giving vertices as a list is correctly processed and ends up as a np.ndarray for the mesh"""
    my_mesh = F.Mesh1D(vertices=vertices)

    assert isinstance(my_mesh.vertices, np.ndarray)
```

RemDelaporteMathurin:
@JonathanGSDUFOUR here's a tip to format your scripts.

Install black

```
pip install black
```

Then run:

```
black .
```

This will format all the files in the current directory.

Alternatively you can also install the black extension in Visual Studio Code (if you're using it)

RemDelaporteMathurin:
@JonathanGSDUFOUR right I don't know why the linting is not passing... Maybe different versions of black? the workflow runs with black 23.12.0, what is your black version?

JonathanGSDUFOUR:
When I managed to use black it was through pip install black in terminal, but it uses the 19.10b0 version, and when trying to use VSCode extension it doesn't seem to work (the extension is on version 23.10.1)

RemDelaporteMathurin:
This is a [very old version of black](https://github.com/psf/black/releases/tag/19.10b0) (2019!!)

Can you try to upgrade it with

```
pip install -U black
```

festim/mesh/mesh_1d.py
RemDelaporteMathurin: @JonathanGSDUFOUR it looks like this slipped! 😄

JonathanGSDUFOUR: Oh yeah sorry, not used to git 😅  yet 😉

RemDelaporteMathurin (approved):
This looks good to me! Thanks @JonathanGSDUFOUR and congrats on your first contribution! 💯 

@jhdark feel free to merge this

PR #668, opened 2023-12-12
Fix codecov upload

RemDelaporteMathurin:
This PR fixes the upload to codecov which randomely failed.
After a long discussion with the codecov team https://github.com/codecov/feedback/issues/180 we found out that the problem was that [the CODECOV_TOKEN secret wasn't passed correctly](https://github.com/festim-dev/FESTIM/actions/runs/7182795041/job/19560097608#step:7:59).

The solution was to define the secret at the job level and not at the step level.

I reported this bug to Github https://github.com/orgs/community/discussions/78847

NOTE: a similar change will have to be done on the main branch

### DIFF
--- a/.github/workflows/ci_conda.yml
+++ b/.github/workflows/ci_conda.yml
@@ -10,7 +10,7 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Echo secret
-      run: echo ${{ secrets.TEST_SECRET }}
+      run: echo $TEST_SECRET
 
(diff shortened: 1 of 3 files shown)


jhdark (approved):
Nice little fix, looks good to me

RemDelaporteMathurin:
Well it didn't fix it correctly since it still seems to fail (just look at the last commit it started not finding the token again...)

Issue #675, opened 2023-12-23
`TXTExports` & `TXTExport` features


KulaginVladimir:
`TXTExport` & `TXTExports` classes can be modified based on two points:

1. Following the [discourse topic](https://festim.discourse.group/t/file-formatting-bug-in-the-task4-workshop-file/25/4), users might need to provide the desired times for `TXTExport` & `TXTExports` in `Stepsize.milestones` to ensure that data is actually exported at these times. As @RemDelaporteMathurin suggested, this situation can be improved, if the missing `TXTExport.times` are added to `Stepsize.milestones` during initialisation of the model. 

2. If I'm not mistaken, the current `TXTExports` class won't export anything if a user doesn't provide `TXTExports.times`. However, the `TXTExport` class alows users to export the needed data at all timesteps, if `TXTExport.times` are not provided. Therefore, the `TXTExports` can be modified a bit to allow users to export the results at all timesteps. I suppose, it can be done as follows:

```python
class TXTExports:
    def __init__(self, fields=[], labels=[], times=None, folder=None) -> None:
        self.fields = fields
        if len(self.fields) != len(labels):
            raise ValueError(
                "Number of fields to be exported "
                "doesn't match number of labels in txt exports"
            )
        if times is not None:
            self.times = sorted(times)
        self.labels = labels
        self.folder = folder
        self.exports = []
        for function, label in zip(self.fields, self.labels):
            self.exports.append(TXTExport(function, label, folder, times))
```

RemDelaporteMathurin:
@KulaginVladimir thanks for kicking off the discussion.

About the first point, I agree we should ensure that if users don't provide milestones but provide times for the TXTExport then we need to add these times to the milestones (and maybe raise a warning or something).

About the second point, I don't know if we want to keep supporting for `TXTExports` (plural). I don't know if it's used a lot or not. The only added value is to be able to define `times` and `folder` once. On the other hand I think your fix can work. So maybe we can implement it and add a `DeprecationWarning` specifying we will deprecate this class in the next minor release? what do you think?

KulaginVladimir:
@RemDelaporteMathurin 

> About the second point, I don't know if we want to keep supporting for TXTExports (plural). I don't know if it's used a lot or not. The only added value is to be able to define times and folder once. On the other hand I think your fix can work. So maybe we can implement it and add a DeprecationWarning specifying we will deprecate this class in the next minor release? what do you think?

[...]

Issue #672, opened 2023-12-13
Formatting Issues With Exporting Mobile Concentration Files in Workshop task1


ee-nn:
I am trying out FESTIM for the first time and am starting to work my way through the workshop notebooks. I am running version 1.0 in Google Colab, and after figuring out how to get FEniCS installed I was able to run cells 1-8 with no problem. However, upon trying to retrieve the data for plotting I get the following error: 
```
import matplotlib.pyplot as plt
import numpy as np

data = np.genfromtxt(results_folder + "/mobile_concentration_100s.txt")

plt.plot(data[:,0], data[:,1])
plt.show()

---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-14-c1e47a4b847f> in <cell line: 4>()
      2 import numpy as np
      3 
----> 4 data = np.genfromtxt(results_folder + "/mobile_concentration_100s.txt")
      5 
      6 plt.plot(data[:,0], data[:,1])

/usr/local/lib/python3.10/dist-packages/numpy/lib/_datasource.py in open(self, path, mode, encoding, newline)
    531                                       encoding=encoding, newline=newline)
    532         else:
--> 533             raise FileNotFoundError(f"{path} not found.")
    534 
    535 

FileNotFoundError: /content/drive/My Drive/FESTIM/task1/mobile_concentration_100s.txt not found.
```
I looked at the naming of the output file and there seems to be a formatting error for `mobile_concentration_*.txt`. In particular, there is a floating point precision error when appending the timestamp to the end of the filename it is `mobile_concentration_99.9999999999986s.txt`. 

I have corresponded with Remi on this and he says that this is a quick fix that will be addressed in the next version of the code. In the meantime, one can get around the issue by either installing a different version, as provided by Remi: 
> Just fyi, you can install any released version of the code like:
> 
> pip install festim==1.0
> 
> but also a github branch:
> 
> pip install git+https://github.com/festim-dev/festim@main
> pip install git+https://github.com/festim-dev/festim@fenicsx
> 
> but even the version of the code at a specific commit:
> 
> pip install git+https://github.com/festim-dev/festim@56089a5
Of course, this opens the possibility of other bugs, but might help get around this formatting issue for now until a stable version of FESTIM with this bug fix comes out. 

Hope this helps anyone else for which this was causing confusion!

RemDelaporteMathurin:
A solution would be to either:


1. replace this line by

```python
[...]

PR #676, opened 2024-01-07
TXTExport features

KulaginVladimir:
## Proposed changes

This PR implements several minor modifications to `TXTExport` & `TXTExports` classes:
1. Based on #672, `TXTExport` was modified to accept a user-defined format for column headers in the output files.
2. Based on #675:
     - At the initialisation stage, `TXTExport.times` are passed to `Stepsize.milestones` (if they are not present) in order to ensure that data is actually exported at these times.
     - `TXTExports` was modified to allow users export data at all timesteps. Also, a deprication warning was added.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/exports/txt_export.py
+++ b/festim/exports/txt_export.py
@@ -18,9 +18,11 @@ class TXTExport(festim.Export):
         times (list, optional): if provided, the field will be
             exported at these timesteps. Otherwise exports at all
             timesteps. Defaults to None.
+        header_format (str, optional): the format of column headers.
+            Defautls to ".2e".
     """
 
-    def __init__(self, field, label, folder, times=None) -> None:
+    def __init__(self, field, label, folder, times=None, header_format=".2e") -> None:
         super().__init__(field=field)
         if times:
             self.times = sorted(times)
@@ -28,6 +30,7 @@ class TXTExport(festim.Export):
             self.times = times
         self.label = label
         self.folder = folder
+        self.header_format = header_format
         self._first_time = True
 
     @property
@@ -60,7 +63,7 @@ class TXTExport(festim.Export):
             if steady:
                 header = "x,t=steady"
             else:
-                header = "x,t={}s".format(current_time)
+                header = f"x,t={format(current_time, self.header_format)}s"
 
             # if the directory doesn't exist
             # create it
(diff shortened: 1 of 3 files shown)


jhdark:
This could be a good opportunity to add in `filename` as a parameter of the `TXTExport` class same as the 'XDMFExport' class, rather than just as a property. What do you think @RemDelaporteMathurin ? Even if not, the doc strings of the class should be updated to include the attributes.

RemDelaporteMathurin:
@jhdark that's an excellent idea! We didn't have a file name initially since it was defined from the label (and we had several files per export) but now we can definitely have a file name argument! 

@KulaginVladimir do you think you could add this?

test/unit/test_exports/test_txt_exports.py
RemDelaporteMathurin, line 1: Maybe we need a new test that checks that, if no milestones are given, when no simulation milestones are given and txt exports times are given then the export times are added to milestones upon calling `initialise`

festim/generic_simulation.py
```diff
@@ -273,6 +274,20 @@ def initialise(self):
...
+            if isinstance(export, festim.TXTExport) and export.times:
+                if not self.dt.milestones:
+                    self.dt.milestones = []
+                for time in export.times:
+                    if time not in self.dt.milestones:
+                        msg = "To ensure that TXTExport exports data at the desired times "
+                        msg += "TXTExport.times are added to milestones"
+                        warnings.warn(msg)
+                        self.dt.milestones.append(time)
+                self.dt.milestones.sort()
```
RemDelaporteMathurin, line 289: As per my comment below, we may need a test that explicitly checks these lines. In other words, checking that `TXTExport` integrates well with `Simulation`

festim/exports/txt_export.py
```diff
@@ -60,7 +63,7 @@ def write(self, current_time, steady):
             if steady:
                 header = "x,t=steady"
             else:
-                header = "x,t={}s".format(current_time)
+                header = f"x,t={format(current_time, self.header_format)}s"
```
RemDelaporteMathurin, line 66: For more clarity, shouldn't these lines be after line 77?

RemDelaporteMathurin (changes requested):
Hi @KulaginVladimir a few comments. Also as @jhdark said maybe we could get rid of `label` and simply expose `filename` just like we do for other exports.

festim/generic_simulation.py
KulaginVladimir: Done

RemDelaporteMathurin (approved):
This looks great to me! Thanks @KulaginVladimir 

I'll let @jhdark review this before merging since he suggested the `filename` property

jhdark (approved):
Looks good to me

PR #679, opened 2024-01-16
Radioactive decay

### LINKED ISSUE #488: Radioactive decay
It'd be nice to have support for radioactive decay.

I guess it can be done externally in the meantime

RemDelaporteMathurin:
## Proposed changes

This PR implements radioactive decay (fixes #488).

The new formulation is:

$$
    \frac{\partial c_m}{\partial t} =  \nabla \cdot (D \nabla c_m)  - \lambda c_m - \sum \left( -k_i \ c_m \ (n_i - c_{t,i}) + p_i \ c_{t,i} \right)
$$

$$
    \frac{\partial c_{t,i}}{\partial t}=k_i \ c_m \ (n_i - c_{t,i})- p_i \ c_{t,i} - \lambda c_{t,i}
$$

where $\lambda$ is the decay constant (expressed in s^-1).

It is implemented as a volumetric source object inheriting from `F.Source`.
It takes a decay constant parameter.

**Usage**:

```python
import festim as F
import numpy as np


my_model = F.Simulation()

my_model.traps = [
    F.Trap(0, 1, 0, 3, materials=my_model.materials.materials[0], density=100)
]

half_life = 50  # seconds
decay_constant = np.log(2) / half_life

my_model.sources = [
    F.RadioactiveDecay(decay_constant=decay_constant, volume=1, field="solute"),
    F.RadioactiveDecay(decay_constant=decay_constant, volume=1, field="1"),
]
```

Alternatively, a single decay source can be applied to all concentrations (default behaviour):
```python
import festim as F
import numpy as np


my_model = F.Simulation()

my_model.traps = [
    F.Trap(0, 1, 0, 3, materials=my_model.materials.materials[0], density=100)
]

half_life = 50  # seconds
decay_constant = np.log(2) / half_life

my_model.sources = [
    F.RadioactiveDecay(decay_constant=decay_constant, volume=1, field="all"),
]
```

On a simple simulation with no BCs, no trapping/detrapping, but initial conditions (200 for mobile and 100 for trapped concentration), a half-life of 50 s, things work as expected:

image

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

@jhdark I had to modify the formulation in order to make this work. Before, we had:

$$
\frac{\partial c_m}{\partial t} =  \nabla \cdot (D \nabla c_m )  + S - \sum \frac{\partial c_{t,i}}{\partial t}
$$

This trapping term wouldn't work when adding decay. Since:

$$
    \frac{\partial c_{t,i}}{\partial t}=k_i \ c_m \ (n_i - c_{t,i})- p_i \ c_{t,i} - \lambda c_{t,i}
$$

then the decay of trapped particles would create mobile particles!

I therefore had to:
- modify the formulation
- modify the MMS tests
- modify some other units tests

I'm happy to break this PR down and open a smaller PR that just does this

Also, it might be interesting to allow several materials per source!

### DIFF
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -188,7 +188,19 @@ class Simulation:
 
         # set sources
         for source in self.sources:
-            field_to_object[source.field].sources.append(source)
+            if isinstance(source, festim.RadioactiveDecay) and source.field == "all":
+                # this could be refactored if sources accept several fields
+                # (e.g. festim.Source(field=["0", "T"]))
+                # for field in source.field:
+                #    field_to_object[field].sources.append(source)
+
+                # create list of all unique festim.Concentration objects in field_to_object
+                # and assign source to each of them
+                for obj in set(field_to_object.values()):
+                    if isinstance(obj, festim.Concentration):
+                        obj.sources.append(source)
+            else:
+                field_to_object[source.field].sources.append(source)
 
     def attribute_boundary_conditions(self):
         """Assigns boundary_conditions to mobile and T"""
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -189,13 +189,8 @@ class Simulation:
         # set sources
         for source in self.sources:
             if isinstance(source, festim.RadioactiveDecay) and source.field == "all":
-                # this could be refactored if sources accept several fields
-                # (e.g. festim.Source(field=["0", "T"]))
-                # for field in source.field:
-                #    field_to_object[field].sources.append(source)
-
-                # create list of all unique festim.Concentration objects in field_to_object
-                # and assign source to each of them
+                # assign source to each of the unique festim.Concentration
+                # objects in field_to_object
                 for obj in set(field_to_object.values()):
                     if isinstance(obj, festim.Concentration):
                         obj.sources.append(source)
--- a/festim/sources/radioactive_decay.py
+++ b/festim/sources/radioactive_decay.py
@@ -6,5 +6,17 @@ class RadioactiveDecay(Source):
         self.decay_constant = decay_constant
         super().__init__(value=None, volume=volume, field=field)
 
+    @property
+    def decay_constant(self):
+        return self._decay_constant
+
+    @decay_constant.setter
+    def decay_constant(self, value):
+        if not isinstance(value, (float, int)):
+            raise TypeError("decay_constant must be a float or an int")
+        if value <= 0:
+            raise ValueError("decay_constant must be positive")
+        self._decay_constant = value
+
     def form(self, concentration):
         return self.decay_constant * concentration
--- a/festim/sources/radioactive_decay.py
+++ b/festim/sources/radioactive_decay.py
@@ -2,6 +2,19 @@ from festim import Source
 
 
 class RadioactiveDecay(Source):
+    """
+    A radioactive decay source.
+    dc/dt = ... - lambda * c
+    where lambda is the decay constant in 1/s.
+
+    Args:
+        decay_constant (float): The decay constant of the source in 1/s.
+        volume (int): The volume of the source.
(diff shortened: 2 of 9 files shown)


festim/generic_simulation.py
```diff
@@ -188,7 +188,19 @@ def attribute_source_terms(self):
 
         # set sources
         for source in self.sources:
-            field_to_object[source.field].sources.append(source)
+            if isinstance(source, festim.RadioactiveDecay) and source.field == "all":
+                # this could be refactored if sources accept several fields
+                # (e.g. festim.Source(field=["0", "T"]))
+                # for field in source.field:
+                #    field_to_object[field].sources.append(source)
```
jhdark, line 195: Think this could be removed an put into an issue no?

festim/sources/radioactive_decay.py
```diff
@@ -0,0 +1,22 @@
+from festim import Source
+
+
+class RadioactiveDecay(Source):
```
jhdark, line 4: Needs documentation

test/system/test_system.py
```diff
@@ -181,7 +181,9 @@ def test_run_MMS(tmpdir):
 
     f = (
         sp.diff(u, festim.t)
-        + sp.diff(v, festim.t)
+        # + sp.diff(v, festim.t)
```
jhdark, line 184:
```suggestion
```

test/unit/test_radioactive_decay.py
```diff
@@ -0,0 +1,31 @@
+from festim import RadioactiveDecay
+import pytest
+
+
+def test_init():
```
jhdark, line 5: Dont know if its worth adding doc strings on these tests given how simple they are

test/system/test_system.py
```diff
@@ -1031,3 +1034,74 @@ def test_completion_tone():
...
+    my_sim.run()
+    error_max_u = compute_error(
+        u,
+        computed=my_sim.mobile.post_processing_solution,
+        t=my_sim.t,
+        norm="error_max",
+    )
+
+    tol_u = 1e-7
+    dt = 0.1 / 50
```
jhdark, line 1104: dont need this if its steady state
```suggestion
```

test/system/test_system.py
```diff
@@ -1031,3 +1034,74 @@ def test_completion_tone():
...
+        sources=my_sources,
+        settings=my_settings,
+    )
+
+    my_sim.initialise()
+    my_sim.run()
+    error_max_u = compute_error(
+        u,
+        computed=my_sim.mobile.post_processing_solution,
+        t=my_sim.t,
```
jhdark, line 1099: Does the simulation have a t value if its steady? the test seems to work but curious

jhdark (changes requested):
Just some minor things from me, and it seems there is still a little coverage missing. Otherwise looks good, could be very useful!

festim/generic_simulation.py
RemDelaporteMathurin: I was wondering if I would do it in this PR or not

test/system/test_system.py
RemDelaporteMathurin: I think for steady state sims `my_sim.t` is zero, it's just required by `compute_error`

test/unit/test_radioactive_decay.py
RemDelaporteMathurin: That's what I thought

festim/generic_simulation.py
RemDelaporteMathurin: Maybe it's not worth it, i'll just remove the comment

RemDelaporteMathurin: It's not worth it actually, I'll simply remove it

PR #683, opened 2024-01-17
Reduce number of warnings

RemDelaporteMathurin:
## Proposed changes

This PR removes a lot of the warnings triggered when running FESTIM (and the test suite).

A couple of warnings are hard to kill and are out of our range:

- `ffcx/ir/representation.py:206: DeprecationWarning: `product` is deprecated as of NumPy 1.25.0, and will be removed in NumPy 2.0. Please use `prod` instead.` already fixed but not released https://github.com/FEniCS/ffcx/pull/626
- `DeprecationWarning: attach_operators_from_hash_data deprecated, please use UFLObject instead.` will be fixed in the next UFL release (see https://github.com/FEniCS/ufl/commit/4b6142a7c69f5360af10897c98aa9fec0f7d96ad)
- `DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).` will be fixed at some point https://github.com/tqdm/tqdm/pull/1519

## Types of changes

- [ ] Code refactoring
- [ ] New tests 

## Checklist

- [x] Black formatted

### DIFF
--- a/.github/workflows/ci_conda.yml
+++ b/.github/workflows/ci_conda.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Create Conda environment
       shell: bash -l {0}
       run: |
-        mamba install -c conda-forge fenics-dolfinx=0.7.0
+        mamba install -c conda-forge fenics-dolfinx=0.7.2
 
     - name: Install local package and dependencies
       shell: bash -l {0}
--- a/.github/workflows/ci_docker.yml
+++ b/.github/workflows/ci_docker.yml
@@ -4,7 +4,7 @@ on: [pull_request, push]
 jobs:
   run-tests:
       runs-on: ubuntu-latest
-      container: dolfinx/dolfinx:v0.7.0
+      container: dolfinx/dolfinx:v0.7.2
       steps:
         - name: Checkout code
           uses: actions/checkout@v2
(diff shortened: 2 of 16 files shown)


jhdark:
Is it worth adding in a test within festim/__init__.py or somewhere where we can test the version of dolfinx installed locally to ensure users have the correct version for full support of the application?

like:
```python
import dolfinx

if dolfinx.__version__ != 0.7.2:
    raise TypeError("local dolfinx version is {dolfinx.__version__}, please change it to v0.7.2 for full support within FESTIM")
```

Then we can have a common variable thats used in both the github action files and here?

RemDelaporteMathurin:
> Is it worth adding in a test within festim/**init**.py or somewhere where we can test the version of dolfinx installed locally to ensure users have the correct version for full support of the application?
> 
> like:
> 
> ```python
> import dolfinx
> 
> if dolfinx.__version__ != 0.7.2:
>     raise TypeError("local dolfinx version is {dolfinx.__version__}, please change it to v0.7.2 for full support within FESTIM")
> ```
> 
> Then we can have a common variable thats used in both the github action files and here?

I wouldn't advise doing this as maybe you could have a future version that is still compatible with FESTIM.
It is also too restrictive for users, imagine if for some reason they NEED 0.7.3 because it patches a bug, well until we update (and release) FESTIM, they won't be able to use it
Instead we could find a way to show which version of dolfinx the code is tested against, in the README maybe

PR #688, opened 2024-01-23
Warn about derived quantities in non cartesian

RemDelaporteMathurin:
## Proposed changes

This PR adds a warning when users employ `SurfaceFlux` with a non-cartesian mesh (eg. an axisymmetric 2D mesh).

Indeed, the flux (example here for particles) expression is

$$
J_\mathrm{total} = \int J \cdot n \ dS
$$

where $n$ is the normal vector and

$$
J = - D \nabla c
$$

In cartesian coordinates, the element $dS$ is $dS = dx \ dy $, assuming it is $y$ independent $dS = y \ dx$.
And this is the default behaviour in fenics.

However in cylindrical $(r, \theta, z)$, the surface element is $dS = r \ dr \ d\theta$, assuming it is axisymmetric ($\theta$ independent), then $dS =\theta \ r \ dr$

The flux becomes:

$$
J_\mathrm{total} = \int J \cdot n \ dS = \theta \int r \ J \cdot n \ dr
$$

which is NOT the default behaviour of `SurfaceFlux`.
While we wait for a native support of fluxes in cylindrical coordinates to be implemented, users should at least be warn that this may be an issue.

The same goes for `TotalVolume` , `AverageVolume` , etc.

Thanks @WeaverC20 for helping spotting this

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted


RemDelaporteMathurin:
@jhdark can we merge this before #689 ?

festim/generic_simulation.py
```diff
@@ -270,6 +270,26 @@ def initialise(self):
 
         self.h_transport_problem.initialise(self.mesh, self.materials, self.dt)
 
+        # raise warning if SurfaceFlux is used with non-cartesian meshes
+        for export in self.exports.exports:
```
KulaginVladimir, line 274: just to remind, `.exports` will be deprecated

RemDelaporteMathurin: Yes I probably have some merge conflicts I need to take care of

jhdark (approved):
Looks good to me!

PR #689, opened 2024-01-24
Cylindrical and spherical versions of SurfaceFlux

RemDelaporteMathurin:
## Proposed changes

This PR implements the SurfaceFlux in cylindrical and spherical coordinates: `SurfaceFluxCylindrical` and `SurfaceFluxSpherical` as well as a little bit of refactoring.

Usage:

```python
full_coverage_cylindrical_surface_flux = F.SurfaceFluxCylindrical(field="solute", surface=1)
half_coverage_cylindrical_surface_flux = F.SurfaceFluxCylindrical(field="solute", surface=1, azimuthal_range=(0, np,pi))
```

Similarly for spherical meshes:
```python
full_coverage_spherical_surface_flux = F.SurfaceFluxSpherical(field="solute", surface=1)
quarter_coverage_spherical_surface_flux = F.SurfaceFluxSpherical(field="solute", surface=1, azimuthal_range=(0, np,pi), polar_range=(0, np.pi))
```

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted


RemDelaporteMathurin:
@jhdark let me know what you think and I'll complete with documentation

jhdark (changes requested):
This looks great to me, also like the fact that the units of the results are given in the doc strings, think this is smth we could implement in other export classes (accounting for differences in different dimensions).

However, just some small coverage tests that need to be added.

Then its good to go!

RemDelaporteMathurin:
> the units of the results are given in the doc strings, think this is smth we could implement in other export classes (accounting for differences in different dimensions)

Maybe what we should do is add the units in the docstrings and also in the title of the csv column based on the dimension of the mesh

jhdark:
I like the idea, although this could be a big change, we can add in another PR. I can make a new issue

jhdark (approved):
LGTM

Issue #472, opened 2022-05-12
A warning should be thrown when multiple bcs are set on the same surface (convective, dirichlets)


RemDelaporteMathurin:
Currently, when several dirichlet boundary conditions are set on the same surface, only the last one will be really applied to the model. However, no warnings are thrown.
For instance:

```python
import festim as F
import numpy as np

my_model = F.Simulation()

my_model.mesh = F.MeshFromVertices(np.linspace(0, 1, 10))

my_model.T = F.Temperature(100)

my_model.materials = F.Material(1, 1, 0)


my_model.boundary_conditions = [
    F.DirichletBC(value=100, field="solute", surfaces=1),
    F.DirichletBC(value=200, field="solute", surfaces=1),
]

my_model.settings = F.Settings(
    absolute_tolerance=1e-10,
    relative_tolerance=1e-10,
    transient=False,
)

my_model.initialise()
my_model.run()
```

Doesn't raise any warning. Similarly:

```python
my_model.boundary_conditions = [
    F.DirichletBC(value=1e15, field="solute", surfaces=[1, 2]),
    F.RecombinationFlux(Kr_0=1e12, E_Kr=0, order=1, surfaces=1),
]
```

This doesn't raise any warning - it should.

RemDelaporteMathurin:
To fix this I would simply add a check at in `initialise()`, would need to do the same for `HeatTransferProblem`

KulaginVladimir:
Hi @RemDelaporteMathurin. Is this issue still open?

RemDelaporteMathurin:
@KulaginVladimir it is still waiting for a fix

KulaginVladimir:
@RemDelaporteMathurin. I can start doing it just to deal with these ~ 2-year-old issues. I assume, this small fix would help users to avoid accident mistakes

RemDelaporteMathurin:
Sounds good to me! Yes this is a non-critical issue that I initially left open for newcomers to give it a go but we can fix it now 🙂

KulaginVladimir:
@RemDelaporteMathurin some details

> To fix this I would simply add a check at in initialise(), would need to do the same for HeatTransferProblem

[...]

PR #692, opened 2024-01-29
RaiseError when DirichletBC is set with another BC

KulaginVladimir:
## Proposed changes

With this PR, the error will be raised when `F.DirichletBC` is simultaneously set with `F.FluxBC` or `F.DirichletBC` on the same surface, according to the proposal of @RemDelaporteMathurin in #472. 

@RemDelaporteMathurin, @jhdark I added a test for this, but it worth examining.

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -210,9 +210,7 @@ class Simulation:
 
         # collect all DirichletBCs
         dc_bcs = [
-            bc
-            for bc in self.boundary_conditions
-            if isinstance(bc, festim.DirichletBC) and (bc.field in valid_fields)
+            bc for bc in self.boundary_conditions if isinstance(bc, festim.DirichletBC)
         ]
 
         for bc in self.boundary_conditions:
@@ -229,11 +227,14 @@ class Simulation:
                     bc == dc_bc or bc.field != dc_bc.field
                 ):  # skip if the same BC or different fields
                     continue
-                for surf in bc.surfaces:
-                    if surf in dc_bc.surfaces:
-                        raise ValueError(
-                            f"A DirichletBC is simultaneously set with another boundary condition on surface {surf} for field {dc_bc.field}"
-                        )
+                # check if BCs share the same surfaces using the set().isdisjoint() method
+                # that returns True if the first set has no elements in common with other containers
+                if not set(bc.surfaces).isdisjoint(dc_bc.surfaces):
+                    # convert lists of surfaces to sets and obtain their intersetion
+                    intersection = set(bc.surfaces) & set(dc_bc.surfaces)
+                    raise ValueError(
+                        f"A DirichletBC is simultaneously set with another boundary condition on surfaces {intersection} for field {dc_bc.field}"
+                    )
 
     def initialise(self):
         """Initialise the model. Defines markers, create the suitable function
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -208,6 +208,13 @@ class Simulation:
             + [i + 1 for i, _ in enumerate(self.traps.traps)]
         )
 
+        # collect all DirichletBCs
+        dc_bcs = [
+            bc
+            for bc in self.boundary_conditions
+            if isinstance(bc, festim.DirichletBC) and (bc.field in valid_fields)
+        ]
+
         for bc in self.boundary_conditions:
             if bc.field not in valid_fields:
                 raise ValueError(f"{bc.field} is not a valid field for BC")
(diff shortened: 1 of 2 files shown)


festim/generic_simulation.py
```diff
@@ -208,13 +208,32 @@ def attribute_boundary_conditions(self):
...
+        # collect all DirichletBCs
+        dc_bcs = [
+            bc
+            for bc in self.boundary_conditions
+            if isinstance(bc, festim.DirichletBC) and (bc.field in valid_fields)
+        ]
+
         for bc in self.boundary_conditions:
             if bc.field not in valid_fields:
                 raise ValueError(f"{bc.field} is not a valid field for BC")
```
RemDelaporteMathurin, line 220: This could me moved out of the loop into something concise like

```python
if any([bc.field not in valid_fields for bc in self.boundary_conditions]):
    raise ValueError(...)
```

RemDelaporteMathurin (changes requested):
Thanks @KulaginVladimir this looks great! 

I think we can simplify the logic a little bit see comments

festim/generic_simulation.py
KulaginVladimir: So, the error message should be reduced to `"Some fields are not valid for BC"`?

festim/generic_simulation.py
```diff
@@ -208,13 +208,32 @@ def attribute_boundary_conditions(self):
             + [i + 1 for i, _ in enumerate(self.traps.traps)]
         )
 
+        # collect all DirichletBCs
+        dc_bcs = [
+            bc
+            for bc in self.boundary_conditions
+            if isinstance(bc, festim.DirichletBC) and (bc.field in valid_fields)
```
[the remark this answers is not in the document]
KulaginVladimir: Agree

festim/generic_simulation.py
```diff
@@ -208,13 +208,32 @@ def attribute_boundary_conditions(self):
...
+            for dc_bc in dc_bcs:
+                if (
+                    bc == dc_bc or bc.field != dc_bc.field
+                ):  # skip if the same BC or different fields
+                    continue
+                for surf in bc.surfaces:
+                    if surf in dc_bc.surfaces:
+                        raise ValueError(
+                            f"A DirichletBC is simultaneously set with another boundary condition on surface {surf} for field {dc_bc.field}"
+                        )
```
KulaginVladimir, line 236: I also wonder if we can speed up this a bit, since its a nested loop. I don't know how complicated simulations might be, but another possibility is to use the `set` method as mentioned [here](https://stackoverflow.com/questions/3170055/test-if-lists-share-any-items-in-python). The modified part would be:
```python
for dc_bc in dc_bcs:
    if (
        bc == dc_bc or bc.field != dc_bc.field
    ):  # skip if the same BC or different fields
        continue
    intersection = set(bc.surfaces) & set(dc_bc.surfaces)
    if bool(intersection):
        raise ValueError(
            f"A DirichletBC is simultaneously set with another boundary condition on surfaces {intersection} for field {bc.field}"
        )
                    )
``` 
@RemDelaporteMathurin, please, let me know what do you think

RemDelaporteMathurin: Sure why not! Just make sure to add enough comments to describe what is done here cause this may look complex for non-python experts

festim/generic_simulation.py
```diff
@@ -229,11 +227,14 @@ def attribute_boundary_conditions(self):
...
-                            f"A DirichletBC is simultaneously set with another boundary condition on surface {surf} for field {dc_bc.field}"
-                        )
+                # check if BCs share the same surfaces using the set().isdisjoint() method
+                # that returns True if the first set has no elements in common with other containers
+                if not set(bc.surfaces).isdisjoint(dc_bc.surfaces):
+                    # convert lists of surfaces to sets and obtain their intersetion
+                    intersection = set(bc.surfaces) & set(dc_bc.surfaces)
+                    raise ValueError(
+                        f"A DirichletBC is simultaneously set with another boundary condition on surfaces {intersection} for field {dc_bc.field}"
+                    )
```
KulaginVladimir, line 237: Well, I modified it using a more efficient method, but also added a variable for intersection to output the set of surfaces if more than one surface is caught during the check

RemDelaporteMathurin: that is slick thanks!

RemDelaporteMathurin (approved):
Thanks @KulaginVladimir , as usual I'll leave this open to leave people a chance to comment.

Issue #494, opened 2022-07-01
`Materials`, `Exports`, `DerivedQuantities`, ... should be a subclass of `list`


RemDelaporteMathurin:
Users should be able to do

```python
my_materials = festim.Materials()

tungsten = festim.Material(....)

my_materials.append(tungsten)
```

To do so, the current Materials class:
https://github.com/festim-dev/FESTIM/blob/621769f7bf56af1bf6a08612c0cd6d269e8f93da/festim/materials/materials.py#L10-L18

Needs to become:

```python
class Materials(list):

    def __init__(self, *args):
        super().__init__(*args)
```

Then instead of doing:

```python
my_materials = festim.Materials(......)

for mat in my_materials.materials:
    ...
```

one would do:
```python
my_materials = festim.Materials(......)

for mat in my_materials:
    ...
```

Moreover, this:
https://github.com/festim-dev/FESTIM/blob/621769f7bf56af1bf6a08612c0cd6d269e8f93da/festim/materials/materials.py#L78

would become:

```python
for mat in self:
```

Progress track:

- [x] `Materials` : #694 
- [x] `Exports` : #695 
- [ ] `DerivedQuantities`: #697 
- [ ] `Traps`: #697

KulaginVladimir:
@RemDelaporteMathurin. Could you please clarify the practical benefit of it a bit more? As I understand, this enhancement will make more like a built-in class, so the users won't need to manually create lists every time and work with them, but rather use documented methods of the FESTIM class (I assume that the docs also will be modified) + some code refactoring. Am I correct?

RemDelaporteMathurin:
1. It is super annoying for users to write

```python
my_materials = F.Materials([....])

second_material = my_materials.materials[1]
```
thanks to this change, this will become much simpler:

```python
my_materials = F.Materials([....])

second_material = my_materials[1]
```

Same for boundary conditions, traps, derivedquantities, exports, and others.

2. it will clear a lot of the code base

KulaginVladimir:
Ok, I can try. How we'll do it:  for all classes at once or iteratively?

RemDelaporteMathurin:
let's start with `Materials`

KulaginVladimir:
@RemDelaporteMathurin, what do you think of `TXT Exports` considering the deprecation?

RemDelaporteMathurin:
> @RemDelaporteMathurin, what do you think of `TXTExports` considering its deprecation?

I would ignore it

PR #694, opened 2024-01-31
F.Materials -> subclass of list

KulaginVladimir:
## Proposed changes

This PR modifies `F.Materials` to be a subclass of `list` following #494. 

@RemDelaporteMathurin, @jhdark, please, let me know what do you think and if this is what is needed.
## Types of changes

- [x] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -136,8 +136,6 @@ class Simulation:
             self._materials = festim.Materials(value)
         elif isinstance(value, festim.Material):
             self._materials = festim.Materials([value])
-        elif isinstance(value, festim.Materials):
-            self._materials = value
         elif value is None:
             self._materials = value
         else:
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -132,7 +132,9 @@ class Simulation:
 
     @materials.setter
     def materials(self, value):
-        if isinstance(value, list):
+        if isinstance(value, festim.Materials):
+            self._materials = value
+        elif isinstance(value, list):
             self._materials = festim.Materials(value)
         elif isinstance(value, festim.Material):
             self._materials = festim.Materials([value])
--- a/festim/materials/materials.py
+++ b/festim/materials/materials.py
@@ -21,6 +21,24 @@ class Materials(list):
         self.density = None
         self.Q = None
 
+    def __eq__(self, __value: object) -> bool:
+        """
+        A custom method to check equality of F.Materials objects
+
+        Usage:
+            F.Materials([]) == F.Materials([])
+        """
+        return id(self) == id(__value)
+
+    def __ne__(self, __value: object) -> bool:
+        """
+        A custom method to check unequality of F.Materials objects
+
+        Usage:
+            F.Materials([]) != F.Materials([])
+        """
+        return not id(self) == id(__value)
+
     def check_borders(self, size):
         """Checks that the borders of the materials match
 
--- a/festim/materials/materials.py
+++ b/festim/materials/materials.py
@@ -9,9 +9,7 @@ from typing import Union
 
 class Materials(list):
     """
-    Args:
-        materials (list, optional): contains festim.Material objects.
-            Defaults to [].
+    A list of festim.Material objects
     """
 
     def __init__(self, *args):
--- a/festim/materials/materials.py
+++ b/festim/materials/materials.py
@@ -21,24 +21,6 @@ class Materials(list):
         self.density = None
         self.Q = None
 
-    def __eq__(self, __value: object) -> bool:
-        """
-        A custom method to check equality of F.Materials objects
-
-        Usage:
-            F.Materials([]) == F.Materials([])
-        """
-        return id(self) == id(__value)
-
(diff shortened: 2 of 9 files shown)


festim/generic_simulation.py
```diff
@@ -136,8 +136,6 @@ def materials(self, value):
             self._materials = festim.Materials(value)
         elif isinstance(value, festim.Material):
             self._materials = festim.Materials([value])
-        elif isinstance(value, festim.Materials):
-            self._materials = value
```
RemDelaporteMathurin, line 140: Hm I'd double check on this one.
Maybe it's better to have:
```python
        if isinstance(value, festim.Materials):
            self._materials = value
        elif isinstance(value, list):
            self._materials = festim.Materials(value)
        elif isinstance(value, festim.Material):
            self._materials = festim.Materials([value])

```

festim/materials/materials.py
```diff
@@ -7,15 +7,15 @@
 from typing import Union
 
 
-class Materials:
+class Materials(list):
     """
     Args:
         materials (list, optional): contains festim.Material objects.
```
RemDelaporteMathurin, line 13: Documentation needs updating cause it doesn't have a `materials` argument anymore.
I'd simply have

```python
a list of festim.Material objects
```

RemDelaporteMathurin (changes requested):
@KulaginVladimir this is exactly what we need to do. I hope you see why it increases the readibility of the code base, and is just more logical overall.

We'd have to do the same for other classes.

A few small comments on my side

festim/generic_simulation.py
KulaginVladimir: I've double checked it :) Tests pass + manually verified. Alternatively, it'd be required to do what you suggested, yes. Maybe, your approach adds clarity for the code and there won't be any unexpected problems

festim/materials/materials.py
KulaginVladimir: Right

festim/generic_simulation.py
RemDelaporteMathurin: The one thing I'm worry about is what happens if we do:

```python
my_model.materials = F.Materials(....)
```
Theoretically, it is of type list (since it inherits from list). So, according to the code, it should perform:

```python
self._materials = festim.Materials(value)
```
but if value is already a festim.Materials class then any of the classe's attributes could be overwritten

I wonder what happens when you do:


```python

my_materials = F.Materials([1, 2, 3])
my_materials.D = 2 

my_model.materials = my_materials

print(my_model.materials.D)  # does it print 2 or does it print the default value of the argument None?
```

KulaginVladimir: Well, it's a very good point. It returns `None`. Your approach returns `2`. Btw, I've included it in the latest commit

RemDelaporteMathurin: Ha! Knew it! Maybe we want to write a test for this then since it wasn't caught by the CI.

Something like:

```python


@pytest.mark.parametrize("input,expected_output", [(test_materials, test_materials) ])
def test_materials_setter():
   my_model = F.Model()
   test_materials = F.Materials([])
   my_model.materials = test_materials 
   assert my_model.materials == test_materials
```

I know it sounds silly but still good to test!

RemDelaporteMathurin: @KulaginVladimir I edited my code I initially put a parametrize mark but no need

test/unit/test_materials.py
```diff
@@ -297,3 +297,14 @@ def test_error_wrong_solubility_law_string():
         match="Acceptable values for solubility_law are 'henry' and 'sievert'",
     ):
         F.Material(1, 1, 1, solubility_law="foo")
+
+
+def test_materials_setter():
+    my_model = F.Simulation()
+    test_materials = F.Materials([])
+    my_model.materials = test_materials
+    assert my_model.materials == test_materials
```
RemDelaporteMathurin, line 306: This shouldn't really be in test_materials since it's not testing functionality in the Materials class but rather the simulation class

festim/materials/materials.py
```diff
@@ -21,6 +21,24 @@ def __init__(self, *args):
...
+        return id(self) == id(__value)
+
+    def __ne__(self, __value: object) -> bool:
+        """
+        A custom method to check unequality of F.Materials objects
+
+        Usage:
+            F.Materials([]) != F.Materials([])
+        """
+        return not id(self) == id(__value)
```
RemDelaporteMathurin, line 40:
```suggestion
        return not self == __value
```

KulaginVladimir: Removed this

test/unit/test_materials.py
KulaginVladimir: Moved this

festim/generic_simulation.py
KulaginVladimir: Modified this

RemDelaporteMathurin (approved):
Looks good to me! Apologies for the back and forth discussion on equlity!

We should do the same for other classes too!

PR #697, opened 2024-02-04
`F.Traps` and `F.DerivedQuantities` -> subclasses of list

KulaginVladimir:
## Proposed changes

This PR modifies `F.Traps` and `F.DerivedQuantities` to be subclasses of list following #494.

An attempt was made to add a check that:
- a list is passed to `F.Materials` / `F.Exports` / `F.Traps` / `F.DerivedQuantities`
- the passed list contains  `F.Material` / `F.Export` / `F.Trap` / `F.DerivedQuantity` 

Additionally, a `.materials` / `.exports` / `.traps` / `.derived_quantities` property is temporarly added to `F.Materials` / `F.Exports` / `F.Traps` / `F.DerivedQuantities` to minimize a breaking change, as privately discussed with @RemDelaporteMathurin. 

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted


test/unit/test_exports/test_derived_quantities/test_derived_quantities.py
```diff
@@ -26,14 +26,22 @@ class TestMakeHeader:
     min_vol_1 = MinimumVolume("retention", 2)
     max_vol_1 = MaximumVolume("T", 2)
 
+    @property
+    def my_derv_quant(self):
+        return self._my_derv_quant
+
+    @my_derv_quant.setter
+    def my_derv_quant(self, value):
+        self._my_derv_quant = DerivedQuantities(value)
```
RemDelaporteMathurin, line 35: Can you not simply create an empty DerivedQuantities object?

test/unit/test_exports/test_derived_quantities/test_derived_quantities.py
```diff
@@ -165,60 +173,63 @@ class TestCompute:
...
     my_mats.thermal_cond = f.interpolate(f.Constant(2), V)
 
+    @property
+    def my_derv_quant(self):
+        return self._my_derv_quant
+
+    @my_derv_quant.setter
+    def my_derv_quant(self, value):
+        self._my_derv_quant = DerivedQuantities(value)
+
```
RemDelaporteMathurin, line 183: Same comment as above

test/unit/test_exports/test_derived_quantities/test_derived_quantities.py
```diff
@@ -330,3 +337,23 @@ def test_filename_ends_with_csv():
...
+            TypeError,
+            match="festim.DerivedQuantities must be a list",
+        ):
+            DerivedQuantities(dq_combination)
+
+    with pytest.raises(
+        TypeError,
+        match="festim.DerivedQuantities must be a list of festim.DerivedQuantity",
+    ):
+        DerivedQuantities([flux1, 2])
```
RemDelaporteMathurin, line 359: This tests that an error is raised at instantiation but what about after instantiation?

test/unit/test_exports/test_exports.py
```diff
@@ -0,0 +1,22 @@
...
+            TypeError,
+            match="festim.Exports must be a list",
+        ):
+            festim.Exports(export_combination)
+
+    with pytest.raises(
+        TypeError,
+        match="festim.Exports must be a list of festim.Export",
+    ):
+        festim.Exports([export, 2])
```
RemDelaporteMathurin, line 22: Same comment about testing after instantiation

test/unit/test_materials.py
```diff
@@ -307,3 +307,23 @@ def test_equality_identity_two_empty_materials():
...
+            TypeError,
+            match="festim.Materials must be a list",
+        ):
+            F.Materials(mat_combination)
+
+    with pytest.raises(
+        TypeError,
+        match="festim.Materials must be a list of festim.Material",
+    ):
+        F.Materials([my_mat, 2])
```
RemDelaporteMathurin, line 329: Same comment about testing after instantiation

festim/exports/exports.py
```diff
@@ -8,12 +9,33 @@ class Exports(list):
...
         self.nb_iterations = 0
 
+    @property
+    def exports(self):
+        warnings.warn(
+            "The exports attribute will be deprecated in a future release, please use festim.Exports[:] instead",
+            DeprecationWarning,
+        )
+        return self
+
```
RemDelaporteMathurin, line 38: I would have a setter/getter for `exports` (same for traps, derived_quantities, etc)

RemDelaporteMathurin (changes requested):
@KulaginVladimir thanks for this! almost there!

I'm just realising that having the property for .traps (for instance) doesn't break the api if users do: 
```
print(my_traps.traps[2])
```

but users can't do

```
my_traps.traps.append(...)
```
Can you please add a setter/getter in addition to the @property please?

festim/concentration/traps/traps.py
```diff
@@ -1,52 +1,62 @@
...
+
+    def __init__(self, *args):
+        # checks that input is list
+        try:
+            super().__init__(*args)
+        except:
+            raise TypeError("festim.Traps must be a list")
+
+        # checks that list elements are festim.Export
+        if len(self) != 0:
```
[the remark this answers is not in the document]
KulaginVladimir: Not really. I think there appeared an error, so I messed it up a bit

festim/concentration/traps/traps.py
```diff
@@ -1,52 +1,62 @@
...
-    @traps.setter
-    def traps(self, value):
-        if isinstance(value, list):
-            if not all(isinstance(t, festim.Trap) for t in value):
-                raise TypeError("traps must be a list of festim.Trap")
-            self._traps = value
-        else:
-            raise TypeError("traps must be a list")
+        warnings.warn(
+            "The traps attribute will be deprecated in a future release, please use festim.Traps[:] instead",
```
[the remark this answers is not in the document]
KulaginVladimir: Sure, I'll modify warnings in all classes at once

festim/exports/derived_quantities/derived_quantities.py
```diff
@@ -26,22 +27,39 @@ class DerivedQuantities:
...
+            super().__init__(*args)
+        except:
+            raise TypeError("festim.DerivedQuantities must be a list")
+
+        # checks that list elements are festim.Export
+        if len(self) != 0:
+            if not all(isinstance(t, DerivedQuantity) for t in self):
+                raise TypeError(
+                    "festim.DerivedQuantities must be a list of festim.DerivedQuantity"
+                )
```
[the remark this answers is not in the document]
KulaginVladimir: Right. This is a good point. Please, see the comment below

festim/materials/materials.py
```diff
@@ -13,14 +14,32 @@ class Materials(list):
...
+        # checks that input is list
+        try:
+            super().__init__(*args)
+        except:
+            raise TypeError("festim.Materials must be a list")
+
+        # checks that list elements are festim.Material
+        if len(self) != 0:
+            if not all(isinstance(t, festim.Material) for t in self):
+                raise TypeError("festim.Materials must be a list of festim.Material")
```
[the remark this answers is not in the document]
KulaginVladimir: _Here I'll make a comment on `F.Materials` assuming all modified classes_

As I understand, `.append` would also add an object of any type to `F.Traps.traps` in the previous version (Am I wrong?). In the current version, the validaty of appended items also won't be tested, yes. To add a check of item validaty to every list method, I think we have to re-define all of them (at least, those that are connected to list item assignmen: `__setitem__`, `insert`, `append`, `extend`), i.e. to make a real user-defined list. As an example, it can be done like this:
```python
class Materials(list):
    """
    A list of festim.Material objects
    """

    def __init__(self, iterable):
        # checks that input is list
        if isinstance(self, list):
            super().__init__(self._validate_material(item) for item in iterable)
        else:
            raise TypeError("festim.Materials must be a list")
        
        ...

    def __setitem__(self, index, item):
        super().__setitem__(index, self._validate_material(item))

    def insert(self, index, item):
        super().insert(index, self._validate_material(item))

    def append(self, item):
        super().append(self._validate_material(item))
        
    def _validate_material(self, value):
        if isinstance(value, festim.Material):
            return value
        raise TypeError("festim.Materials must be a list of festim.Material")
```    
@RemDelaporteMathurin what do you think?

test/unit/test_exports/test_derived_quantities/test_derived_quantities.py
KulaginVladimir: So, old tests in each method of this class re-assigned `my_derv_quant.derived_quantities` property that is a list, e.g.: 
https://github.com/festim-dev/FESTIM/blob/a484d8d1a30de1a2273431f222e59896d2e4bed0/test/unit/test_exports/test_derived_quantities/test_derived_quantities.py#L17-L33

As I understand, if `F.DerivedQuantities` is now a list itself, than to modify its elements we have to use the list methods. Direct assignment of `[F.DerivedQuantity1,F.DerivedQuantity2]` to `self.my_derv_quant` will make it a list that does not have `F.DerivedQuantities` attributes leading to the test failure. To minimally change tests, I've introduced the @property and its setter, so each re-assignment of  `self.my_derv_quant` introduces a new `F.DerivedQuantities`. Does it sound logically?

test/unit/test_exports/test_derived_quantities/test_derived_quantities.py
KulaginVladimir: Yes, it should be addressed if we want to introduce a check of item validaty

festim/exports/exports.py
KulaginVladimir: As I've checked locally, `my_exports.exports.append()` works fine. Maybe, it's not needed?

festim/concentration/traps/traps.py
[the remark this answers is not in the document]
RemDelaporteMathurin: The following without any issue so I think you can remove this check on line 19.
```python
a = []

print(any([isinstance(x, int) for x in a]))
```

```suggestion
```

festim/concentration/traps/traps.py
```diff
@@ -1,52 +1,62 @@
...
+    """
+    A list of festim.Trap objects
+    """
+
+    def __init__(self, *args):
+        # checks that input is list
+        try:
+            super().__init__(*args)
+        except:
+            raise TypeError("festim.Traps must be a list")
```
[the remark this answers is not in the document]
RemDelaporteMathurin: I'm not sure I'm following.

What I'm suggesting is doing something like:

```python
if not isinstance(args, list):
    raise TypeError(...)
```

What do you think?

festim/materials/materials.py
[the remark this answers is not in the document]
RemDelaporteMathurin: > As I understand, `.append` would also add an object of any type to `F.Materials.materials` in the previous version (Am I wrong?)

Yes



> As an example, it can be done like this:

This seems to be exactly what is suggested [here](https://realpython.com/inherit-python-list/)
I'm happy with it!

Later we can also refactore things a bit and do something similar like [ what is done in OpenMC](https://github.com/openmc-dev/openmc/blob/f14fc55e60463710d3f165b0a9cdc2ef9fee1534/openmc/checkvalue.py#L284) where they have a `CheckedList` class and have classes inherit from it. But let's keep this for later.

festim/exports/exports.py
RemDelaporteMathurin: Ok for append, I suppose that's because we are not setting the attribute but modifying it.

How about:

```python
my_exports.exports = []
```

Running the following throws an error:

```python
class NewList(list):
    @property
    def exports(self):
        return self


my_new_list = NewList([1, 2, 3, 1.2])

my_new_list.exports = [2]
```

festim/concentration/traps/traps.py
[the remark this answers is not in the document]
KulaginVladimir: > What do you think?

Yeah, I've mentioned this below in the example. So, I'll do

> Instead of a try/except I would either have a type check (testing that self is type list) or nothing at all

Just thought that "nothing at all" means that we don't necessarly need a check, since TypeError will appear during `super().__init__`

festim/materials/materials.py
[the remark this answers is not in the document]
KulaginVladimir: > This seems to be exactly what is suggested [here](https://realpython.com/inherit-python-list/)

Ow, yes, but I read the russian topic on it. I suppose it was just a translation :) So, I'll give a try

test/unit/test_exports/test_derived_quantities/test_derived_quantities.py
KulaginVladimir: ok

festim/exports/exports.py
KulaginVladimir: Right, that won't work

festim/concentration/traps/traps.py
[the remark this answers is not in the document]
RemDelaporteMathurin: Gotcha, let's close this thread then

test/system/test_misc.py
```diff
@@ -312,3 +312,121 @@ def test_materials_setter():
...
+        TypeError,
+        match="materials must be a list",
+    ):
+        my_mats.materials = my_mat
+
+    with pytest.raises(
+        TypeError,
+        match="materials must be a list of festim.Material",
+    ):
+        my_mats.materials = [my_mat, 1]
```
RemDelaporteMathurin, line 432: I would split this into several tests.
The issue here is that if an assertion fails, then none of the following ones will be run.

RemDelaporteMathurin: Plus I would move all these tests in the right files (test_exports.py, test_materials.py etc)
System tests are usually for "end-to-end" tests

KulaginVladimir: I splitted tests in `misc` and moved them to the corresponding folders

RemDelaporteMathurin (approved):
Thanks for this @KulaginVladimir the code is much cleaner now!

Issue #507, opened 2022-08-07
`Stepsize` should have a `adapt_when` argument


RemDelaporteMathurin:
Currently, users can set the `t_stop` argument, which corresponds to the time after which the stepsize should stop adapting (increase or decrease).

However, it could be more flexible if users were able to pass a function of time to determine if it should adapt.

For instance:

```python
def func(t):
    return t < 100

my_stepsize = F.Stepsize(..., adapt_when=func)

```

Or maybe a function that returns a boolean and the maximum stepsize at the same time.

KulaginVladimir:
Looks interesting. I think it can really help when modelling periodic transient events on large timescales.

RemDelaporteMathurin:
@KulaginVladimir do you think it would be better to allow `max_stepsize` to be a function of time or to implement an `adapt_when` argument?

KulaginVladimir:
Personally, max_stepsize appears to be more flexible as we can control both the 'when' event and the step-size value. 

Though, I wonder how it'd work. Assume we set max_stepsize = 1 for 0<t<t1 and max_stepsize = 2 otherwise. After t=t1, the step-size will smoothly rise via adaptation or it'll just jump to 2?

RemDelaporteMathurin:
> After t=t1, the step-size will smoothly rise via adaptation or it'll just jump to 2?

The later. We would basically let the stepsize adapt everytime BUT if max_stepsize(t) is not None, then we cap it to some max_stepsize(t) if it's larger.

KulaginVladimir:
Thus, it's similar to your earlier hint in slack on how to adapt step-size periodically via custom class. Sounds nice! 

Certainly, will add flexibility.

KulaginVladimir:
On the modifications:
If we introduce the `max_stepsize` function, then we don't need `t_stop ` and `stepsize_stop_max `, right? If so, would it be required to keep these attributes for some time and introduce a DeprecationWarning to give the users time to adapt?

RemDelaporteMathurin:
Correct! Maybe keeping them for now with a DeprecationWarning would be a good idea yes

PR #698, opened 2024-02-10
`max_stepsize` attribute for `festim.Stepsize`

KulaginVladimir:
## Proposed changes

As discussed in #507, this PR introduces `max_stepsize` attribute for `festim.Stepsize` instead of `t_stop` and `stepsize_stop_max`.  The old parameters are left unchanged, but a `DeprecationWarning` is added. With this change, the basic usage now becomes:

```python
my_stepsize = F.Stepsize(
    initial_value=0.5,
    stepsize_change_ratio=1.1,
    max_stepsize=lambda t: None if t < 1 else 2,
    dt_min=1e-05,
)
``` 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/docs/source/userguide/stepsize.rst
+++ b/docs/source/userguide/stepsize.rst
@@ -28,6 +28,12 @@ To cap the stepsize after some time, the parameters ``t_stop`` and ``stepsize_st
 
     my_stepsize = F.Stepsize(initial_value=1.2, stepsize_change_ratio=1.5, dt_min=1e-6, t_stop=10, stepsize_stop_max=1.5)
 
+Please note that the parameters ``t_stop`` and ``stepsize_stop_max`` will be deprecated in a future release. To set the maximal value of the stepsize, consider using the ``max_stepsize`` parameter:
+
+.. code-block:: python
+
+    my_stepsize = F.Stepsize(initial_value=1.2, stepsize_change_ratio=1.5, dt_min=1e-6, max_stepsize=lambda t: 1 if t < 1 else 2)
+
 The ``milestones`` argument can be used to make sure the simulation passes through specific times.
 This will modify the stepsize as needed.
 
--- a/docs/source/userguide/stepsize.rst
+++ b/docs/source/userguide/stepsize.rst
@@ -28,11 +28,13 @@ To cap the stepsize after some time, the parameters ``t_stop`` and ``stepsize_st
 
     my_stepsize = F.Stepsize(initial_value=1.2, stepsize_change_ratio=1.5, dt_min=1e-6, t_stop=10, stepsize_stop_max=1.5)
 
-Please note that the parameters ``t_stop`` and ``stepsize_stop_max`` will be deprecated in a future release. To set the maximal value of the stepsize, consider using the ``max_stepsize`` parameter:
+.. warning::
+    
+    Please note that the parameters ``t_stop`` and ``stepsize_stop_max`` will be deprecated in a future release. To set the maximal value of the stepsize, consider using the ``max_stepsize`` parameter:
+    
+    .. code-block:: python
 
-.. code-block:: python
-
-    my_stepsize = F.Stepsize(initial_value=1.2, stepsize_change_ratio=1.5, dt_min=1e-6, max_stepsize=lambda t: 1 if t < 1 else 2)
+        my_stepsize = F.Stepsize(initial_value=1.2, stepsize_change_ratio=1.5, dt_min=1e-6, max_stepsize=lambda t: 1 if t < 1 else 2)
 
 The ``milestones`` argument can be used to make sure the simulation passes through specific times.
 This will modify the stepsize as needed.
--- a/festim/stepsize.py
+++ b/festim/stepsize.py
@@ -50,10 +50,14 @@ class Stepsize:
     ) -> None:
         self.adaptive_stepsize = None
         if stepsize_change_ratio is not None:
+            if t_stop or stepsize_stop_max:
+                warnings.warn(
+                    "stepsize_stop_max and t_stop attributes will be deprecated in a future release, please use max_stepsize instead",
+                    DeprecationWarning,
+                )
+                max_stepsize = lambda t: stepsize_stop_max if t >= t_stop else None
             self.adaptive_stepsize = {
                 "stepsize_change_ratio": stepsize_change_ratio,
-                "t_stop": t_stop,
-                "stepsize_stop_max": stepsize_stop_max,
                 "max_stepsize": max_stepsize,
                 "dt_min": dt_min,
             }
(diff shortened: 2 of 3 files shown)


docs/source/userguide/stepsize.rst
```diff
@@ -28,6 +28,12 @@ To cap the stepsize after some time, the parameters ``t_stop`` and ``stepsize_st
 
     my_stepsize = F.Stepsize(initial_value=1.2, stepsize_change_ratio=1.5, dt_min=1e-6, t_stop=10, stepsize_stop_max=1.5)
 
+Please note that the parameters ``t_stop`` and ``stepsize_stop_max`` will be deprecated in a future release. To set the maximal value of the stepsize, consider using the ``max_stepsize`` parameter:
+
+.. code-block:: python
+
+    my_stepsize = F.Stepsize(initial_value=1.2, stepsize_change_ratio=1.5, dt_min=1e-6, max_stepsize=lambda t: 1 if t < 1 else 2)
```
RemDelaporteMathurin, line 35: I would change this to a [warning block](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html#warnings). I hope the syntax is correct!

```suggestion
.. warning::
    Please note that the parameters ``t_stop`` and ``stepsize_stop_max`` will be deprecated in a future release. To set the maximal value of the stepsize, consider using the ``max_stepsize`` parameter:

    .. code-block:: python

        my_stepsize = F.Stepsize(initial_value=1.2, stepsize_change_ratio=1.5, dt_min=1e-6, max_stepsize=lambda t: 1 if t < 1 else 2)
```

festim/stepsize.py
RemDelaporteMathurin, line 1: This is very nice and functional.

However, to simplify the code, could we refactore things slightly.

Instead of having logic in `adapt` (testing if `t_stop` is not None etc.), could we have an implementation purely with `max_stepsize` (the new feature), and upstream, we have (maybe in `__init__`?) some logic that tests if `t_stop` and `stepsize_stop_max` are not None and then produces the correct `max_stepsize`

Something like:

```python
if t_stop or stepsize_stop_max:
    # <-- raise deprecation warning here!
    self. max_stepsize = lambda t: stepsize_stop_max if t > t_stop
```

What do you think?

test/unit/test_stepsize.py
```diff
@@ -38,6 +38,20 @@ def test_hit_stepsize_max(self, my_stepsize):
         new_value = float(my_stepsize.value)
         assert new_value == my_stepsize.adaptive_stepsize["stepsize_stop_max"]
 
+    def test_hit_stepsize_max_float(self, my_stepsize):
```
RemDelaporteMathurin, line 41:
```suggestion
    def test_hit_stepsize_max_float(self, my_stepsize):
        """
        Assign an initial value to the stepsize, then call adapt at t > t_stop and checks that the new value is equal to max_stepsize
        """
```

test/unit/test_stepsize.py
```diff
@@ -38,6 +38,20 @@ def test_hit_stepsize_max(self, my_stepsize):
...
+        my_stepsize.value.assign(10)
+        my_stepsize.adaptive_stepsize["max_stepsize"] = 1
+        my_stepsize.adapt(t=6, converged=True, nb_it=2)
+        new_value = float(my_stepsize.value)
+        assert new_value == my_stepsize.adaptive_stepsize["max_stepsize"]
+
+    def test_hit_stepsize_max_callable(self, my_stepsize):
+        my_stepsize.value.assign(10)
+        my_stepsize.adaptive_stepsize["max_stepsize"] = lambda t: t * 0 + 1
```
RemDelaporteMathurin, line 50: Do you need t*0 here? Would this work?
```suggestion
        my_stepsize.adaptive_stepsize["max_stepsize"] = lambda t: 1
```

test/unit/test_stepsize.py
```diff
@@ -38,6 +38,20 @@ def test_hit_stepsize_max(self, my_stepsize):
...
         assert new_value == my_stepsize.adaptive_stepsize["stepsize_stop_max"]
 
+    def test_hit_stepsize_max_float(self, my_stepsize):
+        my_stepsize.value.assign(10)
+        my_stepsize.adaptive_stepsize["max_stepsize"] = 1
+        my_stepsize.adapt(t=6, converged=True, nb_it=2)
+        new_value = float(my_stepsize.value)
+        assert new_value == my_stepsize.adaptive_stepsize["max_stepsize"]
+
+    def test_hit_stepsize_max_callable(self, my_stepsize):
```
RemDelaporteMathurin, line 48:
```suggestion
    def test_hit_stepsize_max_callable(self, my_stepsize):
         """
         Assign an initial value to the stepsize and a callable for max_stepsize and checks that the new value is equal to max_stepsize
         """
```

test/unit/test_stepsize.py
```diff
@@ -38,6 +38,20 @@ def test_hit_stepsize_max(self, my_stepsize):
...
+        my_stepsize.value.assign(10)
+        my_stepsize.adaptive_stepsize["max_stepsize"] = lambda t: t * 0 + 1
+        my_stepsize.adapt(t=6, converged=True, nb_it=2)
+        new_value = float(my_stepsize.value)
+        assert new_value == my_stepsize.adaptive_stepsize["max_stepsize"](1)
```
RemDelaporteMathurin, line 53: if `t=6` in the adapt call why use `(1)` here? I know it doesn't matter but i would try to be consistent

RemDelaporteMathurin (changes requested):
Hi @KulaginVladimir thanks for this!

I made a few comments + refactoring idea. Tell me what you think!

docs/source/userguide/stepsize.rst
KulaginVladimir: Cool! I'll check

festim/stepsize.py
KulaginVladimir: Cool as well. Will do

KulaginVladimir: Meanwhile, with this change, do we need to keep `t_stop` and `stepsize_stop_max` in the `adaptive_stepsize` dictionary or we can just remove them from everywhere apart from `__init__` and correspondingly update tests? I mean, prepare the code a bit for a future update.

RemDelaporteMathurin: I think we can remove them from this dict

test/unit/test_stepsize.py
KulaginVladimir: Added

test/unit/test_stepsize.py
KulaginVladimir: I updated the test

RemDelaporteMathurin (changes requested):
just a small comment for code readability

RemDelaporteMathurin (approved):
LGTM! Thanks @KulaginVladimir 

As usual, I'll leave the PR open a little while longer to give everyone the chance to jump in

PR #703, opened 2024-03-05
Total volume quantity

RemDelaporteMathurin:
## Proposed changes

This PR adds the total volume quantity in exports

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/exports/volume_quantity.py
+++ b/festim/exports/volume_quantity.py
@@ -1,0 +1,70 @@
+import festim as F
+import csv
+import os
+
+
+class VolumeQuantity:
+    """Export SurfaceQuantity
+
+    Args:
+        field (festim.Species): species for which the surface flux is computed
+        surface (festim.VolumeSubdomain): volume subdomain
+        filename (str, optional): name of the file to which the surface flux is exported
+
+    Attributes:
+        field (festim.Species): species for which the surface flux is computed
+        volume (festim.VolumeSubdomain): volume subdomain
+        filename (str): name of the file to which the surface flux is exported
+        t (list): list of time values
+        data (list): list of values of the surface quantity
+    """
+
+    def __init__(self, field, volume, filename: str = None) -> None:
+        self.field = field
+        self.volume = volume
+        self.filename = filename
+
+        self.t = []
+        self.data = []
+
+    @property
+    def filename(self):
+        return self._filename
+
+    @filename.setter
+    def filename(self, value):
+        if value is None:
+            self._filename = None
+        elif not isinstance(value, str):
+            raise TypeError("filename must be of type str")
+        elif not value.endswith(".csv") and not value.endswith(".txt"):
+            raise ValueError("filename must end with .csv or .txt")
+        self._filename = value
+
+    @property
+    def field(self):
+        return self._field
+
+    @field.setter
+    def field(self, value):
+        # check that field is festim.Species
+        if not isinstance(value, (F.Species, str)):
+            raise TypeError("field must be of type festim.Species")
+
+        self._field = value
+
+    def write(self, t):
+        """If the filename doesnt exist yet, create it and write the header,
+        then append the time and value to the file"""
+
+        if not os.path.isfile(self.filename):
+            title = "Flux surface {}: {}".format(self.surface.id, self.field.name)
+
+            if self.filename is not None:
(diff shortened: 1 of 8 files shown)


test/test_total_volume.py
```diff
@@ -0,0 +1,68 @@
...
+from dolfinx import fem
+import pytest
+import ufl
+import os
+
+dummy_mat = F.Material(D_0=1, E_D=1, name="dummy")
+
+
+def test_total_volume_export_compute():
+    """Test that the surface flux export computes the correct value"""
```
jhdark, line 14:
```suggestion
    """Test that the total volume export computes the correct value"""
```

festim/exports/volume_quantity.py
```diff
@@ -0,0 +1,70 @@
+import festim as F
+import csv
+import os
+
+
+class VolumeQuantity:
+    """Export VolumeQuantity
+
+    Args:
+        field (festim.Species): species for which the surface flux is computed
```
jhdark, line 10:
```suggestion
        field (festim.Species): species for which the volume quantity is computed
```

festim/exports/volume_quantity.py
```diff
@@ -0,0 +1,70 @@
...
+        surface (festim.VolumeSubdomain): volume subdomain
+        filename (str, optional): name of the file to which the surface flux is exported
```
jhdark, line 12:
```suggestion
        filename (str, optional): name of the file to which the volume quantity is exported
```

festim/exports/volume_quantity.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Attributes:
+        field (festim.Species): species for which the surface flux is computed
+        volume (festim.VolumeSubdomain): volume subdomain
+        filename (str): name of the file to which the surface flux is exported
```
jhdark, line 17: see above

festim/exports/volume_quantity.py
```diff
@@ -0,0 +1,70 @@
...
+            title = "Total volume {}: {}".format(self.volume.id, self.field.name)
+
+            if self.filename is not None:
+                with open(self.filename, mode="w", newline="") as file:
+                    writer = csv.writer(file)
+                    writer.writerow(["t(s)", f"{title}"])
+
+        with open(self.filename, mode="a", newline="") as file:
+            writer = csv.writer(file)
+            writer.writerow([t, self.value])
```
jhdark, line 70: I think we messed this up the previous time with surface quantity. In testing, I realised that when re-running scripts, instead of creating a new export file, it appends to a previous one if it exists. I'm not sure if we should fix this now or in another PR.

festim/hydrogen_transport_problem.py
```diff
@@ -808,6 +808,13 @@ def post_processing(self):
...
                 if export.filename is not None:
                     export.write(t=float(self.t))
+            elif isinstance(export, F.VolumeQuantity):
+                export.compute(self.dx)
+                # update export data
+                export.t.append(float(self.t))
 
+                # if filename given write export data to file
+                if export.filename is not None:
+                    export.write(t=float(self.t))
```
jhdark, line 818: Could test this

jhdark (changes requested):
There are some small copy-paste errors, but there is a larger bug at hand with the export file writing, appending a previous file of the same name rather than overwriting it.

festim/exports/volume_quantity.py
RemDelaporteMathurin: Let's fix it in another PR, it's as simple as changing `mode="a"` to `mode="w"`

festim/hydrogen_transport_problem.py
RemDelaporteMathurin: So I think this is gonna be refactored when we start adding more derived quantities anyway, which is why I didn't bother testing it here

RemDelaporteMathurin:
@jhdark I added the suggestions thanks for spotting those errors.
Since this is likely gonna be refactored in the near future, I didn't bother testing all the code duplication.

I also opened an issue to track the bug you found for appending to the files instead of overwriting.

jhdark:
Just needs a rebase with the new hot fix and then should be good to go

Issue #705, opened 2024-03-06
Derived quantities append to file instead of overwriting


RemDelaporteMathurin:
There are some small copy-paste errors, but there is a larger bug at hand with the export file writing, appending a previous file of the same name rather than overwriting it.

_Originally posted by @jhdark in https://github.com/festim-dev/FESTIM/pull/703#pullrequestreview-1920469938_

PR #706, opened 2024-03-06
Patch derived quantity overwriting csv file

RemDelaporteMathurin:
## Proposed changes

This is a patch to #705 

I just reworked the logic to check if it's the first time we write the csv and if so write the file and write the header, otherwise just append to it.

That requires to add another attribute `._first_time_export`, I don't know if we should document it or not but I added _ to say it's a private attribute. What do you think @jhdark ?

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/exports/surface_quantity.py
+++ b/festim/exports/surface_quantity.py
@@ -26,6 +26,7 @@ class SurfaceQuantity:
 
         self.t = []
         self.data = []
+        self._first_time_export = True
 
     @property
     def filename(self):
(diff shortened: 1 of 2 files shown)


jhdark:
Yeah don't think anything is needed in the doc strings for this hidden attribute as it should get a given, we could potentially add a note in the docs later down the line, but otherwise this is good

PR #709, opened 2024-03-14
Add roadmap to README and change layout

RemDelaporteMathurin:
This PR simply changes the README by changing the layout and adding the public roadmap for more visibility


jhdark:
It might be worth adding citations and linking the newly accepted reference paper. Plus we could potentially clump some of these subtitles to make it more readable, not sure

RemDelaporteMathurin:
> It might be worth adding a citation title linking the newly accepted reference paper. Plus we could potentially clump some of these subtitles to make it more readable, not sure

Should we do this once the paper is published?

jhdark (approved):
LGTM

PR #711, opened 2024-03-22
Units in derived quantities title

jhdark:
## Proposed changes

With this change, users can opt to have units in the titles of the derived quantities file. They can do this with a new argument within the `DerivedQuantities` class like so:
```python
my_derived_quantities = F.DerivedQuantities(
    [
        F.TotalVolume("solute", volume=1),
        F.AverageVolume("solute", volume=1),
        F.MaximumVolume("solute", volume=1),
        F.MinimumVolume("solute", volume=1),
        F.SurfaceFlux("solute", surface=2),
    ],
    filename="derived_quantities.csv",
    show_units=True,
)
```
The unit is then added to the derived quantity's title, accounting for the mesh's dimension. For example, should a user call for a `TotalVolume` of the `solute` field in a 1D mesh case with units, the title will be `Total solute volume 1 (H m-2) `, whereas in the same case but in 2D, the title will be' Total solute volume 1 (H m-1) `.

The original format of the titles is left unchanged so as not to introduce breaking changes; however, it could be up for debate to change them to make them more consistent.

I have also taken the liberty to add doc strings to all of the derived quantities

## Types of changes

- [x] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/exports/derived_quantities/maximum_surface.py
+++ b/festim/exports/derived_quantities/maximum_surface.py
@@ -14,6 +14,7 @@ class MaximumSurface(DerivedQuantity):
     def __init__(self, field, surface) -> None:
         super().__init__(field)
         self.surface = surface
+
         self.title = "Maximum {} surface {}".format(self.field, self.surface)
 
     def compute(self, surface_markers):


festim/exports/derived_quantities/average_surface.py
```diff
@@ -3,9 +3,40 @@
 
 
 class AverageSurface(SurfaceQuantity):
+    """
+    Computes the average value of a function on a given surface
```
RemDelaporteMathurin, line 7: You could add the equation here

```suggestion
    Computes the average value of a function on a given surface
    int( f dx) / int (1 * dx)
```

festim/exports/derived_quantities/average_surface.py
```diff
@@ -3,9 +3,40 @@
...
+
+    Notes:
+        Units are in H/m3 for hydrogen and K for temperature
+    """
+
     def __init__(self, field, surface) -> None:
         super().__init__(field=field, surface=surface)
-        self.title = "Average {} surface {}".format(self.field, self.surface)
+
+        self.show_units = False
```
RemDelaporteMathurin, line 28: This attribute lives in `DerivedQuantity`

festim/exports/derived_quantities/average_surface.py
```diff
@@ -3,9 +3,40 @@
...
+
+    @property
+    def title(self):
+        quantity_title = "Average {} surface {}".format(self.field, self.surface)
```
RemDelaporteMathurin, line 32: we could use a f-string here

festim/exports/derived_quantities/derived_quantities.py
```diff
@@ -105,6 +109,8 @@ def filename(self, value):
     def make_header(self):
         header = ["t(s)"]
         for quantity in self:
+            if self.show_units:
+                quantity.show_units = True
```
RemDelaporteMathurin, line 113:
```suggestion
            quantity.show_units = self.show_units
```

festim/exports/derived_quantities/maximum_surface.py
```diff
@@ -1,20 +1,43 @@
-from festim import DerivedQuantity
+from festim import SurfaceQuantity
 import fenics as f
 import numpy as np
 
 
-class MaximumSurface(DerivedQuantity):
+class MaximumSurface(SurfaceQuantity):
     """
+    Computes the maximum value of a function on a given surface
```
RemDelaporteMathurin, line 8: should we replace function by field?

festim/exports/derived_quantities/point_value.py
```diff
@@ -2,20 +2,45 @@
...
+    Attributes:
+        field (str, int):  the field ("solute", 0, 1, "T", "retention")
+        x (int, float, tuple, list): the point coordinates
+        show_units (bool): show the units in the title in the derived quantities
+            file
+        function (dolfin.function.function.Function): the solution function of
+            the field
+
+    Notes:
+        Units are in H/m3 for hydrogen and K for temperature
```
RemDelaporteMathurin, line 21:
```suggestion
        Units are in H/m3 for hydrogen concentration and K for temperature
```
"hydrogen" isn't a quantity

festim/exports/derived_quantities/surface_flux.py
```diff
@@ -3,24 +3,62 @@
...
+        surface (int): the surface id
+        show_units (bool): show the units in the title in the derived quantities
+            file
+        title (str): the title of the derived quantity
+        function (dolfin.function.function.Function): the solution function of
+            the hydrogen solute field
+
+    Notes:
         Object to compute the flux J of a field u through a surface
         J = integral(-prop * grad(u) . n ds)
```
RemDelaporteMathurin, line 24: We discussed this, and this is not what's implemented

festim/exports/derived_quantities/surface_flux.py
```diff
@@ -3,24 +3,62 @@
...
+
+        self.show_units = show_units
+
+    @property
+    def title(self):
+        quantity_title = "Flux surface {}: {}".format(self.surface, self.field)
+        if self.show_units:
+            # obtain domain dimension
+            dim = self.function.function_space().mesh().topology().dim()
+            if "T" in self.field:
```
RemDelaporteMathurin, line 46:
```suggestion
            if self.field == "T":
```

festim/exports/derived_quantities/surface_flux.py
```diff
@@ -3,24 +3,62 @@
...
-        """
+    """
+    Computes the surface flux of a field at a given surface
+
+    Args:
+        field (str, int): the field
+        surface (int): the surface id
+
+    Attribtutes
+        field (str, int): the hydrogen solute field
```
RemDelaporteMathurin, line 14: "hydrogen solute field" it can also be temperature right

festim/exports/derived_quantities/thermal_flux.py
```diff
@@ -2,5 +2,40 @@
...
 class ThermalFlux(SurfaceFlux):
+    """
+    Computes the surface flux of heat at a given surface
+
+    Args:
+        surface (int): the surface id
+
+    Attribtutes
+        surface (int): the surface id
+        field (str, int): the temperature field
```
RemDelaporteMathurin, line 13: cannot be int if it's temperature

festim/exports/derived_quantities/total_surface.py
```diff
@@ -3,9 +3,54 @@
...
+        field (str, int): the field
+        surface (int): the surface id
+        show_units (bool): show the units in the title in the derived quantities
+            file
+        title (str): the title of the derived quantity
+        function (dolfin.function.function.Function): the solution function of
+            the hydrogen solute field
+
+    Notes:
+        units are in H/m2/s in 1D, H/m/s in 2D and H/s in 3D domains
```
RemDelaporteMathurin, line 23: Noup.

test/unit/test_exports/test_derived_quantities/test_derived_quantities.py
```diff
@@ -64,6 +78,46 @@ def test_all_quantities(self):
         ]
         assert header == expected_header
 
+    def test_with_units_simple(self):
```
RemDelaporteMathurin, line 81: We might want to have a more explicit way of testing this.

Imagine that the title of each property is "coucou", this test would pass.

Maybe not here but somewhere you need to test that the title is say "Flux surface X" and not "coucou"

RemDelaporteMathurin (changes requested):
Great addition @jhdark much needed

First round of comments

festim/exports/derived_quantities/maximum_surface.py
jhdark: yep sounds better

test/unit/test_exports/test_derived_quantities/test_derived_quantities.py
jhdark: Yeah I agree

festim/exports/derived_quantities/average_surface.py
```diff
@@ -3,9 +3,39 @@
 
 
 class AverageSurface(SurfaceQuantity):
+    """
+    Computes the average value of a field on a given surface
+    int( f dx) / int (1 * dx)
```
RemDelaporteMathurin, line 8:
```suggestion
    int( f ds) / int (1 * ds)
```

festim/exports/derived_quantities/average_volume.py
```diff
@@ -3,9 +3,38 @@
 
 
 class AverageVolume(VolumeQuantity):
+    """
+    Computes the average value of a field in a given volume
```
RemDelaporteMathurin, line 7: We should add the formula like in AverageSurface but with dx (I maybe got mixed up in my previous review)

festim/exports/derived_quantities/derived_quantities.py
```diff
@@ -105,6 +109,7 @@ def filename(self, value):
     def make_header(self):
         header = ["t(s)"]
         for quantity in self:
+            quantity.show_units = self.show_units
```
RemDelaporteMathurin, line 112: Should we add a deprecation warning here if `self.show_units` is `False`? Warning the users that the new behaviour in a couple of releases will be to show units

festim/exports/derived_quantities/point_value.py
```diff
@@ -2,20 +2,43 @@
...
+    @property
+    def title(self):
+        quantity_title = f"{self.field} value at {self.x}"
+        if self.show_units:
+            if self.field == "T":
+                return quantity_title + " (K)"
+            else:
+                return quantity_title + " (H m-3)"
+        else:
+            return quantity_title
```
RemDelaporteMathurin, line 41: I think this is redundant since it is already the behaviour of `SurfaceFlux`

```suggestion

```

festim/exports/derived_quantities/surface_flux.py
RemDelaporteMathurin: @jhdark can you address this

festim/exports/derived_quantities/surface_flux.py
```diff
@@ -3,24 +3,60 @@
 
 
 class SurfaceFlux(SurfaceQuantity):
-    def __init__(self, field, surface) -> None:
-        """
+    """
+    Computes the surface flux of a field at a given surface
```
RemDelaporteMathurin, line 7:
```suggestion
    Computes the surface flux of a field at a given surface in cartesian coordinates
```

festim/exports/derived_quantities/surface_flux.py
```diff
@@ -3,24 +3,60 @@
...
+        units are in H/m2/s in 1D, H/m/s in 2D and H/s in 3D domains for hydrogen
+        concentration and W/m2 in 1D, W/m in 2D and W in 3D domains for temperature
 
-        Args:
-            field (str, int):  the field ("solute", 0, 1, "T", "retention")
-            surface (int): the surface id
-        """
+    """
+
+    def __init__(self, field, surface, show_units=False) -> None:
```
RemDelaporteMathurin, line 35:
```suggestion
    def __init__(self, field, surface) -> None:
```

There is no reason for `show_units` to be an argument here. Proof is, you don't use it.

festim/exports/derived_quantities/thermal_flux.py
```diff
@@ -2,5 +2,40 @@
...
+            # obtain domain dimension
+            dim = self.function.function_space().mesh().topology().dim()
+            if dim == 1:
+                return quantity_title + " (W m-2)"
+            if dim == 2:
+                return quantity_title + " (W m-1)"
+            if dim == 3:
+                return quantity_title + " (W)"
+        else:
+            return quantity_title
```
RemDelaporteMathurin, line 41: same comment as before, I don't think this is needed as it's redundant with SurfaceFlux

festim/exports/derived_quantities/total_surface.py
```diff
@@ -3,9 +3,53 @@
 
 
 class TotalSurface(SurfaceQuantity):
+    """
+    Computes the total value of a field on a given surface
```
RemDelaporteMathurin, line 7: Let's add the equation here

int(f ds)

festim/exports/derived_quantities/total_volume.py
```diff
@@ -3,9 +3,54 @@
 
 
 class TotalVolume(VolumeQuantity):
+    """
+    Computes the total value of a field in a given volume
```
RemDelaporteMathurin, line 7: Let's add the equation here

int(f dx)

festim/exports/exports.py
```diff
@@ -133,6 +133,6 @@ def initialise_derived_quantities(self, dx, ds, materials):
         """
         for export in self:
             if isinstance(export, festim.DerivedQuantities):
-                export.data = [export.make_header()]
+                export.data = []
```
RemDelaporteMathurin, line 136: Should we do the same for `export.t` 
```suggestion
                export.data = []
                export.t = []
```

test/unit/test_exports/test_derived_quantities/test_derived_quantities.py
```diff
@@ -179,7 +233,7 @@ def test_simple(self):
 
         my_derv_quant.data = []
         my_derv_quant.compute(t)
-        assert my_derv_quant.data[0] == expected_data
+        assert my_derv_quant.data[1] == expected_data
```
RemDelaporteMathurin, line 236: Why is this change needed?

RemDelaporteMathurin (changes requested):
- Now that #689 is merged please rebase and make the according changes for the new classes
- The titles of the quantities isn't consistent. For example, `SurfaceFlux` is "Flux surface X: solute (H s-1)" but `TotalVolume` is "Total solute surface X (H m-1)". Please homogeneise them (only when show_units is True so that it's not a breaking change)

RemDelaporteMathurin:
@jhdark can you push a dummy commit like add a comment somewhere to reopen this?

jhdark:
Starting new PR as branch is not recoverable

Issue #720, opened 2024-03-27
Arbitrary number of reactants for `Reaction`


RemDelaporteMathurin:
Currently `Reaction` only accepts 2 reacting species.

For more flexibility, we should allow to have an arbitrary number of reacting species.
This will help us implement reactions like radioactive decay where there's only one reactant and no product.

PR #730, opened 2024-04-01
Arbitrary reaction

Allentro:
## Proposed changes

Enabled an arbitrary number of reactants and products. This resolves https://github.com/festim-dev/FESTIM/issues/720

I've enabled it so that The number of products can be 0 for radioactive decay (A -> None), but the number of reactants must be at least length 1.

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted


RemDelaporteMathurin (changes requested):
Hi @Allentro thanks for this! This will bring more flexibility to Reactions.

I have a couple of initial comments

festim/reaction.py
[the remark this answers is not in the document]
Allentro: This is a good point!

In the implementation I allowed: 
`A -> None `
but not: 
`None -> B `
I was cautious to prevent species production, but maybe it's better to allow 
`None -> B ` checking that `k_0=0` and `E_k=0`
and 
`A -> None` checking that `p_0=0` and `E_p=0`

What do you think?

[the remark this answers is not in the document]
RemDelaporteMathurin: I think an implementation that makes sense would be allowing 
A -> None 
In that case, k_0 and E_k are None (optional)
If users provide values for k_0 and E_k, an error is raised ("you provided a k but you didn't provide a product)

I don't think it makes sense to require users to set k=0 explicitely in this case

festim/reaction.py
```diff
@@ -113,16 +114,29 @@ def reaction_term(self, temperature):
...
-            products_of_product *= product.solution
+        if len(products) > 0:
+            product_of_products = products[0].solution
+            for product in products[1:]:
+                product_of_products *= product.solution
+        else:
+            product_of_products = 0
 
-        return k * c_A * c_B - p * products_of_product
+        return k * product_of_reactants - p * product_of_products
```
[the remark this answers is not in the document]
Allentro: Yeah, they print the same. Good check though.

festim/reaction.py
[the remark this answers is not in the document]
Allentro: k_0 and E_k are the forward components. Is it not the case that if product=None, p_0 and E_p should both be 0?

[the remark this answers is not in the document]
RemDelaporteMathurin: Sorry, in all I said replace k by p (the backward rate is None if there are no products)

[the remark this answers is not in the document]
Allentro: Okay makes sense. I've added that into the PR with tests.

festim/hydrogen_transport_problem.py
```diff
@@ -655,20 +655,19 @@ def create_formulation(self):
...
-                self.formulation += (
-                    reaction.reaction_term(self.temperature_fenics)
-                    * reaction.reactant2.test_function
-                    * self.dx(reaction.volume.id)
-                )
+            # reactant
+            if isinstance(reaction.reactant, list):
+                reactants = reaction.reactant
+            else:
+                reactants = [reaction.reactant]
```
RemDelaporteMathurin, line 662: No need for this, `.reactant` is always a list

festim/reaction.py
```diff
@@ -113,16 +125,20 @@ def reaction_term(self, temperature):
...
-        for product in products[1:]:
-            products_of_product *= product.solution
-
-        return k * c_A * c_B - p * products_of_product
+        if len(products) > 0:
+            product_of_products = products[0].solution
+            for product in products[1:]:
+                product_of_products *= product.solution
+        else:
+            product_of_products = 0
```
RemDelaporteMathurin, line 143: This line isn't tested. Please either right a test to catch it or write a #TODO to remind ourselves to test it later

test/test_reaction.py
```diff
@@ -128,6 +147,28 @@ def test_reaction_str_2_products():
...
+    # create a reaction between the two species
+    reaction = F.Reaction(
+        species1,
+        None,
+        k_0=1.0,
+        E_k=0.2,
+        p_0=0,
+        E_p=0,
+        volume=my_vol,
+    )
```
RemDelaporteMathurin, line 165: Here for instance, I'd rather have

```python
    reaction = F.Reaction(
        reactant=species1,
        k_0=1.0,
        E_k=0.2,
        volume=my_vol,
    )
```

test/test_reaction.py
```diff
@@ -227,19 +266,50 @@ def test_reactant1_setter_raises_error_with_wrong_type():
...
+    ):
+        F.Reaction(
+            reactant=[F.Species("A")],
+            product=None,
+            k_0=1,
+            E_k=0.1,
+            p_0=0,
+            E_p=2,
+            volume=my_vol,
+        )
```
RemDelaporteMathurin, line 315: So here is a good example of why this could be more straightforward.

Here, if p_0 = 0, then E_p could be equal to any value and it wouldn't matter.

Therefore, I think it's more straightforward to have p_0 and E_p default to `None` in `Reaction` instead of making sure they are equal to zero.

RemDelaporteMathurin (changes requested):

festim/reaction.py
```diff
@@ -48,61 +45,76 @@ class Reaction:
...
     ) -> None:
-        self.reactant1 = reactant1
-        self.reactant2 = reactant2
-        self.product = product
         self.k_0 = k_0
         self.E_k = E_k
-        self.p_0 = p_0
-        self.E_p = E_p
+        self.p_0 = p_0 or 0.0
+        self.E_p = E_p or 0.0
```
RemDelaporteMathurin, line 61:
```suggestion
        self.p_0 = p_0
        self.E_p = E_p
```

festim/reaction.py
```diff
@@ -48,61 +45,76 @@ class Reaction:
...
-        product: F.Species,
+        reactant: Union[
+            F.Species, F.ImplicitSpecies, List[Union[F.Species, F.ImplicitSpecies]]
+        ],
+        product: Optional[Union[F.Species, List[F.Species]]],
         k_0: float,
         E_k: float,
         p_0: float,
         E_p: float,
         volume: F.VolumeSubdomain1D,
```
RemDelaporteMathurin, line 56:
```suggestion
        volume: F.VolumeSubdomain1D,
        p_0: float = None,
        E_p: float = None,

```

festim/reaction.py
```diff
@@ -48,61 +45,76 @@ class Reaction:
...
+    def product(self, value):
+        if value == []:
+            if self.p_0 != 0.0:
+                raise ValueError(
+                    f"p_0 must be 0, not {self.p_0} when no products are present."
+                )
+            elif self.E_p != 0.0:
+                raise ValueError(
+                    f"E_p must be 0, not {self.E_p} when no products are present."
+                )
```
RemDelaporteMathurin, line 99: This test should be here but in `reaction_term`.

Consider:

```python
my_reac = F.Reaction(reactant=A, product=[], p_0=1, E_p=1)  # raises an error
my_reac = F.Reaction(reactant=A, product=B, p_0=None, E_p=None) # doesn't raise an error
```

You would think you just have to make a setter for `p_0` and `E_p` but then you end up having a cyclic loop.

Instead, don't do this test in the setters but in `reaction_term()`

RemDelaporteMathurin (changes requested):
@Allentro here are a few suggestions:
- make p_0 and E_p optional arguments defaulting to `None`
- test that 1) if a product is given, p_0 and E_p are not None, and 2) that if no product is given, p_0 and E_p _are_ None.

festim/reaction.py
Allentro: Aha okay, good point. I was trying to avoid putting the test in `reaction_term()` as that is only called during the actual solve and not in instantiation of the class?

RemDelaporteMathurin: The problem with doing it only at instantiation is that you can have it right on instantiation and then modify it afterwards (between instantiation and the call to reaction_term()) and break things

Allentro: Good call.

RemDelaporteMathurin (approved):

PR #733, opened 2024-04-02
Reset derived quantity objects data and t at the beginning of a simulation

### LINKED ISSUE #728: Derived quantity objects `data` and `t` attributes are not emptied at beginning of simulation
**Describe the bug**
When running the same simulation twice with derived quantities, their `t` and `data` attributes isn't reset to an empty list but instead values are just appended.

**To Reproduce**
Run the following on `fenicsx`

```python
import festim as F

my_model = F.HydrogenTransportProblem()

my_model.mesh = F.Mesh1D([0, 1, 2, 3, 4])

tungsten = F.Material(D_0=1, E_D=0)

my_subdomain = F.VolumeSubdomain1D(id=1, borders=[0, 4], material=tungsten)
left_surface = F.SurfaceSubdomain1D(id=1, x=0)

my_model.subdomains = [my_subdomain, left_surface]

mobile_H = F.Species("H")
my_model.species = [mobile_H]

my_model.temperature = 400

left_flux = F.SurfaceFlux(field=mobile_H, surface=left_surface)

my_model.exports = [left_flux]

    atol=1e-15,
    rtol=1e-15,
    transient=False,
)

for i in range(3):
    my_model.initialise()
    my_model.run()

    print(left_flux.data)

```

Produces:

```
[0.0]
[0.0, 0.0]
[0.0, 0.0, 0.0]
```

**Expected behavior**
left_flux.data should be of length 1 everytime

Allentro:
## Proposed changes

I have implemented a fix to reset export.t and export.data at the beginning of a simulation, accompanied by a respective test. This PR resolves #728 

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted


festim/hydrogen_transport_problem.py
```diff
@@ -324,8 +325,9 @@ def define_temperature(self):
     def initialise_exports(self):
         """Defines the export writers of the model, if field is given as
         a string, find species object in self.species"""
-
         for export in self.exports:
+            export.t = []
+            export.data = []
```
RemDelaporteMathurin, line 330: Doing it here is not the right place.
This iterates through all the exports, which can be derived quantities, but also VTXExport or XDMFExport etc.

Instead, we could do it in `write()`, making use of the `self._is_first_time_export` attribute. For example here:

https://github.com/festim-dev/FESTIM/blob/f08f0b5c83d11b511a129316d040b61d0bf4f3b1/festim/exports/surface_quantity.py#L75-L76

RemDelaporteMathurin: Alternatively we could do it here, but with an additional type check to only do this for derived quantities.

test/test_h_transport_problem.py
```diff
@@ -337,6 +337,35 @@ def test_initialise_exports_multiple_exports_same_species():
...
+def test_export_resets_quantities():
+    """Test that the export.data and export.t are correctly reset every time a simulation is initiated."""
+    my_mat = F.Material(D_0=1, E_D=0)
+    H = F.Species("H")
+    surf = F.SurfaceSubdomain1D(id=1, x=4)
+
+    my_export = F.SurfaceFlux(
+        field=H,
+        surface=surf,
+    )
```
RemDelaporteMathurin, line 349: Can you parametrise this test to also check for other derived quantities (eg. TotalVolume)

RemDelaporteMathurin (changes requested):
Initial round of comments, the fix obviously works fine, but it's not the right place for it.

RemDelaporteMathurin:
Hey @Allentro what is the status of this?

Allentro:
> Hey @Allentro what is the status of this?

Hi Remi, sorry this has slipped my list. I'll get this complete tomorrow!

Allentro:
> Hi @Allentro! Do you want someone else to pick up this PR?

Hi Remi, 

This slipped my inbox.. I was just scrolling through my mentions and saw this. I'm happy to work on this tomorrow evening. Sorry for the delay on this.

RemDelaporteMathurin (approved):

Issue #673, opened 2023-12-14
Custom `Solver`


KulaginVladimir:
Following the [discussion](https://festim.discourse.group/t/convergence-plot/23/2), it might be useful to export some metrics, that can characterise the overall error estimation of the solution (or something else), vs. time step (solver iteration). For example, there is some [information](https://doc.comsol.com/5.5/doc/com.comsol.help.comsol/comsol_ref_solver.27.089.html) on "convergence plots" in COMSOL.

RemDelaporteMathurin:
Pinging @gabriele-ferrero

RemDelaporteMathurin:
I'm still struggling to understand exactly what would this convergence plot show. Is it:

at each time step, we record the final error (absolute and relative) of the Newton solver and keep track of it?

or

at each time step, we show a plot with the Newton iterations in X and the errors in Y

?

KulaginVladimir:
@RemDelaporteMathurin 
I guess, for a time-dependent problem the final absolute/relative tolerance at each time-step is OK. Another option is to plot 1/time_step vs. time-step if adaptive time-step is used. According to the formulation of the adaptive time-step, it also can indicate that the problem converges.  

Another question is related to stationary problems, where there are no time-steps. However, I suppose it is not really needed, but absolute/relative tolerance vs. Newton iterations can be an option. 

What do you think?

RemDelaporteMathurin:
This sounds very good to me and I don't think it would be super hard to implement.
Maybe we can start with a prototype in fenics to see how it's done and then think of how to integrate with FESTIM

KulaginVladimir:
Don't know about not super hard) Possibly, the shown below is not we need.

As I see from [dolfin](https://github.com/live-clones/dolfin/tree/master/dolfin), `NonlinearVariationalSolver` (used for solving in FESTIM) creates `Newton solver` on a call, which in turn returns `(_newton_iteration, newton_converged)`. It appears to me that one or both of these classes have to be change to get the info from `Newton solver`. 

I found [hint1](https://fenicsproject.discourse.group/t/set-krylov-linear-solver-paramters-in-newton-solver/1070/3) and [hint2](https://fenicsproject.discourse.group/t/return-relative-residual-from-newton-solver/7737) on how problem and solver can be adopted for the needs. Maybe, they can help. 

[...]

PR #734, opened 2024-04-03
Custom solver

KulaginVladimir:
## Proposed changes

Following #673, this PR introduces:

- the possibility for the users to set the Newton solver preconditioner via `F.Settings(preconditioner="...")`
- the possibilty for the users to set a custom Newton solver after the model (`HeatTransferProblem`) initialisation. 

The expected usage of a custom Newton solver is:
```python

my_model.initialise()

class CustomSolver(fenics.NewtonSolver):
     .....

custom_solver = CustomSolver()
custom_solver.parameters["..."] = ...

my_model.h_transport_problem.newton_solver = custom_solver

my_model.run()
``` 
The same can be done for for the `F.ExtrinsicTrap` class:
```python
my_model.traps = [F.ExtrinsicTrap(...), ...]
for trap in my_model.traps:
    if isinstance(trap, F.ExtrinsicTrap)
        trap.newton_solver = CustomSolver()
``` 
and for the `HeatTransferProblem` (but before `my_model.initialise()` for the steady-state case!):

```python
my_model.T = F.HeatTransferProblem(...)
my_model.T.newton_solver = CustomSolver()
``` 

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/concentration/traps/extrinsic_trap.py
+++ b/festim/concentration/traps/extrinsic_trap.py
@@ -1,5 +1,5 @@
 from festim import Trap, as_constant_or_expression
-from fenics import NewtonSolver
+from fenics import NewtonSolver, MPI
 
 
 class ExtrinsicTrapBase(Trap):
@@ -57,7 +57,7 @@ class ExtrinsicTrapBase(Trap):
 
     def define_newton_solver(self):
         """Creates the Newton solver and sets its parameters"""
-        self.newton_solver = NewtonSolver()
+        self.newton_solver = NewtonSolver(MPI.comm_world)
         self.newton_solver.parameters["error_on_nonconvergence"] = False
         self.newton_solver.parameters["absolute_tolerance"] = self.absolute_tolerance
         self.newton_solver.parameters["relative_tolerance"] = self.relative_tolerance
--- a/festim/concentration/traps/extrinsic_trap.py
+++ b/festim/concentration/traps/extrinsic_trap.py
@@ -1,4 +1,5 @@
 from festim import Trap, as_constant_or_expression
+from fenics import NewtonSolver
 
 
 class ExtrinsicTrapBase(Trap):
@@ -14,6 +15,7 @@ class ExtrinsicTrapBase(Trap):
         relative_tolerance=1e-10,
         maximum_iterations=30,
         linear_solver=None,
+        preconditioner=None,
         **kwargs,
     ):
         """Inits ExtrinsicTrap
@@ -36,17 +38,32 @@ class ExtrinsicTrapBase(Trap):
                 If None, the default fenics linear solver will be used ("umfpack").
                 More information can be found at: https://fenicsproject.org/pub/tutorial/html/._ftut1017.html.
                 Defaults to None.
+            preconditioner (str, optional): preconditioning method for the newton solver,
+                options can be veiwed by print(list_krylov_solver_preconditioners()).
+                Defaults to None.
         """
         super().__init__(k_0, E_k, p_0, E_p, materials, density=None, id=id)
         self.absolute_tolerance = absolute_tolerance
         self.relative_tolerance = relative_tolerance
         self.maximum_iterations = maximum_iterations
         self.linear_solver = linear_solver
+        self.preconditioner = preconditioner
+        self.newton_solver = None
 
         for name, val in kwargs.items():
             setattr(self, name, as_constant_or_expression(val))
         self.density_previous_solution = None
         self.density_test_function = None
+
+    def define_newton_solver(self):
+        """Creates the Newton solver and sets its parameters"""
+        self.newton_solver = NewtonSolver()
+        self.newton_solver.parameters["error_on_nonconvergence"] = False
+        self.newton_solver.parameters["absolute_tolerance"] = self.absolute_tolerance
+        self.newton_solver.parameters["relative_tolerance"] = self.relative_tolerance
+        self.newton_solver.parameters["maximum_iterations"] = self.maximum_iterations
+        self.newton_solver.parameters["linear_solver"] = self.linear_solver
+        self.newton_solver.parameters["preconditioner"] = self.preconditioner
 
 
 class ExtrinsicTrap(ExtrinsicTrapBase):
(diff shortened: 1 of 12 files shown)


KulaginVladimir:
One thing to mention is that the `info` messages, shown when `log_level=20`, differ with `f.NewtonSolver`, since `f.NonlinearVariationalSolver` [prints](https://bitbucket.org/fenics-project/dolfin/src/1c52e837eb54cc34627f90bde254be4aa8a2ae17/dolfin/fem/NonlinearVariationalSolver.cpp#lines-48) additional messages. A possible workaround is to invoke `f.info` before sovling the problem in FESTIM.

RemDelaporteMathurin:
@KulaginVladimir thanks for this I think it is much needed to give more flexibility.

Would you mind listing the differences - if any - from the user point of view when running the default behaviour?
Are there any differences in absolute errors, usage, etc.

RemDelaporteMathurin:
Also tagging @Allentro as he was interested in this feature

KulaginVladimir:
As I tested, there shouldn't appear significant differences (in errors or usege) with these changes if, for example, `my_model.h_transport_problem.newton_solver` is not provided manually. 

During the model initialisation, a default `fenics.NewtonSolver` object is created and the solver parameters are set from `my_model.settings`: `absolute_tolerance`, `relative_tolerance`, `maximum_iterations`, `linear_solver`, and `preconditioner`, which were kept unchanged (the default preconditioner is set to `default`). Together with the introduced `festim.Problem` class, new changes should reproduce the solbing procedure of `f.NonlinearVariationalProblem` and `f.NonlinearVariationalSolver`, used in the current version of FESTIM. 

As I mentioned above, there is a change in fenics logs (i.e. when `log_level=20`). When `NewtonSolver.solve` is called, it doesn't print "Solving nonlinear variational problem.", whereas `f.NonlinearVariationalSolver` does.

@RemDelaporteMathurin is this what you asked? I also wonder about an example of using a custom solver. Shall we add it somewhere?

festim/concentration/traps/traps.py
```diff
@@ -120,7 +120,12 @@ def solve_extrinsic_traps(self):
                 du_t = f.TrialFunction(trap.density[0].function_space())
                 J_t = f.derivative(trap.form_density, trap.density[0], du_t)
                 problem = festim.Problem(J_t, trap.form_density, [])
+
+                f.begin(
+                    "Solving nonlinear variational problem."
+                )  # Add message to fenics logs
                 trap.newton_solver.solve(problem, trap.density[0].vector())
+                f.end()
```
KulaginVladimir, line 128: So, I added `fenics.begin()`/`fenics.end()` (see [decription](https://fenicsproject.org/olddocs/dolfin/1.4.0/python/programmers-reference/cpp/common/begin.html)) to increase the indentation level of info messages printed by the Newton solver.

RemDelaporteMathurin:
I would like to keep this open while we test it thoroughly. Playing with the solver can be a bit dangerous so I want to make sure we're not merging any breaking stuff.

festim/concentration/traps/extrinsic_trap.py
```diff
@@ -36,18 +38,33 @@ def __init__(
                 If None, the default fenics linear solver will be used ("umfpack").
                 More information can be found at: https://fenicsproject.org/pub/tutorial/html/._ftut1017.html.
                 Defaults to None.
+            preconditioner (str, optional): preconditioning method for the newton solver,
+                options can be veiwed by print(list_krylov_solver_preconditioners()).
```
RemDelaporteMathurin, line 42:
```suggestion
                options can be viewed by print(list_krylov_solver_preconditioners()).
```

RemDelaporteMathurin (changes requested):
Hi @KulaginVladimir my deepest apologies for taking so much time reviewing this.

I tested it and it seems to work perfectly well!

Several points:
- If users specify a custom solver for say the H transport problem, then the parameters given in `Settings` (absolute, relative tolerances, etc.) will be ignored. This is ok, but we should add a print statement to inform the users
- Can be done in a separate PR but it would be nice to add some documentation and maybe a short demo in the workshop?

festim/concentration/traps/extrinsic_trap.py
```diff
@@ -55,6 +56,23 @@ def __init__(
...
+
+    @newton_solver.setter
+    def newton_solver(self, value):
+        if value is None:
+            self._newton_solver = value
+        elif isinstance(value, NewtonSolver):
+            if self._newton_solver:
+                warnings.warn(
+                    "Settings for the Newton solver will be overwritten", UserWarning
+                )
```
RemDelaporteMathurin, line 71: I wouldn't make this an actual `UserWarning` but rather a print statement. The reason for this is that users could never get rid of this warning. 

Same for other classes.

```suggestion
                print("Settings for the Newton solver will be overwritten")
```

RemDelaporteMathurin (changes requested):

festim/concentration/traps/extrinsic_trap.py
KulaginVladimir: Corrected

festim/concentration/traps/extrinsic_trap.py
KulaginVladimir: OK

festim/nonlinear_problem.py
```diff
@@ -0,0 +1,31 @@
...
+
+class Problem(f.NonlinearProblem):
+    """
+    Class to set up a nonlinear variational problem (F(u, v)=0) to solve
+    by the Newton method based on the form of the variational problem, the Jacobian
+    form of the variational problem, and the boundary conditions
+
+    Args:
+        jacobian_form (ufl.Form): the Jacobian form of the variational problem
+        residual_form (ufl.Form): the form of the variational problem
```
RemDelaporteMathurin, line 12:
```suggestion
        J (ufl.Form): the Jacobian form of the variational problem
        F (ufl.Form): the form of the variational problem
```

RemDelaporteMathurin (approved):
Thank you so much for this awesome contribution @KulaginVladimir !!

Issue #638, opened 2023-11-13
Additional derived quantities


RemDelaporteMathurin:
- [ ] TotalVolume 
    - [x] Cartesian #703 
    - [ ] Cylindrical
    - [ ] Spherical
- [ ] AverageVolume
    - [ ] Cartesian
    - [ ] Cylindrical
    - [ ] Spherical
- [ ] TotalSurface
    - [ ] Cartesian
    - [ ] Cylindrical
    - [ ] Spherical
- [ ] AverageSurface
    - [ ] Cartesian
    - [ ] Cylindrical
    - [ ] Spherical
- [ ] MinimumVolume
- [ ] MaximumVolume
- [ ] MinimumSurface
- [ ] MaximumSurface

RemDelaporteMathurin:
> I just started working on this issue, is there any specificities to keep in mind while doing it ?

Glad to hear! So I've opened a PR for TotalVolume #703 since I needed it for one of my branches but the rest is still left to do.
Maybe one difference with FESTIM 1 is that we would like derived quantities to be "stand alone".

In FESTIM-1 all the `DerivedQuantity` objects have to be bundled in a `DerivedQuantities` object, and this bigger object has a filename attribute to export to csv.

In FESTIM2, we would like to be able to export each `DerivedQuantity` independently which means they all have a `filename` attribute.

Later down the line, we'll think of how we can also bundle them to avoid having too many files, but I think this new architecture will be easier to deal with and more flexible.

Apart from that, I'm not aware of any specificity. I remember that finding the minimum and maximum of a solution in a subdomain in legacy-fenics was a bit tricky, I don't know how to do it in dolfinx. The fenics discourse group is your friend! 😄

PR #735, opened 2024-04-04
DerivedQuantities for fenicsx

JonathanGSDUFOUR:
## Proposed changes

A first look into what I have for now in trying to work on Issue #638 . 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted


RemDelaporteMathurin:
@JonathanGSDUFOUR glad to see progress on this! Have you tried testing your implementation on an example?

JonathanGSDUFOUR:
> @JonathanGSDUFOUR glad to see progress on this! Have you tried testing your implementation on an example?

Not yet, but it's a good next step indeed

festim/exports/average_surface.py
```diff
@@ -1,32 +1,35 @@
...
+        return f"Average {self.field.name} surface {self.surface.id}"
 
     def compute(self):
-        """Compute the total value of the function on the surface"""
-        # placeholder for the coming function
+        """
+        Computes the average value of the field on the defined surface
+        subdomain, and appends it to the data list
+        """
+        self.value = np.mean(self.field.solution.x.array[self.surface.indices])
```
RemDelaporteMathurin, line 34: No this won't work.

Imagine you have an array of 100 cells with 90 that are at value 10.
The other 10 cells have a value 1.

The average according to your implementation is 910/100=9.10

But you have no information about the localisation of these cells.

Maybe the 90 cells are only located on a very small region and 85% of the domain is at value 1?

The correct formulation for the average of a function f on a surface is:

integral(f * dS)/integral(1 * dS)

festim/exports/surface_flux.py
```diff
@@ -24,9 +22,13 @@ def __init__(
...
     ) -> None:
         super().__init__(field, surface, filename)
+    
+    @property
+    def title(self):
+        return f"{self.field.name} flux surface {self.surface.id}"
 
     def compute(self, n, ds):
-        """Computes the value of the surface flux at the surface
+        """Computes the value of the flux of a at the surface
```
RemDelaporteMathurin, line 31:
```suggestion
        """Computes the value of the flux at the surface
```

RemDelaporteMathurin:
> Would it be better to test each quantity separately or can we test them all in one case ?

We want to test each quantity ensuring it computes the expected value.
A way to do it would be to create a function, store it as the `species.solution` and then check that the value computed by the derived quantity is the expected one.

RemDelaporteMathurin:
@JonathanGSDUFOUR I don't know if you are aware but:
- the tests don't pass with your changes. You can run the tests locally with `python -m pytest .`
- there are a couple of unadressed comments

RemDelaporteMathurin:
> I currently have an issue with the version of dolfinx on my environment

What's the issue?

JonathanGSDUFOUR:
> > I currently have an issue with the version of dolfinx on my environment
> 
> What's the issue?

I had the latest version of dolfinx in my conda env (0.8.0) but it's solved now (I have one with 0.7.3 that runs the test correctly)

RemDelaporteMathurin:
@JonathanGSDUFOUR tests pass yay 🎉

test/test_average_surface.py
```diff
@@ -0,0 +1,49 @@
...
+    facet_indices = np.array(
+        dummy_surface.locate_boundary_facet_indices(my_mesh.mesh, 0),
+        dtype=np.int32,
+    )
+    tags_facets = np.array(
+        [1],
+        dtype=np.int32,
+    )
+    facet_meshtags = meshtags(my_mesh.mesh, 0, facet_indices, tags_facets)
+    ds = ufl.Measure("ds", domain=my_mesh.mesh, subdomain_data=facet_meshtags)
```
RemDelaporteMathurin, line 29: I think we may be able to simplify this using `Mesh1D.define_meshtags()`

test/test_derived_quantities.py
```diff
@@ -0,0 +1,49 @@
...
+        (surf_quant, "H surface 1"),
+        (vol_quant, "H volume 1"),
+    ],
+)
+def test_title(quantity, expected_title, tmp_path):
+    quantity.filename = os.path.join(tmp_path, "test.csv")
+    quantity.value = 1
+    quantity.write(t=1)
+
+    assert quantity.title == expected_title
```
RemDelaporteMathurin, line 49: What is this testing? If it's testing that the quantity has the correct title, then `tmp_path` and `.write()` etc are not needed. 

If it's testing the `.write` method then we should assert things like "Does the file exist?" "Does it contain the expected number of rows?" etc.

RemDelaporteMathurin (changes requested):
This is looking very good!

- Not for this PR but I think it would be good to raise an issue to remind ourselves to write tests for the surface classes in 2D since `MaximumSurface` `MinimumSurface` `AverageSurface` and `TotalSurface` all give the exact same value in 1D.
- To test that `MinimumVolume` and `MaximumVolume` compute the right value in the right subdomain, could we write tests for a multi-material problem? I guess I have the same concerns about other classes

test/test_surface_quantity.py
```diff
@@ -7,7 +7,7 @@
 import os
 
 
-def test_surface_flux_export_compute():
+def test_surface_flux_export_compute(tmp_path):
```
RemDelaporteMathurin, line 10: Don't think we need tmp_path here

test/test_surface_quantity.py
```diff
@@ -43,7 +43,7 @@ def test_surface_flux_export_compute():
     my_export.D = D
 
     # RUN
-    my_export.compute(ds=ds)
+    my_export.compute(n=c, ds=ds)
```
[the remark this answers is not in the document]
JonathanGSDUFOUR: Oh right not the good argument. But since n is re defined in the compute function of `SurfaceFlux` that's why the test still pass. We should go back to `my_export.compute(ds=ds)` with no n in argument of the function right ?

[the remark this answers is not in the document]
RemDelaporteMathurin: Yes you are correct we should remove this argument

festim/exports/surface_quantity.py
```diff
@@ -27,6 +27,10 @@ def __init__(self, field, surface, filename: str = None) -> None:
         self.data = []
         self._first_time_export = True
 
+    @property
+    def title(self):
+        return f"{self.field.name} surface {self.surface.id}"
+
```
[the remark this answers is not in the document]
JonathanGSDUFOUR: The function `title` is called in the `write` function further in the file, but since it's its only occurence it may be better to replace it by something else ?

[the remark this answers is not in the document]
RemDelaporteMathurin: Well the thing is that this title is meaningless really. No one will ever just create a `SurfaceQuantity`, do nothing with it, and then call `compute` (because `compute` doesn't exist!). The only use for this class is to bundle the shared functionality between classes like `MaximumSurface` `SurfaceFlux` etc.

Really the only use for this class is to be inherited by another class that will define the `compute` method and the `title` property.

We should simply remove this property `title` at the `SurfaceQuantity` and `VolumeQuantity` level

festim/hydrogen_transport_problem.py
```diff
@@ -787,23 +787,26 @@ def post_processing(self):
         for export in self.exports:
             # TODO if export type derived quantity
             if isinstance(export, F.SurfaceQuantity):
-                if isinstance(export, F.SurfaceFlux):
+                if (
+                    isinstance(export, F.SurfaceFlux)
+                    or isinstance(export, F.TotalSurface)
+                    or isinstance(export, F.AverageSurface)
+                ):
```
RemDelaporteMathurin, line 794: You can simplify this to:

```suggestion
                if isinstance(export, (F.SurfaceFlux, F.TotalSurface, F.AverageSurface)):
```

festim/hydrogen_transport_problem.py
```diff
@@ -787,23 +787,26 @@ def post_processing(self):
...
                 # update export data
                 export.t.append(float(self.t))
 
                 # if filename given write export data to file
                 if export.filename is not None:
                     export.write(t=float(self.t))
             elif isinstance(export, F.VolumeQuantity):
-                if isinstance(export, F.TotalVolume):
+                if isinstance(export, F.TotalVolume) or isinstance(
+                    export, F.AverageVolume
```
RemDelaporteMathurin, line 808: Same comment as above

RemDelaporteMathurin (changes requested):
hey @JonathanGSDUFOUR this is good to go! Just two small refactoring comments and we can merge 🎉

RemDelaporteMathurin (approved):
This is all good to go @JonathanGSDUFOUR ! Thanks for this effort! 🎉

PR #736, opened 2024-04-04
Derived Quantites refresh: Units in title

jhdark:
## Proposed changes

New PR restoring #711 

With this change, users can opt to have units in the titles of the derived quantities file. They can do this with a new argument within the `DerivedQuantities` class like so:
```python
my_derived_quantities = F.DerivedQuantities(
    [
        F.TotalVolume("solute", volume=1),
        F.AverageVolume("solute", volume=1),
        F.MaximumVolume("solute", volume=1),
        F.MinimumVolume("solute", volume=1),
        F.SurfaceFlux("solute", surface=2),
    ],
    filename="derived_quantities.csv",
    show_units=True,
)
```
The unit is then added to the derived quantity's title, accounting for the mesh's dimension. For example, should a user call for a `TotalVolume` of the `solute` field in a 1D mesh case with units, the title will be `Total solute volume 1 (H m-2) `, whereas in the same case but in 2D, the title will be' Total solute volume 1 (H m-1) `.

The original format of the titles is kept the same to avoid introducing breaking changes; however, it could be debated whether to change them to make them more consistent.

I have also taken the liberty to add doc strings to all derived quantities.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Black formatted

## Further comments

### DIFF
--- a/festim/exports/derived_quantities/average_volume.py
+++ b/festim/exports/derived_quantities/average_volume.py
@@ -3,9 +3,35 @@ import fenics as f
 
 
 class AverageVolume(VolumeQuantity):
+    """
+    Computes the average value of a field in a given volume
+    Args:
+        field (str, int):  the field ("solute", 0, 1, "T", "retention")
+        volume (int): the volume id
+    Attributes:
+        field (str, int):  the field ("solute", 0, 1, "T", "retention")
+        volume (int): the volume id
+        show_units (bool): show the units in the title in the derived quantities
+            file
+        function (dolfin.function.function.Function): the solution function of
+            the field
+    Notes:
+        Units are in H/m3 for hydrogen concentration and K for temperature
+    """
+
     def __init__(self, field, volume: int) -> None:
-        super().__init__(field, volume)
-        self.title = "Average {} volume {}".format(self.field, self.volume)
+        super().__init__(field=field, volume=volume)
+
+    @property
+    def title(self):
+        quantity_title = f"Average {self.field} volume {self.volume}"
+        if self.show_units:
+            if self.field == "T":
+                return quantity_title + " (K)"
+            else:
+                return quantity_title + " (H m-3)"
+        else:
+            return quantity_title
 
     def compute(self):
         return f.assemble(self.function * self.dx(self.volume)) / f.assemble(
--- a/festim/exports/derived_quantities/derived_quantities.py
+++ b/festim/exports/derived_quantities/derived_quantities.py
@@ -23,6 +23,8 @@ class DerivedQuantities(list):
         nb_iterations_between_exports (int, optional): number of
             iterations between each export. If None, the file will be
             exported at the last timestep. Defaults to None.
+        show_units (bool, optional): will show the units of each
+            derived quantity in the title in export
     """
 
     def __init__(
@@ -31,6 +33,7 @@ class DerivedQuantities(list):
         filename: str = None,
         nb_iterations_between_compute: int = 1,
         nb_iterations_between_exports: int = None,
+        show_units=False,
     ) -> None:
         # checks that input is list
         if len(args) == 0:
@@ -43,8 +46,9 @@ class DerivedQuantities(list):
         self.filename = filename
         self.nb_iterations_between_compute = nb_iterations_between_compute
         self.nb_iterations_between_exports = nb_iterations_between_exports
+        self.show_units = show_units
 
-        self.data = [self.make_header()]
+        self.data = []
         self.t = []
 
     @property
@@ -108,6 +112,7 @@ class DerivedQuantities(list):
     def make_header(self):
         header = ["t(s)"]
         for quantity in self:
+            quantity.show_units = self.show_units
             header.append(quantity.title)
         return header
 
@@ -139,6 +144,11 @@ class DerivedQuantities(list):
                 value = quantity.compute(self.volume_markers)
             else:
                 value = quantity.compute()
+
+            # check if first time writing data
+            if len(self.data) == 0:
+                self.data = [self.make_header()]
+
             quantity.data.append(value)
             quantity.t.append(t)
             row.append(value)
--- a/festim/exports/derived_quantities/maximum_volume.py
+++ b/festim/exports/derived_quantities/maximum_volume.py
@@ -4,9 +4,39 @@ import numpy as np
 
 
 class MaximumVolume(VolumeQuantity):
+    """
+    Computes the maximum value in a field in a given volume
+
+    Args:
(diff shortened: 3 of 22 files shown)


RemDelaporteMathurin (changes requested):
I think it could be worth having a `units` property as it would simplify a lot of the logic.

For instance, this:

```python
    @property
    def title(self):
        quantity_title = f"Flux surface {self.surface}: {self.field}"
        if self.show_units:
            # obtain domain dimension
            dim = self.function.function_space().mesh().topology().dim()
            # new title but only with show_units to not introduce breaking change
            quantity_title = f"{self.field} flux surface {self.surface}"
            if self.field == "T":
                quantity_title = f"Heat flux surface {self.surface}"
                if dim == 1:
                    return quantity_title + " (W m-2)"
                if dim == 2:
                    return quantity_title + " (W m-1)"
                if dim == 3:
                    return quantity_title + " (W)"
            else:
                if dim == 1:
                    return quantity_title + " (H m-2 s-1)"
                if dim == 2:
                    return quantity_title + " (H m-1 s-1)"
                if dim == 3:
                    return quantity_title + " (H s-1)"
        else:
            return quantity_title
```

would be simplified to:

```python
    @property
    def title(self):
        quantity_title = f"Flux surface {self.surface}: {self.field}"
        if self.show_units:
            # new title but only with show_units to not introduce breaking change
            quantity_title = f"{self.field} flux surface {self.surface}"
            if self.field == "T":
                quantity_title = f"Heat flux surface {self.surface}"
            else:
                quantity_title = f"{self.field} flux surface {self.surface}"
            return quantity_title + f" ({self.units})"
        else:
            return quantity_title
```

festim/exports/derived_quantities/derived_quantities.py
```diff
@@ -108,6 +112,7 @@ def filename(self, value):
     def make_header(self):
         header = ["t(s)"]
         for quantity in self:
+            quantity.show_units = self.show_units
```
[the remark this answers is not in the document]
jhdark: Should we only throw the error in SurfaceFlux as it is the only one that will end up having a breaking change? Plus, what do you think the warning should say? `"The current derived_quantities title style will be deprecated in a future release"`

festim/exports/derived_quantities/surface_flux.py
```diff
@@ -4,24 +4,62 @@
...
+        surface (int): the surface id
+        title (str): the title of the derived quantity
+        show_units (bool): show the units in the title in the derived quantities
+            file
+        function (dolfin.function.function.Function): the solution function of
+            the field
+
+    Notes:
         Object to compute the flux J of a field u through a surface
         J = integral(-prop * grad(u) . n ds)
```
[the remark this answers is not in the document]
jhdark: rgr

festim/exports/derived_quantities/derived_quantities.py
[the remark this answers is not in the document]
RemDelaporteMathurin: Why would it be the only one with breaking changes? like TotalVolume is also affected no?

I think the warning message should be more explicit and tell users to set `show_units=True`

RemDelaporteMathurin (approved):
lgtm, @SamueleMeschini @KulaginVladimir @gabriele-ferrero feel free to comment if you have suggestions

festim/exports/derived_quantities/average_volume.py
```diff
@@ -3,9 +3,40 @@
 
 
 class AverageVolume(VolumeQuantity):
+    """
+    Computes the average value of a field in a given volume
+    int(f ds) / int (1 * ds)
```
KulaginVladimir, line 8: Volume integration?

festim/exports/derived_quantities/maximum_volume.py
```diff
@@ -4,9 +4,40 @@
 
 
 class MaximumVolume(VolumeQuantity):
+    """
+    Computes the maximum value in a field in a given volume
```
KulaginVladimir, line 8: Computes the maximum value of a field in a given volume

KulaginVladimir: The same for `MinimumVolume`

festim/exports/derived_quantities/average_volume.py
jhdark: Yeah, good catch! I thought I had gone through and fixed all these

festim/exports/derived_quantities/maximum_volume.py
jhdark: Yeah that is arguably better, I'll update it

festim/exports/derived_quantities/surface_flux.py
```diff
@@ -4,24 +4,70 @@
...
+                return "W m-1"
+            if dim == 3:
+                return "W"
+        else:
+            if dim == 1:
+                return "H m-2 s-1"
+            if dim == 2:
+                return "H m-1 s-1"
+            if dim == 3:
+                return "H s-1"
```
RemDelaporteMathurin, line 56:
```suggestion
        if self.field == "T":
            return f"W m{dim-3}".replace(" m0", "")
        else:
            return f"H m{dim-3} s-1".replace(" m0", "")
```

festim/exports/derived_quantities/total_surface.py
```diff
@@ -3,9 +3,59 @@
...
+                return "K m"
+            if dim == 3:
+                return "K m2"
+        else:
+            if dim == 1:
+                return "H m-2"
+            if dim == 2:
+                return "H m-1"
+            if dim == 3:
+                return "H"
```
RemDelaporteMathurin, line 50:
```suggestion
        if self.field == "T":
            return f"K m{dim-1}".replace(" m0", "")
        else:
            return f"H m{dim-3}".replace(" m0", "")
```

festim/exports/derived_quantities/total_volume.py
```diff
@@ -3,9 +3,59 @@
...
+                return "K m2"
+            if dim == 3:
+                return "K m3"
+        else:
+            if dim == 1:
+                return "H m-2"
+            if dim == 2:
+                return "H m-1"
+            if dim == 3:
+                return "H"
```
RemDelaporteMathurin, line 50:
```suggestion
        if self.field == "T":
            return f"K m{dim}".replace("1", "")
        else:
            return f"H m{dim-3}".replace(" m0", "")
```

RemDelaporteMathurin:
@jhdark are we good to merge?

jhdark:
I think so

PR #737, opened 2024-04-04
fenicsx: Flux BC support

jhdark:
## Proposed changes

With this change, users will be able to impose flux boundary conditions:

$-D\nabla c_{\mathrm{m}} \cdot \mathbf{n} = f(x, y, z, t) \ \mathrm{on} \ \partial\Omega$

Users can implement this like:

```python3
        F.FluxBC(subdomain=right, value=10, species=H),
]
```

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments


festim/boundary_conditions/flux_bc.py
```diff
@@ -0,0 +1,149 @@
...
+        if not isinstance(value, (fem.Function, fem.Constant, np.ndarray)):
+            raise TypeError(
+                f"Value must be a dolfinx.fem.Function, dolfinx.fem.Constant, or a np.ndarray not {type(value)}"
+            )
+        self._value_fenics = value
+
+    @property
+    def time_dependent(self):
+        if self.value is None:
+            return False
```
RemDelaporteMathurin, line 66: Isn't this a bit dangerous? Say I check if it's time-dependent before assigning it a value by mistake, I may think it's not time-dpeendent but then it becomes time dependent if the value is say a function of time.

Should we raise an error instead?

festim/boundary_conditions/flux_bc.py
RemDelaporteMathurin, line 1: Should we start off by having `FluxBCBase` and `ParticleFluxBC` and an empty `HeatFluxBC` class?

festim/boundary_conditions/flux_bc.py
```diff
@@ -0,0 +1,149 @@
...
+                )
+                self.value_fenics.interpolate(self.bc_expr)
+
+    def update(self, t):
+        """Updates the source value
+
+        Args:
+            t (float): the time
+        """
+        if callable(self.value):
```
RemDelaporteMathurin, line 144: Is this needed? I would replace by `if self.time_dependent`? what if it isn't? should we raise an error? or just a print?

RemDelaporteMathurin (changes requested):

festim/boundary_conditions/flux_bc.py
jhdark: For sure!

festim/boundary_conditions/flux_bc.py
jhdark: This is how we also handle it in sources, and Dirichlet bc, so maybe we could change this for all of them in another PR?

festim/boundary_conditions/flux_bc.py
jhdark: Yeah this is valid, but would be worth changing the other cases of this property in other classes too

festim/boundary_conditions/flux_bc.py
```diff
@@ -0,0 +1,258 @@
...
+        >>> ParticleFluxBC(subdomain=my_subdomain, value=lambda T: 1 + T, species="H")
+        >>> ParticleFluxBC(subdomain=my_subdomain, value=lambda x, t: 1 + x[0] + t, species="H")
+    """
+
+    def __init__(self, subdomain, value, species):
+        super().__init__(subdomain=subdomain, value=value)
+        self.species = species
+
+        self.value_fenics = None
+        self.bc_expr = None
```
RemDelaporteMathurin, line 113: can be removed

festim/boundary_conditions/flux_bc.py
```diff
@@ -0,0 +1,258 @@
...
+                # store the expression of the boundary condition
+                # to update the value_fenics later
+                self.bc_expr = fem.Expression(
+                    self.value(**kwargs),
+                    function_space.element.interpolation_points(),
+                )
+                self.value_fenics.interpolate(self.bc_expr)
+
+
+class HeatFluxBC(FluxBCBase):
```
RemDelaporteMathurin, line 179: With this implementation, can the HeatFluxBC be temperature-dependent?
I don't think so. Users should be able to do:

```python
my_flux = HeatFluxBc(lambda T: h * (T-T_ext))
```

RemDelaporteMathurin: With this change, I think that a lot of the code will be duplicated from `ParticleFlux` therefore all this functionality can be moved to `FluxBCBase`

festim/heat_transfer_problem.py
```diff
@@ -53,7 +53,9 @@ def boundary_conditions(self):
 
     @boundary_conditions.setter
     def boundary_conditions(self, value):
-        if not all(isinstance(bc, F.FixedTemperatureBC) for bc in value):
+        if not all(
+            isinstance(bc, (F.FixedTemperatureBC, F.HeatFluxBC)) for bc in value
+        ):
             raise TypeError(
                 "boundary_conditions must be a list of festim.FixedTemperatureBC objects"
```
RemDelaporteMathurin, line 60:
```suggestion
                "boundary_conditions must be a list of festim.FixedTemperatureBC or festim.HeatFluxBC objects"
```

festim/heat_transfer_problem.py
RemDelaporteMathurin, line 1: In this, shouldn't we make sure we update the value of a time-dependent HeatFluxBC? Is it already done?

RemDelaporteMathurin (changes requested):

festim/boundary_conditions/flux_bc.py
jhdark: Ah of course yes

festim/boundary_conditions/flux_bc.py
RemDelaporteMathurin: Sounds good with me

festim/boundary_conditions/flux_bc.py
RemDelaporteMathurin: oh wow that's a lot of code that went to FluxBCBase haha

festim/heat_transfer_problem.py
RemDelaporteMathurin: https://github.com/festim-dev/FESTIM/blob/0375d58c0267aa5c56f6d82b5e97aa8af6a9e642/festim/heat_transfer_problem.py#L337-L345

yes it is arleady done here

festim/boundary_conditions/flux_bc.py
RemDelaporteMathurin: I opened an issue #743 to remind us of this

festim/boundary_conditions/flux_bc.py
RemDelaporteMathurin: Opened an issue to track progress on this #744

RemDelaporteMathurin (approved):

PR #747, opened 2024-04-10
Surface Flux fix

jhdark:
## Proposed changes

With this change, the mesh normal needed to evaluate the surface flux can be evaluated from the field rather than passing it in Hydrogen or HeatTransferProblem. The method is different depending if the problem being solved is multispecies or not, as the solution will either be a function or an index.

This should make it easier to process each future surface quantity export as each one will only need the `ds` ufl.measure:
```python
for export in self.exports:
    if isinstance(export, F.SurfaceQuantity):
        export.compute(self.ds)
```

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/exports/surface_flux.py
+++ b/festim/exports/surface_flux.py
@@ -25,13 +25,21 @@ class SurfaceFlux(F.SurfaceQuantity):
     ) -> None:
         super().__init__(field, surface, filename)
 
-    def compute(self, n, ds):
+    def compute(self, ds):
         """Computes the value of the surface flux at the surface
 
         Args:
-            n (ufl.geometry.FacetNormal): normal vector to the surface
             ds (ufl.Measure): surface measure of the model
         """
+
+        # obtain mesh normal from field
+        # if case multispecies, solution is an index, use sub_function_space
+        if isinstance(self.field.solution, ufl.indexed.Indexed):
+            mesh = self.field.sub_function_space.mesh
+        else:
+            mesh = self.field.solution.function_space.mesh
+        n = ufl.FacetNormal(mesh)
+
         self.value = fem.assemble_scalar(
             fem.form(
                 -self.D
--- a/festim/hydrogen_transport_problem.py
+++ b/festim/hydrogen_transport_problem.py
@@ -826,9 +826,7 @@ class HydrogenTransportProblem:
         for export in self.exports:
             # TODO if export type derived quantity
             if isinstance(export, F.SurfaceQuantity):
-                export.compute(
-                    self.ds,
-                )
+                export.compute(self.ds)
                 # update export data
                 export.t.append(float(self.t))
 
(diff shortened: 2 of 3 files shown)


jhdark:
Not sure why the conda CI is failing, I can re-run the tests tomorrow and then it should be grand

RemDelaporteMathurin (commented):
This would work. However, FacetNormal would be computed at each write of each derived quantity.

For large meshes, computing it is expensive. We could test this with a very large 3D mesh, and do something like:

```python
mesh = large_3d_mesh
u = Function()

# time this loop
for i in range(1000):
    n = ufl.FacetNormal(mesh)
    assemble_scalar(fem.form(ufl.dot(ufl.grad(self.field.solution), n) * ds))
```

as opposed to
```python
mesh = large_3d_mesh
u = Function()
n = ufl.FacetNormal(mesh)
# time this loop
for i in range(1000):
    assemble_scalar(fem.form(ufl.dot(ufl.grad(self.field.solution), n) * ds))
```


I think that, since it is unique to the mesh, it ought to stay in mesh and is simply passed to `compute`

RemDelaporteMathurin (approved):
@jhdark did some timing experiments and computing FacetNormal doesn't seem expensive at all even on big meshes

jhdark:
For reference these were the test files I used:
```python
from dolfinx.mesh import create_unit_cube, meshtags
from mpi4py import MPI
import ufl
from dolfinx import fem
import numpy as np
import time

test_mesh = create_unit_cube(MPI.COMM_WORLD, 100, 100, 200)
x = ufl.SpatialCoordinate(test_mesh)
facet_indices = np.array([], dtype=np.int32)
facet_tags = np.array([], dtype=np.int32)
facet_meshtags = meshtags(test_mesh, 2, facet_indices, facet_tags)
ds = ufl.Measure("ds", domain=test_mesh, subdomain_data=facet_meshtags)
V = fem.FunctionSpace(test_mesh, ("CG", 1))
u = fem.Function(V)

# time this loop
start_time = time.time()
for i in range(1000):
    n = ufl.FacetNormal(test_mesh)
    fem.assemble_scalar(fem.form(ufl.dot(ufl.grad(u), n) * ds))

time_with = time.time() - start_time

print("Time with: ", time_with)
```
and:
```python
from dolfinx.mesh import create_unit_cube, meshtags
from mpi4py import MPI
import ufl
from dolfinx import fem
import numpy as np
import time

test_mesh = create_unit_cube(MPI.COMM_WORLD, 100, 100, 200)
x = ufl.SpatialCoordinate(test_mesh)
facet_indices = np.array([], dtype=np.int32)
facet_tags = np.array([], dtype=np.int32)
facet_meshtags = meshtags(test_mesh, 2, facet_indices, facet_tags)
ds = ufl.Measure("ds", domain=test_mesh, subdomain_data=facet_meshtags)
V = fem.FunctionSpace(test_mesh, ("CG", 1))
u = fem.Function(V)
n = ufl.FacetNormal(test_mesh)

# time this loop
start_time_1 = time.time()
for i in range(1000):
    fem.assemble_scalar(fem.form(ufl.dot(ufl.grad(u), n) * ds))

time_without = time.time() - start_time_1

print("Time without: ", time_without)
```

Issue #749, opened 2024-04-10
Robin BC for H transport


RemDelaporteMathurin:
The `ParticleFluxBC` class introduced in #737 doesn't provide support for concentration-dependent flux BCs.

We need to add it before we can tackle #727

PR #750, opened 2024-04-10
Refactor `create_value_fenics` for source and fluxes

RemDelaporteMathurin:
## Proposed changes

Based on a discussion I had with @jhdark we realised that we only needed to interpolate values onto a Function for dirichletBCs (and maybe InitialConditions). For fluxes and sources, since those are added into the formulation we can just put space-time dependent expressions as a ufl Expression, moreover it doesn't need to be updated in this case.

This is a first step towards #749 

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/boundary_conditions/flux_bc.py
+++ b/festim/boundary_conditions/flux_bc.py
@@ -49,7 +49,7 @@ class FluxBCBase:
             value, (fem.Function, fem.Constant, np.ndarray, ufl.core.expr.Expr)
         ):
             raise TypeError(
-                f"Value must be a dolfinx.fem.Function, dolfinx.fem.Constant, or a np.ndarray not {type(value)}"
+                f"Value must be a dolfinx.fem.Function, dolfinx.fem.Constant, np.ndarray or ufl.core.expr.Expr not {type(value)}"
             )
         self._value_fenics = value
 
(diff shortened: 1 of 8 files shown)


festim/boundary_conditions/flux_bc.py
```diff
@@ -45,7 +45,9 @@ def value_fenics(self, value):
         if value is None:
             self._value_fenics = value
             return
-        if not isinstance(value, (fem.Function, fem.Constant, np.ndarray)):
+        if not isinstance(
+            value, (fem.Function, fem.Constant, np.ndarray, ufl.core.expr.Expr)
+        ):
             raise TypeError(
                 f"Value must be a dolfinx.fem.Function, dolfinx.fem.Constant, or a np.ndarray not {type(value)}"
```
jhdark, line 52:
```suggestion
                f"Value must be a dolfinx.fem.Function, dolfinx.fem.Constant, np.ndarray or ufl.core.expr.Expr not {type(value)}"
```

jhdark (approved):
We're going to be looking into how to replace all of these create_value_fenics into a simpler global function, but at least this cleans things up for now and theoretically should give us better errors since we're no longer interpolating onto further functions. It might be worth retesting some of the MMS tests to see if we can lower the L2 error threshold as a results of this.

PR #755, opened 2024-04-17
Simplify custom tagging of meshes

RemDelaporteMathurin:
## Proposed changes

This PR brings some simplification in subdomains:
- moved the meshtags creation from `Mesh1D` to `Mesh` to make it more generic
- moved locate entities methods to base classes
- added default behaviour of tagging all surfaces with 1 and all cells with 1 to `SurfaceSubdomain` and `VolumeSubdomain` respectively.

An example of how this simplifies a lot of things is the refactoring of the system tests!

@jhdark an additional simplification would be to enable users to provide a geometrical locator function when defining a subdomain (instead of creating a custom class). What do you think?

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Black formatted


test/test_surface_quantity.py
```diff
@@ -148,16 +148,3 @@ def test_writer(tmp_path, value):
...
-    """Test that a TypeError is raised when the surface is not a
-    F.SurfaceSubdomain1D"""
-
-    with pytest.raises(
-        TypeError, match="surface should be an int or F.SurfaceSubdomain1D"
-    ):
-        F.SurfaceQuantity(
-            field=F.Species("H"),
-            surface="1",
-        )
```
jhdark, line 163: This test just needs changing rather than deleting to improve the coverage

PR #761, opened 2024-04-19
Map of institutions in documentation

RemDelaporteMathurin:
## Proposed changes

This PR adds an interactive map of institutions using FESTIM in the documentation, alongside with script to reproduce it

## Types of changes

- [ ] Code refactoring
- [ ] New tests


KulaginVladimir:
@RemDelaporteMathurin 
Looks cool! 

Just a couple of questions.

Should it be above the contents? As an [example](https://deepxde.readthedocs.io/en/latest/user/research.html#), the list of institutions/organisations can be on a dedicated page, but not necessary.

What are these blue areas? Polygons which vertices are the institutions/organisations?
image

How did you place the institutions/organisations? For example, MEPhI is rather here :)
image

RemDelaporteMathurin:
> Should it be above the contents? As an [example](https://deepxde.readthedocs.io/en/latest/user/research.html#), the list of institutions/organisations can be on a dedicated page, but not necessary.

I kind of like the idea that it would be on the front page but we can also have a dedicated page. One day we may have also a list of publications associated with each institution? who knows!



> What are these blue areas? Polygons which vertices are the institutions/organisations?

I think that is yes

> How did you place the institutions/organisations? For example, MEPhI is rather here :)

I asked ChatGPT!!! thanks for pointing this out I'll correct it

RemDelaporteMathurin:
> Should it be above the contents? As an [example](https://deepxde.readthedocs.io/en/latest/user/research.html#), the list of institutions/organisations can be on a dedicated page, but not necessary.

I kind of like the idea that it would be on the front page but we can also have a dedicated page. One day we may have also a list of publications associated with each institution? who knows!



> What are these blue areas? Polygons which vertices are the institutions/organisations?

I think that is yes

> How did you place the institutions/organisations? For example, MEPhI is rather here :)

I asked ChatGPT!!! thanks for pointing this out I'll correct it

Issue #765, opened 2024-05-28
[BUG] Cannot compute derived quantity with retention with `show_units=True`


RemDelaporteMathurin:
**Describe the bug**
When trying to compute a derived quantity (say `TotalVolume`) with `field="retention"` the code crashes with the following error:

This bug occurs on `main` and was introduced by #736

```
Traceback (most recent call last):
  File "/home/remidm/FESTIM/MWE2.py", line 40, in <module>
    my_model.run()
  File "/home/remidm/FESTIM/festim/generic_simulation.py", line 360, in run
    self.run_steady()
  File "/home/remidm/FESTIM/festim/generic_simulation.py", line 390, in run_steady
    self.run_post_processing()
  File "/home/remidm/FESTIM/festim/generic_simulation.py", line 442, in run_post_processing
    self.exports.write(self.label_to_function, self.mesh.dx)
  File "/home/remidm/FESTIM/festim/exports/exports.py", line 97, in write
    export.compute(self.t)
  File "/home/remidm/FESTIM/festim/exports/derived_quantities/derived_quantities.py", line 155, in compute
    self.data = [self.make_header()]
                 ^^^^^^^^^^^^^^^^^^
  File "/home/remidm/FESTIM/festim/exports/derived_quantities/derived_quantities.py", line 121, in make_header
    header.append(quantity.title)
                  ^^^^^^^^^^^^^^
  File "/home/remidm/FESTIM/festim/exports/derived_quantities/total_volume.py", line 47, in title
    return quantity_title + f" ({self.export_unit})"
                                 ^^^^^^^^^^^^^^^^
  File "/home/remidm/FESTIM/festim/exports/derived_quantities/total_volume.py", line 36, in export_unit
    dim = self.function.function_space().mesh().topology().dim()
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Sum' object has no attribute 'function_space'
```

**To Reproduce**

Run:

```python
import festim as F

my_model = F.Simulation()

import numpy as np

my_model.mesh = F.MeshFromVertices(np.linspace(0, 1, num=200))

my_model.materials = F.Material(id=1, D_0=1, E_D=0)

trap = F.Trap(
    k_0=1,
    E_k=0,
    p_0=1,
    E_p=0,
    density=1.3e-3,
    materials=my_model.materials[0],
)

my_model.traps = [trap]

my_model.T = F.Temperature(500)

my_model.settings = F.Settings(
[...]

PR #767, opened 2024-05-29
Fix units for "retention" with TotalVolume and TotalSurface

RemDelaporteMathurin:
Fix for #765

### DIFF
--- a/test/unit/test_exports/test_exports.py
+++ b/test/unit/test_exports/test_exports.py
@@ -1,5 +1,6 @@
 import festim
 import pytest
+import numpy as np
 
 
 class TestExportsMethods:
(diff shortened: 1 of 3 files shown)


test/unit/test_exports/test_exports.py
```diff
@@ -139,3 +140,39 @@ def test_instanciate_with_no_elements():
...
+            [
+                festim.TotalVolume("retention", volume=1),
+                festim.TotalSurface("retention", surface=1),
+            ],
+            show_units=True,
+        )
+    ]
+
+    my_model.initialise()
+    my_model.run()
```
jhdark, line 178: This could potentially be simplified to not have a whole system test:
```suggestion
def test_show_units_true_with_retention():
    """Test to catch the bug in issue #765"""
     test_mesh = f.UnitIntervalMesh(10)

    my_exports = [
        F.TotalVolume(volume=3, field="retention"),
        F.TotalSurface(surface=2, field="retention"),
    ]
    for export in my_exports:
        export.dx = f.Measure("dx", domain=test_mesh)
        export.ds = f.Measure("ds", domain=test_mesh)
        export.show_units = True

    expected_titles = [
        "Total retention volume 3 (H m-2)",
        "Total retention surface 2 (H m-2)",
    ]

    for export, expected_title in zip(my_exports, expected_titles):
        assert export.title == expected_title
```

jhdark (approved):
This is a good quick fix. We could also leave a tip and/or the solution in the issue #765 so what we could update it to a good first issue

test/unit/test_exports/test_exports.py
RemDelaporteMathurin: This does not catch the bug described in the original issue. Running this on main produces:

```

    @property
    def export_unit(self):
        # obtain domain dimension
>       dim = self.function.function_space().mesh().topology().dim()
E       AttributeError: 'NoneType' object has no attribute 'function_space'
```

Which is not the bug. Moreover, one could fabricate the retention sum function but then this completely circumvent the integration testing.

RemDelaporteMathurin:
> so what we could update it to a good first issue

We already have quite a few good first issues and they seem to be left untouched haha
We'll probably have to close a few before 1.3 btw

PR #770, opened 2024-05-31
Temperature can be given as float or sp.Expr

### LINKED ISSUE #741: Give temperature as float
Users should be allowed to do:

```python
import festim as F

model = F.Simulation()
model.T = 500
```

equivalent to

```python
import festim as F

model = F.Simulation()
model.T = F.Temperature(500)
```

It would be a small (?) QoL improvement.

All we would have to do is add a setter for the T attribute in Simulation:

```python
    @property
    def T(self):
        return self._T
    
    @T.setter
    def T(self, value):
        if value is None or isinstance(value, festim.Temperature):
            self._T = value
        elif isinstance(value, float):
            self._T = festim.Temperature(value)
        else:
            raise TypeError("Accepted types for T are festim.Temperature or float")
```

RemDelaporteMathurin:
## Proposed changes

This fixes #741 by adding a setter/getter to the `T` attribute of `Simulation`.

Now, users can simply give the temperature as a numerical value or sympy Expression without having to create a `Temperature` object.

Little QoL improvement!

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -164,6 +164,19 @@ class Simulation:
                 "accepted types for exports are list, festim.Export or festim.Exports"
             )
 
+    @property
+    def T(self):
+        return self._T
+
+    @T.setter
+    def T(self, value):
+        if isinstance(value, festim.Temperature):
+            self._T = value
+        elif value is None:
+            self._T = value
+        elif isinstance(value, (int, float)):
+            self._T = festim.Temperature(value)
+
     def attribute_source_terms(self):
         """Assigns the source terms (in self.sources) to the correct field
         (self.mobile, self.T, or traps)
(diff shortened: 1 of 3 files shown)


festim/generic_simulation.py
```diff
@@ -164,6 +165,19 @@ def exports(self, value):
...
+        return self._T
+
+    @T.setter
+    def T(self, value):
+        if isinstance(value, festim.Temperature):
+            self._T = value
+        elif value is None:
+            self._T = value
+        elif isinstance(value, (int, float, sp.Expr)):
+            self._T = festim.Temperature(value)
```
KulaginVladimir, line 179: Possibly, an `else` case with an error can be included to add clarity? Similar to https://github.com/festim-dev/FESTIM/blob/fb2c99f593d4f695573c28acf88137ddad252b38/festim/generic_simulation.py#L124-L127

RemDelaporteMathurin: added

PR #774, opened 2024-06-04
IC improvements for Heat Transfer

RemDelaporteMathurin:
## Proposed changes

Currently when running a transient Heat Transfer problem, an initial condition is required.
The process is a bit cumbersome as if users just want to initialise the temperature to say 300 K then they have to
- create a `InitialCondition` object
- give it the field `"T"` (which is redundant!)
- pass it to `HeatTransferProblem`

This PR simplifies the process.
Now users can simply give a float, int or sympy Expression as the `initial_condition` argument.
If users want to give a initial condition from XDMF then the process is unchanged.

Moreover, this adds an error when transient heat transfer sims are run without giving an IC.

Documentation was also updated as the confusion was first reported [here](https://festim.discourse.group/t/heat-transfer-problem-how-to-set-up-transient-calcumations/39/1).

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/temperature/temperature_solver.py
+++ b/festim/temperature/temperature_solver.py
@@ -86,7 +86,7 @@ class HeatTransferProblem(festim.Temperature):
         self.v_T = f.TestFunction(V)
         if self.transient and self.initial_condition is None:
             raise AttributeError(
-                "Initial condition is required for transient simulation"
+                "Initial condition is required for transient heat transfer simulations"
             )
         if self.transient and self.initial_condition:
             if isinstance(self.initial_condition.value, str):
--- a/festim/temperature/temperature_solver.py
+++ b/festim/temperature/temperature_solver.py
@@ -8,7 +8,7 @@ class HeatTransferProblem(festim.Temperature):
     Args:
         transient (bool, optional): If True, a transient simulation will
             be run. Defaults to True.
-        initial_condition (festim.InitialCondition, optional): The initial condition.
+        initial_condition (int, float, sp.Expr, festim.InitialCondition, optional): The initial condition.
             Only needed if transient is True.
         absolute_tolerance (float, optional): the absolute tolerance of the newton
             solver. Defaults to 1e-03
@@ -55,6 +55,17 @@ class HeatTransferProblem(festim.Temperature):
         self.sources = []
         self.boundary_conditions = []
         self.sub_expressions = []
+
+    @property
+    def initial_condition(self):
+        return self._initial_condition
+
+    @initial_condition.setter
+    def initial_condition(self, value):
+        if isinstance(value, (int, float, sp.Expr)):
+            self._initial_condition = festim.InitialCondition(field="T", value=value)
+        else:
+            self._initial_condition = value
 
     # TODO rename initialise?
     def create_functions(self, materials, mesh, dt=None):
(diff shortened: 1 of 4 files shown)


KulaginVladimir:
@RemDelaporteMathurin 
Looks good to me. I also got a bit tired of using `InitialCondition` with a constant initial temperature.

Just two unnecessary moments

festim/temperature/temperature_solver.py
```diff
@@ -8,7 +8,7 @@ class HeatTransferProblem(festim.Temperature):
     Args:
         transient (bool, optional): If True, a transient simulation will
             be run. Defaults to True.
-        initial_condition (festim.InitialCondition, optional): The initial condition.
+        initial_condition (int, float, sp.Expr, festim.InitialCondition, optional): The initial condition.
```
KulaginVladimir, line 11: Won't it be useful to add a small test to check `create_functions` with `sp.Expr`?

festim/temperature/temperature_solver.py
```diff
@@ -56,6 +56,17 @@ def __init__(
...
+    @property
+    def initial_condition(self):
+        return self._initial_condition
+
+    @initial_condition.setter
+    def initial_condition(self, value):
+        if isinstance(value, (int, float, sp.Expr)):
+            self._initial_condition = festim.InitialCondition(field="T", value=value)
+        else:
+            self._initial_condition = value
```
KulaginVladimir, line 68: If I'm correct, users can provide `F.InitialCondition` with any `field` if they set directly `initial_condition=F.InitialCondition(...)`. Is there any point in restricting the allowable `field` for the initial condition in the heat transfer problem?

festim/temperature/temperature_solver.py
RemDelaporteMathurin: Can do

festim/temperature/temperature_solver.py
RemDelaporteMathurin: This is correct. What do you mean by "restricting the allowable field"?

KulaginVladimir: I mean, we can add a check for the `field` value. If it's not `T` than raise an error. Though, such restriction doesn't make a lot of sense

RemDelaporteMathurin: Since in this specific case `field` isn't even required (users could put anything in there) I don't think that's a good idea.
What we could do is make different classes `InitialTemperature` and `InitialConcentration` (a bit like the approach we hvae in `fenicsx`). For `InitialTemperature` we wouldn't need to specify field.

Maybe worth doing it in another PR? Maybe it's not even needed at all

KulaginVladimir: `InitialTemperature` and `InitialConcentration` sound interesting. However, I assume that the current users should be familiar with `InitialCondition` as it is, if they use it. Possibly, these new classes can be considered later, yes

festim/temperature/temperature_solver.py
RemDelaporteMathurin: Done! Let me know what you think

KulaginVladimir: Perfect

Issue #739, opened 2024-04-05
Guide to meshing


RemDelaporteMathurin:
FESTIM provides very minimal meshing capabilities limited to 1D meshes.

For 2D/3D meshes we usually rely on Salome and `meshio` to convert the mesh to XDMF.

It would be nice to have a guide in the documentation explaining the workflow to produce a FESTIM-ready mesh step-by-step with Salome, Cubit, and Gmsh.

PR #776, opened 2024-06-05
SALOME meshing guide

RemDelaporteMathurin:
## Proposed changes

This is an attempt at a simple SALOME to FESTIM tutorial advancing towards #739.

## Types of changes

- [ ] Code refactoring
- [ ] New tests


KulaginVladimir:
@RemDelaporteMathurin, I followed the guide. In my opinion, it's nice! However, it's a bit hard to read, since the bullet points are just stacked to screenshots, but not really a problem. 

Possibly, a 3D example could be a better choice, as it's usually easier to do a simpler case based on a more complex example, rather than otherwise. What do you think?

Moreover, what does the last screenshot show? I expected a mesh representation in ParaView (similar to the mesh in SALOME from the pre-last screenshot of your guide). By default, I get this view in ParaView:
image

RemDelaporteMathurin:
> it's a bit hard to read, since the bullet points are just stacked to screenshots, but not really a problem.

I adapted the size of the images let me know what you think



> Moreover, what does the last screenshot show? I expected a mesh representation in ParaView (similar to the mesh in SALOME from the pre-last screenshot of your guide). By default, I get this view in ParaView:

So we discussed about this on the side and it appeared that this is just from using the wrong XDMF reader in Paraview.
Maybe similar to this PR we need a small Paraview tutorial or at least write somewhere which reader users should use.

RemDelaporteMathurin:
@KulaginVladimir @jhdark  I added the CAD based workflow, let me know what you think

PR #790, opened 2024-06-27
Surface Kinetics

KulaginVladimir:
## Proposed changes

This PR represents the initial attempt on implementing Surface Kinetics feature in FESTIM and is aimed at starting the discussion. 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Black formatted


festim/generic_simulation.py
```diff
@@ -21,6 +21,9 @@ class Simulation:
             concentration). Defaults to None.
         traps (festim.Traps or list or festim.Trap, optional): The model's traps. Defaults
             to None.
+        surface_concentrations (festim.SurfaceConcentrations or 
+            list or festim.SurfaceConcentration, optional): The model's surface species. 
+            Defaults to None.
```
KulaginVladimir, line 26: From discussions, it seems that an introduction of a `SurfaceKinetics` boundary condition would be a better option rather than a seperate attribute of `Simulation`. With that, I got a question concerning the object that will have `solution`, `previous_solution` attributes. What will be better: to keep newly introduced class `SurfaceConcentration` (which is based on the `Concentration` class and, thus, has `solution`, `previous_solution`, etc attributes) or to put all of the parameters in a new BC class? Or maybe another option?

PR #797, opened 2024-07-04
Docs fixes

RemDelaporteMathurin:
## Proposed changes

We noticed some bugs in the API section of the docs in #791 

Also take the opportunity to remove the numbering from the user guide and other docs improvements.

## Types of changes

- [ ] Code refactoring
- [ ] New tests


KulaginVladimir:
Maybe, you could also add indents to `Note` (I suppose, should be `Notes`) [here](https://github.com/festim-dev/FESTIM/blob/b56a3dc1dca432109d489c012aefea280698b807/festim/exports/derived_quantities/surface_flux.py#L99) and [here](https://github.com/festim-dev/FESTIM/blob/b56a3dc1dca432109d489c012aefea280698b807/festim/exports/derived_quantities/surface_flux.py#L175)?

Issue #798, opened 2024-07-04
Theory background for `SurfaceKinetics`


RemDelaporteMathurin:
Now that #791 is merged, we need to write a theory section in the docs

PR #805, opened 2024-07-09
`SurfaceKinetics` docs and form

KulaginVladimir:
## Proposed changes
From discussions with @RemDelaporteMathurin, I suppose that there are mistakes in the variational form for `SurfaceKinetics` BC and API docs. 

The current docs states that the BC for solute species is:
$-D \nabla c_\mathrm{m} \cdot \mathbf{n} = -\lambda_{\mathrm{IS}} \dfrac{\partial c_{\mathrm{m}}}{\partial t}- J_{\mathrm{bs}} + J_{\mathrm{sb}}.$
If $\mathbf{n}$ is the outward normal vector of the boundary, at steady-state 1D, this should yield: $D\dfrac{\partial c_\mathrm{m}}{\partial x}= -J_{\mathrm{bs}}+J_{\mathrm{sb}}$, what contradicts [Pick & Sonnenberg](https://www.sciencedirect.com/science/article/abs/pii/0022311585904593). I assume, the correct form is:
$-D \nabla c_\mathrm{m} \cdot \mathbf{n} = \lambda_{\mathrm{IS}} \dfrac{\partial c_{\mathrm{m}}}{\partial t}+J_{\mathrm{bs}} - J_{\mathrm{sb}}.$

To check the implementation, here is a simple [MMS test](https://github.com/KulaginVladimir/FESTIM-SurfaceKinetics-Validation/blob/main/MMS/MMS.ipynb), based on the approach of @RemDelaporteMathurin 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/boundary_conditions/fluxes/surface_kinetics.py
+++ b/festim/boundary_conditions/fluxes/surface_kinetics.py
@@ -13,8 +13,8 @@ class SurfaceKinetics(FluxBC):
 
     .. math::
 
-        -D \nabla c_\mathrm{m} \cdot \mathbf{n} = -\lambda_{\mathrm{IS}} \dfrac{\partial c_{\mathrm{m}}}{\partial t}
-        - J_{\mathrm{bs}} + J_{\mathrm{sb}},
+        -D \nabla c_\mathrm{m} \cdot \mathbf{n} = \lambda_{\mathrm{IS}} \dfrac{\partial c_{\mathrm{m}}}{\partial t}
+        + J_{\mathrm{bs}} - J_{\mathrm{sb}},
 
     where :math:`c_{\mathrm{m}}` is the concentration of mobile hydrogen (:math:`\mathrm{H \ m}^{-3}`),
     :math:`c_{\mathrm{s}}` is the surface concentration of adsorbed hydrogen (:math:`\mathrm{H \ m}^{-2}`),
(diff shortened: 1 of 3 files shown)


RemDelaporteMathurin:
Hey I'm just looking at the signs of terms when we have a negative gradient at x=0.

Then we should have $J_{\mathrm{sb}}>J_{\mathrm{bs}} $. So in the following expression:

 $-D \nabla c_\mathrm{m} \cdot \mathbf{n} = J_{\mathrm{bs}} - J_{\mathrm{sb}}.$

The first term is negative, and the RHS is also negative. Sounds good to me!

RemDelaporteMathurin (approved):
Thanks for this fix.

When we take care of #798 maybe it would be beneficial to add a short derivation showing that, at x=0, steady state, then we have $D \frac{dc}{dx} = J_{\mathrm{bs}} - J_{\mathrm{sb}}$ which is what Pick & Sonnenberg have in Equation X 

Also, if we are happy with the MMS test, it might be good to add it to the CI here?

KulaginVladimir:
We can discuss the MMS. Personally, I don't like that it's defined on $t\in[0,5]$.

KulaginVladimir:
@RemDelaporteMathurin, I added here (also in this [repo](https://github.com/KulaginVladimir/FESTIM-SurfaceKinetics-Validation/blob/main/MMS/MMS.ipynb)) a new MMS. Let me know what you think of it.

RemDelaporteMathurin (approved):
Looks good to me! Nice to have a MMS case for this!

Feel free to merge

Issue #787, opened 2024-06-18
Warn when `Temperature` is mixed with BCs, sources, etc.


RemDelaporteMathurin:
Currently users can setup temperature boundary conditions, sources, etc while setting up a `F.Temperature()` object.

When this happens, the BCs and sources are [ignored](https://festim.discourse.group/t/set-different-temperature-at-different-boundaries/52). 

It would be nice to either raise an error or at least a warning that `F.Temperature()` and `F.DirichletBC(field="T", ....)` should not be used together.

JonathanGSDUFOUR:
I'll start working on this issue, should not take too long if I understand well the code

RemDelaporteMathurin:
> I'll start working on this issue, should not take too long if I understand well the code

Sounds good! I assigned you to it.

RemDelaporteMathurin:
@JonathanGSDUFOUR I think there has been a confusion in what exactly we need to fix.

Here's an example:

```python
import festim as F
import numpy as np

my_model = F.Simulation()


# Here, we define boundary conditions for heat transfer
# and at the same time we prescribe a temperature field which overwrites everything!

my_model.T = F.Temperature(873)

my_model.boundary_conditions = [
    F.DirichletBC(field="T", value=1200, surfaces=1),
    F.DirichletBC(field="T", value=373, surfaces=2),
]

my_model.materials = F.Material(1, 1, 0)

my_model.mesh = F.MeshFromVertices(np.linspace(0, 1))

my_model.settings = F.Settings(
    absolute_tolerance=1e-10, relative_tolerance=1e-10, transient=False
)

my_model.initialise()
my_model.run()
```

In this model, we set the temperature to 873 K but also set boundary conditions. This should not be possible as it is not compatible.

However in the current state of FESTIM, the code above runs without any issue.

It should return an error saying: `can't use boundary conditions with Temperature(), use HeatTransferProblem() instead.`

Does that make sense?

JonathanGSDUFOUR:
Okay I understand where is the issue !
So I should reset what I did also right ? If it wasn't a real issue

RemDelaporteMathurin:
I think you can close the PR you opened and open a new PR. 

I advise you to start by writing a test. Basically the test would be running the code I sent above and checking that an error is raised.

Then once you check that this test catches the bug you can implement a fix

PR #806, opened 2024-07-10
added test in init and corresponding tests

JonathanGSDUFOUR:
## Proposed changes

Added a tests in the `__init__` part of the `DirichletBC` and the `Source` to raise an error when a `Festim.Temperature` is used  (Issue #787).

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

## Further comments

I went with ValueError to warn the user because it was already used well in other parts of the code, but we could think of warning or other types of errors maybe.


RemDelaporteMathurin:
@JonathanGSDUFOUR I think there was a confusion in the problem we need to solve. I'll add details to the issue to better explain the problem

Issue #800, opened 2024-07-05
Steady-state version of `SurfaceKinetics`


RemDelaporteMathurin:
Following the implementation of the Pick & Sonnenberg surface model by @KulaginVladimir in #791 it would be nice to have a steady state version of it.

The equilibration time for this model is typically short. Therefore it would be interesting to derive the equivalent model at steady state and work out the resulting Dirichlet BC value. 

This would allow users to have a simpler (more performant?) Pick & Sonnenberg model for transient sims.

RemDelaporteMathurin:
@kaelyndunnell this should be of interest to you given your work at ITER

PR #808, opened 2024-07-11
Theory and UG for the `SurfaceKinetics` BC

KulaginVladimir:
## Proposed changes

Following #798, this PR adds documentation on the `SurfaceKinetics` BC to Theory and UG sections of docs.

To-do:
- [x] Theory section
- [x] UG section

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Black formatted


KulaginVladimir:
@RemDelaporteMathurin, I think this can be used as a starting point.

There are two points that I'd like to discuss:
1. Theory section: what do you think of it? Shall we extend it?
2. Shall we note in UG that `AdsorbedHydrogen` was designed specifically for `SurfaceKinetics`?

RemDelaporteMathurin:
I think we also need @ehodille 's review on this one

docs/source/theory.rst
```diff
@@ -272,9 +272,76 @@ Finally, convective heat fluxes can be applied to boundaries:
...
 where :math:`h` is the heat transfer coefficient and :math:`T_{\mathrm{ext}}` is the external temperature.
 
----------------
+Kinetic surface model
+^^^^^^^^^^^^^^^^^^^^^
+
+Modelling hydrogen retention or outgassing might require considering the kinetics of surface processes. 
+A representative example is the hydrogen uptake from a gas phase, when the energy of incident atoms/molecules is not high enough to 
+overcome the surface barrier for implantation. The general approach to account for surface kinetics :cite:`Pick1985, Hodille2017, Guterl2019, Schmid2021` consists in 
+introducing hydrogen surface species :math:`c_\mathrm{s}`.
```
RemDelaporteMathurin, line 281: We should specify the dimension of $c_\mathrm{s}$ since it differs from the bulk concentration

docs/source/theory.rst
```diff
@@ -272,9 +272,76 @@ Finally, convective heat fluxes can be applied to boundaries:
...
+^^^^^^^^^^^^^^^^^^^^^
+
+Modelling hydrogen retention or outgassing might require considering the kinetics of surface processes. 
+A representative example is the hydrogen uptake from a gas phase, when the energy of incident atoms/molecules is not high enough to 
+overcome the surface barrier for implantation. The general approach to account for surface kinetics :cite:`Pick1985, Hodille2017, Guterl2019, Schmid2021` consists in 
+introducing hydrogen surface species :math:`c_\mathrm{s}`. 
+
+Evolution of hydrogen surface concentration is determined by the atomic flux balance at the surface, as sketched in the simplified energy diagram below.
+
+.. figure:: images/potential_diagram.png
```
RemDelaporteMathurin, line 285: Don't you want to use thumbnail like in https://github.com/festim-dev/FESTIM/blob/a2fd832f83b952c812b2e8abc2c63d8297ce9261/docs/source/userguide/mesh.rst?plain=1#L89-L91

docs/source/theory.rst
```diff
@@ -272,9 +272,76 @@ Finally, convective heat fluxes can be applied to boundaries:
...
+    :align: center
+    :width: 800
+    :alt: Idealised potential energy diagram for hydrogen near a surface of an endothermic metal. Energy levels are measured from the :math:`\mathrm{H}_2` state
+
+    Idealised potential energy diagram for hydrogen near a surface of an endothermic metal. Energy levels are measured from the :math:`\mathrm{H}_2` state
```
RemDelaporteMathurin, line 290: This looks duplicated, do we need both?

docs/source/theory.rst
```diff
@@ -272,9 +272,76 @@ Finally, convective heat fluxes can be applied to boundaries:
...
+
+.. math::
+    :label: eq_surf_conc
+    
+    \dfrac{d c_\mathrm{s}}{d t} = J_\mathrm{bs} - J_\mathrm{sb} + J_\mathrm{vs}
+
+where :math:`J_\mathrm{bs}` is the flux of hydrogen atoms from the subsurface (bulk region just beneath the surface) onto the surface, 
+:math:`J_\mathrm{sb}` is the flux of hydrogen atoms from the surface into the subsurface, and :math:`J_\mathrm{vs}` is the net flux of hydrogen 
+atoms from the vacuum onto the surface. The current model does not account for possible surface diffusion and, therefore, is limited to
+one-dimensional hydrogen transport simulations.
```
RemDelaporteMathurin, line 302: Well in theory, you could use this model in 2D or 3D if the diffusion is zero. For example if you have an homogenous J_vs on the surface ?

At the moment it's a technical limitation. The steady state version of it #800 could be applied to multi dimension problems

docs/source/theory.rst
```diff
@@ -272,9 +272,76 @@ Finally, convective heat fluxes can be applied to boundaries:
...
+
+where :math:`n_\mathrm{surf}` is the surface concentration of adsorption sites, :math:`n_\mathrm{IS}` is the bulk concentration of interstitial sites,
+:math:`\lambda_\mathrm{abs}=n_\mathrm{surf}/n_\mathrm{IS}` is the characteristic distance between surface and subsurface sites, :math:`k_\mathrm{bs}` 
+and :math:`k_\mathrm{sb}` are the rate constants for subsurface-to-surface and surface-to-subsurface transitions, respectively. 
+Usually, these rate constants are expressed in the Arrhenius form: :math:`k_i=k_{i,0}\exp(-E_i / kT)`. Both these processes are assumed to take place
+if there are available sites on the surface (in the subsurface). Possible surface/subsurface saturation is accounted for with terms in brackets.
+
+.. note::
+
+    In eq. :eq:`eq_Jsb`, the last term in brackets is usually omitted, since :math:`c_\mathrm{m} \ll n_\mathrm{IS}` is assumed.
```
RemDelaporteMathurin, line 338: You could site papers where it's omitted

docs/source/userguide/boundary_conditions.rst
```diff
@@ -105,6 +105,31 @@ Where :math:`Kd` is the dissociation coefficient, :math:`P` is the partial press
 
     my_bc = DissociationFlux(surfaces=2, Kd_0=2, E_Kd=0.1, P=1e05)
 
+Kinetic surface model (1D)
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Kinetic surface model can be included to account for the evolution of adsorbed hydrogen on a surface with the :class:`festim.SurfaceKinetics` class.
+Refer to the :ref:`theory` section for more details.
```
RemDelaporteMathurin, line 112: You can refer directly to the right subsection.

```suggestion
Refer to the :ref:`Kinetic surface model` section for more details.
```

docs/source/userguide/boundary_conditions.rst
```diff
@@ -105,6 +105,31 @@ Where :math:`Kd` is the dissociation coefficient, :math:`P` is the partial press
...
+        return 1e13*f.exp(-1.0/k_b/T)
+
+    def J_vs(T, surf_conc, t):
+
+        J_des = 2e5*surf_conc**2*f.exp(-1.2/k_b/T)
+        J_ads = 1e17*(1-surf_conc/1e17)**2*f.conditional(t<10, 1, 0)
+
+        return J_ads - J_des
+
+    my_bc = SurfaceKinetics(k_bs=k_bs, k_sb=k_sb, lambda_IS=1.1e-10, n_surf=1e17, n_IS=6.3e28, J_vs=J_vs, surfaces=3, initial_condition=0, t=t)
```
RemDelaporteMathurin, line 132:
```suggestion
    my_bc = SurfaceKinetics(
        k_bs=k_bs,
        k_sb=k_sb,
        lambda_IS=1.1e-10,
        n_surf=1e17,
        n_IS=6.3e28,
        J_vs=J_vs,
        surfaces=3,
        initial_condition=0,
        t=t
        )
```

RemDelaporteMathurin (changes requested):
This is a very good starting point. I would specify that the AdsorbedConcentration is only for SurfaceKinetics yes.

docs/source/theory.rst
KulaginVladimir: In theory, yes. However, I think it can't be used if there is any inhomogeneity (e.g., temperature distribution over the surface, solute source distribution). 

Overall, I wanted to highlight that currently it's 1D+0D. Maybe, we can keep only the part about: The current model does not account for possible surface diffusion?

docs/source/theory.rst
KulaginVladimir: Nope

docs/source/theory.rst
KulaginVladimir: Meanwhile, I omitted a bunch of papers and kept the minimally needed (in my opinion). Is it OK? There are far more than those I included.

docs/source/theory.rst
KulaginVladimir: The units for parameters are not listed at all. Possibly, it'd beneficial to add units to all quantities. What do you think?

docs/source/theory.rst
RemDelaporteMathurin: You are probably right. That's of course if surface diffusion is faster than the other processes.

We want to highlithg that it's only available for 1D problems in FESTIM. We also want to keep that the current model doesn't account for potential surface diffusion 👍

docs/source/theory.rst
RemDelaporteMathurin: Yes of course this is not a paper :-)

docs/source/theory.rst
RemDelaporteMathurin: Yes I think readers would look for units here.

docs/source/theory.rst
KulaginVladimir: > We want to highlithg that it's only available for 1D problems in FESTIM.

Shall we add a `warning`? Though we have warnings in API.

docs/source/theory.rst
```diff
@@ -272,9 +276,75 @@ Finally, convective heat fluxes can be applied to boundaries:
...
+.. math::
+    :label: eq_subsurf_conc
+    
+    -D \nabla c_\mathrm{m} \cdot \mathbf{n} = \lambda_{\mathrm{IS}} \dfrac{\partial c_{\mathrm{m}}}{\partial t} + J_{\mathrm{bs}} - J_{\mathrm{sb}}
+
+where :math:`\lambda_\mathrm{IS}\,[\mathrm{m}]` is the distance between two interstitial sites in the bulk. 
+
+.. note::
+
+    At steady state and :math:`x=0`, :eq:`eq_subsurf_conc` is reduced to :math:`D\frac{\partial c_\mathrm{m}}{\partial x}=J_\mathrm{bs}-J_\mathrm{sb}`
```
RemDelaporteMathurin, line 317: Actually that's not completely correct. Consider the domain $\Omega = [-1, 0]$ then at $x=0$ the outward normal is $+\vec{x}$.

I know that most people assume positive x values only but is it worth adding something?

docs/source/theory.rst
```diff
@@ -272,9 +276,75 @@ Finally, convective heat fluxes can be applied to boundaries:
...
+
+.. math::
+    :label: eq_Jsb
+
+    J_\mathrm{sb} = k_\mathrm{sb} c_\mathrm{s} \left(1-\dfrac{c_\mathrm{m}}{n_\mathrm{IS}}\right)
+
+where :math:`n_\mathrm{surf}\,[\mathrm{m}^{-2}]` is the surface concentration of adsorption sites, :math:`n_\mathrm{IS}\,[\mathrm{m}^{-3}]` is the bulk concentration of interstitial sites,
+:math:`\lambda_\mathrm{abs}=n_\mathrm{surf}/n_\mathrm{IS}\,[\mathrm{m}]` is the characteristic distance between surface and subsurface sites, :math:`k_\mathrm{bs}\,[\mathrm{s}^{-1}]` 
+and :math:`k_\mathrm{sb}\,[\mathrm{s}^{-1}]` are the rate constants for subsurface-to-surface and surface-to-subsurface transitions, respectively. 
+Usually, these rate constants are expressed in the Arrhenius form: :math:`k_i=k_{i,0}\exp(-E_i / kT)`. Both these processes are assumed to take place
```
RemDelaporteMathurin, line 335:
```suggestion
Usually, these rate constants are expressed in the Arrhenius form: :math:`k_i=k_{i,0}\exp(-E_{k,i} / kT)`. Both these processes are assumed to take place
```

docs/source/theory.rst
KulaginVladimir: What do you think of: `At steady-state, eq.(28) for the left domain boundary is reduced to ...`

RemDelaporteMathurin: Yes that sounds good. Maybe to be sure we can add "at the left boundary, the normal vector $\mathbf{n}$ is $-\vec{x}$ therefore ..." or something like that?

KulaginVladimir: So, maybe this will fulfill all the gaps?

At the left boundary, the normal vector $\textbf{n}$ is $-\vec{x}$. The steady-state approximation of eq. (8) at the left boundary is, therefore, $D\frac{\partial c_\mathrm{m}}{\partial x}=J_{\mathrm{bs}}-J_{\mathrm{sb}}$

RemDelaporteMathurin: perfect!

RemDelaporteMathurin (approved):
This looks good to me! 
I'll let @ehodille add his review.

docs/source/theory.rst
```diff
@@ -272,9 +276,75 @@ Finally, convective heat fluxes can be applied to boundaries:
 
 where :math:`h` is the heat transfer coefficient and :math:`T_{\mathrm{ext}}` is the external temperature.
 
----------------
+Kinetic surface model
+^^^^^^^^^^^^^^^^^^^^^
+
+Modelling hydrogen retention or outgassing might require considering the kinetics of surface processes. 
+A representative example is the hydrogen uptake from a gas phase, when the energy of incident atoms/molecules is not high enough to 
+overcome the surface barrier for implantation. The general approach to account for surface kinetics :cite:`Pick1985, Hodille2017, Guterl2019, Schmid2021` consists in
```
ehodille, line 284: Vladimir, do you not want to also mention your papers ?  I think you implemented and used this kind of model in you laser desorption papers (but it is up to you).

KulaginVladimir: @ehodille, well, I used, yes. However, I intended to mention here the pioneers, whereas my papers on LID were mainly based on your works.

docs/source/theory.rst
```diff
@@ -272,9 +276,75 @@ Finally, convective heat fluxes can be applied to boundaries:
...
+
+The governing equation for surface species is:
+
+.. math::
+    :label: eq_surf_conc
+    
+    \dfrac{d c_\mathrm{s}}{d t} = J_\mathrm{bs} - J_\mathrm{sb} + J_\mathrm{vs}~\text{on}~\delta\Omega
+
+where :math:`J_\mathrm{bs}\,[\mathrm{m}^{-2}\,\mathrm{s}^{-1}]` is the flux of hydrogen atoms from the subsurface (bulk region just beneath the surface) onto the surface, 
+:math:`J_\mathrm{sb}\,[\mathrm{m}^{-2}\,\mathrm{s}^{-1}]` is the flux of hydrogen atoms from the surface into the subsurface, and :math:`J_\mathrm{vs}\,[\mathrm{m}^{-2}\,\mathrm{s}^{-1}]` is the net flux of hydrogen
```
ehodille, line 303: does J_vs contains the desorbing flux and all abstraction fluxes (ELy Rideal recombination etc) ?

KulaginVladimir: In general, yes, but it should be defined by a user. Here is an example from [one](https://github.com/KulaginVladimir/FESTIM-SurfaceKinetics-Validation/blob/main/D_damagedW/D_damagedW.ipynb) validation case on how it can be defined:
```python
def J_vs(T, surf_conc, t):
    phi_atom = SP * Gamma_atom * (1 - surf_conc / n_surf) # absorption
    phi_exc = Gamma_atom * sigma_exc * surf_conc # abstraction
    phi_des = 2 * nu0 * (lambda_des * surf_conc) ** 2 * f.exp(-2 * E_des / F.k_B / T) # desorption
    return phi_atom - phi_exc - phi_des
```

RemDelaporteMathurin: Maybe this could be explicited then? something like

```
is the net flux of hydrogen atoms from the vacuum onto the surface (eg recombination, dissociation, Eley-Rideal recombination, etc.).
```

KulaginVladimir: In this case, should we somehow clarify that the flux comes **onto the surface**: `J_vs = +Dissoc. - Recomb. - Eley-Rideal recomb ...`?

RemDelaporteMathurin: Probably

KulaginVladimir: What about this:

Assuming that there are $N$ processes leading to adsorption of hydrogen with flux $J_{\mathrm{ads},i}$ (e.g. dissociation from the gase phase, adsorption of low-energy ions, etc) and $M$ processes leading to desorption of hydrogen with flux $J_{\mathrm{des},j}$ (e.g. molecular recombindation, Eley-Rideal recombination, etc), the resulting flux onto the surface is: $J_\mathrm{vs}=\sum\limits_i^N J_{\mathrm{ads},i}-\sum\limits_j^M J_{\mathrm{des},j}$

RemDelaporteMathurin: Maybe just say J_vs = J_in - J_out where J_in is the sum of all fluxes coming from the vacuum to the surface and J_out from the surface to the vacuum

Or something like that and then give a few examples

KulaginVladimir: A few examples on different processes that can be included?

RemDelaporteMathurin: Yes just like you had in your previous comment

KulaginVladimir: I updated this part. Let me know what you think.

RemDelaporteMathurin: Looks good to me!

RemDelaporteMathurin:
@KulaginVladimir I think we can merge this unless @ehodille has something else to add

Issue #804, opened 2024-07-08
[BUG] trap id is None if append is used


RemDelaporteMathurin:
**Describe the bug**
When a trap is appended to the list of traps then its id is None and we can't perform derived quantity calculations on it.

**To Reproduce**

MWE

```python
import festim as F
import numpy as np


model = F.Simulation()
model.mesh = F.MeshFromVertices(np.linspace(0, 1, num=100))

model.materials = F.Material(id=1, D_0=1, E_D=0, borders=[0, 1])

model.traps = [F.Trap(k_0=1, E_k=0, E_p=0, p_0=1, density=1, materials=model.materials)]

model.traps.append(
    F.Trap(k_0=1, E_k=0, E_p=0, p_0=1, density=1, materials=model.materials)
)


model.T = F.Temperature(500)

model.settings = F.Settings(
    absolute_tolerance=1e-10,
    relative_tolerance=1e-10,
    transient=False,
)

model.initialise()

for trap in model.traps:
    print(trap.id)
    assert trap.id is not None
```

KulaginVladimir:
This bug concerns each overwritten method of list.

RemDelaporteMathurin:
Yes, which means we need to assign ids elsewhere, or, find a way to trigger this every time we change the elements

RemDelaporteMathurin:
After looking at the code a little bit it would seem that the best way to go (and most straightforward) is to move these lines somewhere in `initialise` https://github.com/festim-dev/FESTIM/blob/1add5fc412427c02ed7714d3afc656e81a1e1745/festim/concentration/traps/traps.py#L24-L27

KulaginVladimir:
Yes, otherwise we'd have to modify existing `Traps` methods and other (e.g. `pop`). Though, moving id assigment from `__init__` would require changing tests.

KulaginVladimir:
Does it make sense to add an additional check `if trap.id is None` in this method?
https://github.com/festim-dev/FESTIM/blob/1add5fc412427c02ed7714d3afc656e81a1e1745/festim/concentration/traps/traps.py#L70-L72

so it'd be:
```python
def make_traps_materials(self, materials):
    for i, trap in enumerate(self, 1):
        if trap.id is None:
            trap.id = i
        trap.make_materials(materials)
``` 

Therefore, we  would not have to re-write the tests and the ids would be added after `Simulation.initialise()`, since `make_traps_materials` are called during `h_transport_problem.initialise`

[...]

PR #810, opened 2024-07-11
Fix bug with trap id is None

KulaginVladimir:
## Proposed changes

This is the fix for #804. 

Also, several tests were updated, since the missing ids are now assigned during `Simulation.initialise()`.

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/concentration/traps/traps.py
+++ b/festim/concentration/traps/traps.py
@@ -21,11 +21,6 @@ class Traps(list):
         self.extrinsic_formulations = []
         self.sub_expressions = []
 
-        # add ids if unspecified
-        for i, trap in enumerate(self, 1):
-            if trap.id is None:
-                trap.id = i
-
     @property
     def traps(self):
         warnings.warn(
@@ -70,6 +65,11 @@ class Traps(list):
     def make_traps_materials(self, materials):
         for trap in self:
             trap.make_materials(materials)
+
+    def assign_traps_ids(self):
+        for i, trap in enumerate(self, 1):
+            if trap.id is None:
+                trap.id = i
 
     def create_forms(self, mobile, materials, T, dx, dt=None):
         self.F = 0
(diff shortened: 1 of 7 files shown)


RemDelaporteMathurin (approved):
Thanks for this!

It's a bit annoying that we have to add `id=` to all these tests... We'll probably have to sit down and discuss a proper implementation for this.

The only reason why we need id for traps are for exports, boundary conditions, sources. Right? 

```my_bc = DirichletBC(value=2, field=1)```

But we could do something similar to what we do on the `fenicsx` branch:

```python
my_trap = F.Trap(....)

my_export = F.TotalVolume(..., field=my_trap)
```

KulaginVladimir:
I agree. 

However, I think that keeping ids to use for traps is fine, but they shouldn't be available for the users. I mean that `trap.id` should be a unique attribute related to the actual index of a trap in the traps list. For an easier access to specific traps, I suppose other attributes (like trap label or tag) or approaches (as you mentioned above) would be better.

RemDelaporteMathurin:
> I mean that `trap.id` should be a unique attribute related to the actual index of a trap in the traps list.

In that case you don't need to have .id as an attribute of the trap itself since you can just do `self.traps.index(trap)` to have the index of a specific trap.

the id attribute as an int is just used for creating the functionspace, and here we can just do

```
for id, trap in enumerate(self.traps, 1):
```

Just like we already do in this new method

KulaginVladimir:
oh, yes. I just meant that, personally, I find convinient to define export for (and then access) traps with their ids. Though it might be a matter of taste.

RemDelaporteMathurin:
> oh, yes. I just meant that, personally, I find convinient to define export for (and then access) traps with their ids. Though it might be a matter of taste.

Right, so i think we need to make a different between the index of the traps in the list (only used for constructing the function space) and the tag/name of the trap that is a user-friendly way of referring to it.

We can have users refer to a trap in for example exports by either using the actual trap object, or its name, or its index in the trap list.
We remove the id as a user-defined argument and just have it for convenience as an attribute (maybe).
Then we simply have to forbid users to use digits-only names.

Anyway we can merge this PR and ship the discussion elswehere

PR #812, opened 2024-07-12
Error if temperature BC or source with `F.Temperature`

JonathanGSDUFOUR:
## Proposed changes

Added check in the `generic_simulation` to prevent the use of festim.Temperature with boundary condition set on the `T` field (Issue #787 ).

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

Some other tests had to be slightly modified, is there a better way than what I did ?

### DIFF
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -207,9 +207,9 @@ class Simulation:
         # set sources
         for source in self.sources:
             if source.field == "T" and not isinstance(
-                self.T, festim.temperature.temperature_solver.HeatTransferProblem
+                self.T, festim.HeatTransferProblem
             ):  # check that there is not a source defined in T as the same time as a festim.Temperature
-                raise ValueError(
+                raise TypeError(
                     "Heat transfer sources can only be used with HeatTransferProblem"
                 )
             if isinstance(source, festim.RadioactiveDecay) and source.field == "all":
@@ -249,10 +249,8 @@ class Simulation:
                 raise ValueError("SurfaceKinetics can only be used in 1D simulations")
 
             # check that there is not a Temperature defined at the same time as a boundary condition in T
-            if bc.field == "T" and not isinstance(
-                self.T, festim.temperature.temperature_solver.HeatTransferProblem
-            ):
-                raise ValueError(
+            if bc.field == "T" and not isinstance(self.T, festim.HeatTransferProblem):
+                raise TypeError(
                     "Heat transfer boundary conditions can only be used with HeatTransferProblem"
                 )
 
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -242,6 +242,14 @@ class Simulation:
             ):
                 raise ValueError("SurfaceKinetics can only be used in 1D simulations")
 
+            # check that there is not a Temperature defined at the same time as a Dirichlet in T
+            if bc.field == "T" and not isinstance(
+                self.T, festim.temperature.temperature_solver.HeatTransferProblem
+            ):
+                raise ValueError(
+                    f"cannot use boundary conditions with Temperature, use HeatTransferProblem instead."
+                )
+
             # checks that DirichletBC or SurfaceKinetics is not set with another bc on the same surface
             # iterate through all BCs
             for dc_sk_bc in dc_sk_bcs:
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -206,6 +206,12 @@ class Simulation:
 
         # set sources
         for source in self.sources:
+            if source.field == "T" and not isinstance(
+                self.T, festim.temperature.temperature_solver.HeatTransferProblem
+            ):  # check that there is not a source defined in T as the same time as a festim.Temperature
+                raise ValueError(
+                    "Heat transfer sources can only be used with HeatTransferProblem"
+                )
             if isinstance(source, festim.RadioactiveDecay) and source.field == "all":
                 # assign source to each of the unique festim.Concentration
                 # objects in field_to_object
@@ -242,12 +248,12 @@ class Simulation:
             ):
                 raise ValueError("SurfaceKinetics can only be used in 1D simulations")
 
-            # check that there is not a Temperature defined at the same time as a Dirichlet in T
+            # check that there is not a Temperature defined at the same time as a boundary condition in T
             if bc.field == "T" and not isinstance(
                 self.T, festim.temperature.temperature_solver.HeatTransferProblem
             ):
                 raise ValueError(
-                    f"cannot use boundary conditions with Temperature, use HeatTransferProblem instead."
+                    "Heat transfer boundary conditions can only be used with HeatTransferProblem"
                 )
 
             # checks that DirichletBC or SurfaceKinetics is not set with another bc on the same surface
--- a/test/system/test_misc.py
+++ b/test/system/test_misc.py
@@ -170,9 +170,65 @@ def test_error_DirichletBC_on_same_surface(field, surfaces):
     ):  # with festim.Temperature already defined there is another error given
         with pytest.raises(
             ValueError,
-            match="cannot use boundary conditions with Temperature, use HeatTransferProblem instead.",
+            match="Heat transfer boundary conditions can only be used with HeatTransferProblem",
         ):
             sim.initialise()
+
+
+@pytest.mark.parametrize(
+    "bc",
+    [
+        F.FluxBC(value=1, field="T", surfaces=1),
+        F.ConvectiveFlux(h_coeff=1, T_ext=1, surfaces=1),
+        F.DirichletBC(value=1, surfaces=1, field="T"),
+    ],
+)
+def test_BC_on_T_with_Temp(bc):
+    """
+    Test that the code returns an error when a boundary condition is set
+    on the Temperature field when there is also a Festim.Temperature already set.
+    """
+
+    sim = F.Simulation()
+    sim.mesh = F.MeshFromVertices([0, 1, 2, 3])
+    sim.materials = F.Material(id=1, D_0=1, E_D=0)
+    sim.T = F.Temperature(value=500)
+    sim.boundary_conditions = [bc]
+    sim.settings = F.Settings(
+        transient=False, absolute_tolerance=1e8, relative_tolerance=1e-8
+    )
+    sim.sources = []
+    sim.dt = None
+    sim.exports = []
+    with pytest.raises(
+        ValueError,
+        match="Heat transfer boundary conditions can only be used with HeatTransferProblem",
+    ):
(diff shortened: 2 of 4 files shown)


RemDelaporteMathurin:
Since this is about the `HeatTransferProblem` class I'll let @KulaginVladimir review this first

test/unit/test_initial_condition.py
```diff
@@ -19,3 +20,32 @@ def test_error_initialisation_from_xdmf_missing_label():
     """
     with pytest.raises(ValueError, match=r"label"):
         festim.InitialCondition(value="my_file.xdmf", label=None, time_step=1)
+
+
+def test_temperature_and_DirichletBC():
```
KulaginVladimir, line 25: I suppose this test is similar to the one above. Moreover, it's not really related to inital conditions.

KulaginVladimir (changes requested):
Hi @JonathanGSDUFOUR. Thank you for this fix. 

I have a few comments on the current implementation below. Moreover, it appears to me that your fix covers only the case when boundary conditions for the temperature field are set simultaniously with `F.Temperature`, but doesn't catch the cases when a source or an initial condition are set together with `F.Temperature` which were mentioned at the start of #787. However, I think the latter can be discussed with @RemDelaporteMathurin after his review.

festim/generic_simulation.py
```diff
@@ -242,6 +242,14 @@ def check_boundary_conditions(self):
             ):
                 raise ValueError("SurfaceKinetics can only be used in 1D simulations")
 
+            # check that there is not a Temperature defined at the same time as a Dirichlet in T
+            if bc.field == "T" and not isinstance(
+                self.T, festim.temperature.temperature_solver.HeatTransferProblem
+            ):
+                raise ValueError(
+                    f"cannot use boundary conditions with Temperature, use HeatTransferProblem instead."
```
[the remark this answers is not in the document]
RemDelaporteMathurin: Agreed! Maybe something like `heat transfer boundary conditions can only be used with HeatTransferProblem`

test/system/test_misc.py
```diff
@@ -159,11 +159,20 @@ def test_error_DirichletBC_on_same_surface(field, surfaces):
...
+        ):
+            sim.initialise()
+    elif (
+        field == "T"
+    ):  # with festim.Temperature already defined there is another error given
+        with pytest.raises(
+            ValueError,
+            match="cannot use boundary conditions with Temperature, use HeatTransferProblem instead.",
+        ):
+            sim.initialise()
```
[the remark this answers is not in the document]
RemDelaporteMathurin: Agreed, instead of adding an error check, we should adapt the test and replace by a heat transfer model

test/unit/test_boundary_conditions.py
```diff
@@ -551,7 +551,7 @@ def custom_fun(T, solute, param1):
     "bc",
     [
         (festim.DissociationFlux(surfaces=[1], Kd_0=1, E_Kd=0, P=1e4)),
-        (festim.ConvectiveFlux(h_coeff=1, T_ext=1, surfaces=1)),
+        # (festim.ConvectiveFlux(h_coeff=1, T_ext=1, surfaces=1)),
```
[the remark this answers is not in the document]
RemDelaporteMathurin: Agreed!
Either remove the comment or as for another test, replace by `HeatTransferProblem`

test/unit/test_boundary_conditions.py
```diff
@@ -605,6 +605,26 @@ def test_flux_BC_initialise(bc):
     sim.initialise()
 
 
+def test_flux_BC_initialise_special_case():
```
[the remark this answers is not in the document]
RemDelaporteMathurin: Yes this would cover more cases

test/unit/test_initial_condition.py
```diff
@@ -1,5 +1,6 @@
 import festim
 import pytest
+import numpy
```
RemDelaporteMathurin, line 3:
```suggestion
import numpy as np
```

test/unit/test_initial_condition.py
```diff
@@ -19,3 +20,32 @@ def test_error_initialisation_from_xdmf_missing_label():
...
+def test_temperature_and_DirichletBC():
+    """
+    Test that the code returns an error when a Dirichlet BC is set with T when there
+    is also a Festim.Temperature already set
+    """
+    my_model = festim.Simulation()
+
+    my_model.T = festim.Temperature(873)
+
+    my_model.mesh = festim.MeshFromVertices(vertices=numpy.linspace(0, 1, num=1000))
```
RemDelaporteMathurin, line 34:
```suggestion
    my_model.mesh = festim.MeshFromVertices(vertices=np.linspace(0, 1, num=1000))
```

RemDelaporteMathurin (changes requested):
I agree with @KulaginVladimir we would need to perform this check also if `Source` objects are set.

Vladimir, initial temperatures are set as an argument of `HeatTransferProblem` right? so maybe no need to check this too? idk

KulaginVladimir:
> Vladimir, initial temperatures are set as an argument of HeatTransferProblem right? so maybe no need to check this too? idk

Indeed. There is no need in such a check. I assume that I was thinking of the `HTransportProblem` and just transferred the thoughts to the `HeatTransferProblem`. What happens if we conduct a steady-state simulation and set an initial condition for solute or traps?

RemDelaporteMathurin:
> What happens if we conduct a steady-state simulation and set an initial condition for solute or traps?

I'm not entirely sure that an error is raised

JonathanGSDUFOUR:
Thank you @KulaginVladimir and @RemDelaporteMathurin for the feedback !
I started with the test in `test_initial_conditions` but realised `Convective_flux` had to be removed from the `test_flux_BC_initialise` because of the new error message. I will clean the duplication and extend the checks to 'Sources' too.
I'll make sure to learn from those errors !

festim/generic_simulation.py
```diff
@@ -206,6 +206,12 @@ def attribute_source_terms(self):
 
         # set sources
         for source in self.sources:
+            if source.field == "T" and not isinstance(
+                self.T, festim.temperature.temperature_solver.HeatTransferProblem
+            ):  # check that there is not a source defined in T as the same time as a festim.Temperature
+                raise ValueError(
```
RemDelaporteMathurin, line 212: Maybe this should be a `TypeError` instead since the issue is that the temperature doesn't have the right type?

```suggestion
                raise TypeError(
```

festim/generic_simulation.py
```diff
@@ -242,12 +248,12 @@ def check_boundary_conditions(self):
             ):
                 raise ValueError("SurfaceKinetics can only be used in 1D simulations")
 
-            # check that there is not a Temperature defined at the same time as a Dirichlet in T
+            # check that there is not a Temperature defined at the same time as a boundary condition in T
             if bc.field == "T" and not isinstance(
                 self.T, festim.temperature.temperature_solver.HeatTransferProblem
             ):
                 raise ValueError(
```
RemDelaporteMathurin, line 255:
```suggestion
                raise TypeError(
```

festim/generic_simulation.py
```diff
@@ -206,6 +206,12 @@ def attribute_source_terms(self):
 
         # set sources
         for source in self.sources:
+            if source.field == "T" and not isinstance(
+                self.T, festim.temperature.temperature_solver.HeatTransferProblem
```
RemDelaporteMathurin, line 210:
```suggestion
                self.T, festim.HeatTransferProblem
```

festim/generic_simulation.py
```diff
@@ -242,12 +248,12 @@ def check_boundary_conditions(self):
             ):
                 raise ValueError("SurfaceKinetics can only be used in 1D simulations")
 
-            # check that there is not a Temperature defined at the same time as a Dirichlet in T
+            # check that there is not a Temperature defined at the same time as a boundary condition in T
             if bc.field == "T" and not isinstance(
                 self.T, festim.temperature.temperature_solver.HeatTransferProblem
```
RemDelaporteMathurin, line 253:
```suggestion
                self.T, festim.HeatTransferProblem
```

test/system/test_misc.py
```diff
@@ -170,11 +170,67 @@ def test_error_DirichletBC_on_same_surface(field, surfaces):
...
+    sim.mesh = F.MeshFromVertices([0, 1, 2, 3])
+    sim.materials = F.Material(id=1, D_0=1, E_D=0)
+    sim.T = F.Temperature(value=500)
+    sim.boundary_conditions = [bc]
+    sim.settings = F.Settings(
+        transient=False, absolute_tolerance=1e8, relative_tolerance=1e-8
+    )
+    sim.sources = []
+    sim.dt = None
+    sim.exports = []
```
RemDelaporteMathurin, line 202: These are already the default values

```suggestion
```

test/system/test_misc.py
```diff
@@ -170,11 +170,67 @@ def test_error_DirichletBC_on_same_surface(field, surfaces):
...
+    sim.mesh = F.MeshFromVertices([0, 1, 2, 3])
+    sim.materials = F.Material(id=1, D_0=1, E_D=0)
+    sim.T = F.Temperature(value=500)
+    sim.boundary_conditions = []
+    sim.settings = F.Settings(
+        transient=False, absolute_tolerance=1e8, relative_tolerance=1e-8
+    )
+    sim.sources = [F.Source(value=1, volume=1, field="T")]
+    sim.dt = None
+    sim.exports = []
```
RemDelaporteMathurin, line 226:
```suggestion
```

RemDelaporteMathurin (changes requested):
Hey @JonathanGSDUFOUR here are a few very small changes 👍 

After that I think it's good to go, unless @KulaginVladimir sees something else

RemDelaporteMathurin (approved):
I'll let this open for a little while longer just in case someone else has comments

test/system/test_misc.py
```diff
@@ -159,9 +159,69 @@ def test_error_DirichletBC_on_same_surface(field, surfaces):
...
+    """
+    Test that the code returns an error when a source is set
+    on the Temperature field when there is also a Festim.Temperature already set.
+    """
+
+    sim = F.Simulation()
+    sim.mesh = F.MeshFromVertices([0, 1, 2, 3])
+    sim.materials = F.Material(id=1, D_0=1, E_D=0)
+    sim.T = F.Temperature(value=500)
+    sim.boundary_conditions = []
```
KulaginVladimir, line 217: I suppose this line is also not required
```suggestion
```

RemDelaporteMathurin:
Thank you so much for this contribution Jonathan!! 🎉

PR #816, opened 2024-07-23
Radioactive decay in Theory

### LINKED ISSUE #785: Documentation" add formulation and theory of radioactive decay
Thanks @rekomodo

RemDelaporteMathurin:
## Proposed changes

Fixes #785 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

### DIFF
--- a/docs/source/theory.rst
+++ b/docs/source/theory.rst
@@ -113,6 +113,19 @@ According to the latter, the rate :math:`k(T)` of a thermally activated process 
 
 where :math:`k_0` is the pre-exponential factor, :math:`E_k` is the process activation energy, and :math:`T` is the temperature. 
 
+Radioactive decay
+^^^^^^^^^^^^^^^^^
+
+Radioactive decay can be included in FESTIM. Neglecting other terms (diffusion, trapping...), the temporal evolution of the concentration of a species :math:`c` is governed by:
+
+.. math::
+    :label: eq_radioactive_decay
+
+    \frac{\partial c}{ \partial t} = -\lambda c
+
+where :math:`\lambda` is the decay constant.
+In FESTIM it is possible to include radioactive decay in the mobile hydrogen concentration and/or the trapped hydrogen concentrations.
+
 Heat transfer
 ^^^^^^^^^^^^^^
 To properly account for the temperature-dependent parameters, an accurate representation of the temperature field is often required. FESTIM can solve a heat transfer problem governed by the heat equation:


docs/source/theory.rst
```diff
@@ -113,6 +113,19 @@ According to the latter, the rate :math:`k(T)` of a thermally activated process
...
+^^^^^^^^^^^^^^^^^
+
+Radioactive decay can be included in FESTIM. Neglecting other terms (diffusion, trapping...), the temporal evolution of the concentration of a species :math:`c` is governed by:
+
+.. math::
+    :label: eq_radioactive_decay
+
+    \frac{\partial c}{ \partial t} = -\lambda c
+
+where :math:`\lambda` is the decay constant.
```
KulaginVladimir, line 126:
```suggestion
where :math:`\lambda\,[\mathrm{s}^{-1}]` is the decay constant.
```

docs/source/theory.rst
```diff
@@ -113,6 +113,19 @@ According to the latter, the rate :math:`k(T)` of a thermally activated process
...
+In FESTIM it is possible to include radioactive decay in the mobile hydrogen concentration and/or the trapped hydrogen concentrations.
```
KulaginVladimir, line 127: Concerning the statement formulation, I'd write that radioactive decay can be included in the concentration evolution or considered for mobile/trapped species, rather than _include radioactive decay in the mobile hydrogen concentration_. What do you think?

KulaginVladimir (changes requested):
Fair good PR. One major comment concerns the units. We discussed earlier that it might be beneficial to also add the units in the theory section.

docs/source/theory.rst
RemDelaporteMathurin: Yes I was wondering about adding units then realised that units were never specified anywhere in the Theory section.

KulaginVladimir: They are in #808

docs/source/theory.rst
RemDelaporteMathurin: Done

docs/source/theory.rst
RemDelaporteMathurin: Right so maybe we would need to go through the whole theory section and make sure units are everywhere then

RemDelaporteMathurin: Nvm I forgot that you added units everywhere! 🚀

KulaginVladimir (approved):

Issue #830, opened 2024-07-30
[BUG] `SurfaceFlux` doesn't account for Soret effect


RemDelaporteMathurin:
**Describe the bug**
When running a simulation with the Soret effect, I realised that setting a `SurfaceFlux` derived quantity would result in the wrong values. This is because the `soret` flag in `compute` seems to be ignored.

**To Reproduce**
Simple 1D steady state simulation, checking the flux balance with soret activated.

```python
import numpy as np
import festim as F

my_model = F.Simulation()
my_model.mesh = F.MeshFromVertices(vertices=np.linspace(0, 1, num=100))
my_model.materials = F.Material(id=1, D_0=1, E_D=0, Q=2)
my_model.T = F.Temperature(value=700 + 1000 * F.x)

my_model.boundary_conditions = [
    F.DirichletBC(surfaces=[1], value=10, field=0),
    F.DirichletBC(surfaces=[2], value=1, field=0),
]
my_model.settings = F.Settings(
    absolute_tolerance=1e-10, relative_tolerance=1e-10, transient=False, soret=True
)

flux_left = F.SurfaceFlux(field=0, surface=1)
flux_right = F.SurfaceFlux(field=0, surface=2)
my_model.exports = [F.DerivedQuantities([flux_left, flux_right])]

my_model.initialise()
my_model.run()


# these two should be equal in steady state
print(flux_left.data)
print(flux_right.data)
```

Produces:
```
[370.7625458744958]
[1.1058666646404112]
```

**Expected behavior**
The two fluxes should have the same absolute value.

To convince us even further, try and add `assert False` in the compute method with soret:

```python
    def compute(self, soret=False):
        flux = f.assemble(
            self.prop * f.dot(f.grad(self.function), self.n) * self.ds(self.surface)
        )
        if soret and self.field in [0, "0", "solute"]:
            assert False
            flux += f.assemble(
                self.prop
                * self.function
                * self.Q
                / (k_B * self.T**2)
                * f.dot(f.grad(self.T), self.n)
                * self.ds(self.surface)
            )
        return flux
```

And you will see everything runs fine.

Moreover, I found this TODO comment:
https://github.com/festim-dev/FESTIM/blob/40cdd22f0832d6bcc8f4ca2172d2965dfba2d5ab/festim/exports/derived_quantities/derived_quantities.py#L157

PR #831, opened 2024-07-30
Fix SurfaceFlux with Soret

RemDelaporteMathurin:
## Proposed changes

This is a fix for #830.

The `compute` method of `SurfaceFlux` had a `soret` argument defaulting to `False`. However, this was not integrated correctly with the rest of the implementation.

What I did is I removed this argument and replaced it by an attribute of `SurfaceFlux`. This attribute is set to `True` or `False` during `Simulation.initialise`. Also, the `T` attribute of `SurfaceFlux` is set at the same time in order to calculate the soret flux.

This bug did not affect any of the actual solving step and the equations were correctly solved with Soret. This is only a post-processing bug with SurfaceFlux (the other derived quantities are unaffected).

Note: I propagated this fix to the cylindrical and spherical versions of `SurfaceFlux`.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -418,8 +418,7 @@ class Simulation:
                         self.dt.milestones.append(time)
                 self.dt.milestones.sort()
 
-        # set Soret to True
-        for export in self.exports:
+            # set Soret to True for SurfaceFlux quantities
             if isinstance(export, festim.DerivedQuantities):
                 for q in export:
                     if isinstance(q, festim.SurfaceFlux):
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -418,6 +418,14 @@ class Simulation:
                         self.dt.milestones.append(time)
                 self.dt.milestones.sort()
 
+        # set Soret to True
+        for export in self.exports:
+            if isinstance(export, festim.DerivedQuantities):
+                for q in export:
+                    if isinstance(q, festim.SurfaceFlux):
+                        q.soret = self.settings.soret
+                        q.T = self.T.T
+
     def run(self, completion_tone=False):
         """Runs the model.
 
--- a/test/system/test_misc.py
+++ b/test/system/test_misc.py
@@ -796,15 +796,22 @@ def test_soret_surface_flux_mass_balance():
 
     1D steady state simulation with Soret on.
     Compute the surface flux on both surfaces.
-    The surface fluxes should be equal in magnitude but opposite in sign.
+    The surface fluxes should be equal in magnitude
+    but opposite in sign.
+
+    The case is made so that the concentration profile
+    is flat (ie. diffusive flux is zero) but there's a
+    high T gradient (ie. Soret flux is not zero).
     """
     my_model = F.Simulation()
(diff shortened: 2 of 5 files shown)


festim/generic_simulation.py
```diff
@@ -418,6 +418,14 @@ def initialise(self):
...
                 self.dt.milestones.sort()
 
+        # set Soret to True
+        for export in self.exports:
+            if isinstance(export, festim.DerivedQuantities):
+                for q in export:
+                    if isinstance(q, festim.SurfaceFlux):
+                        q.soret = self.settings.soret
+                        q.T = self.T.T
+
```
KulaginVladimir, line 428: Do you want to set the soret flag independently? There is the same `for export in self.exports` loop above.

test/system/test_misc.py
```diff
@@ -788,3 +788,54 @@ def test_catch_bug_804():
...
+    flux_left = F.SurfaceFlux(field=0, surface=1)
+    flux_right = F.SurfaceFlux(field=0, surface=2)
+    my_model.exports = [F.DerivedQuantities([flux_left, flux_right])]
+
+    my_model.initialise()
+    my_model.run()
+
+    # these two should be equal in steady state
+    print(flux_left.data[0])
+    print(flux_right.data[0])
```
KulaginVladimir, line 838: Do we need print-statements in tests?

test/system/test_misc.py
```diff
@@ -788,3 +788,54 @@ def test_catch_bug_804():
...
+    The surface fluxes should be equal in magnitude
+    but opposite in sign.
+
+    The case is made so that the concentration profile
+    is flat (ie. diffusive flux is zero) but there's a
+    high T gradient (ie. Soret flux is not zero).
+    """
+    my_model = F.Simulation()
+
+    my_model.mesh = F.MeshFromVertices(vertices=np.linspace(0, 0.05, num=100))
```
KulaginVladimir, line 808: Have you checked cylindrical/spherical? I suppose it should be ok.

KulaginVladimir (commented):
The fix seems ok.

I have a couple of questions below.

festim/generic_simulation.py
RemDelaporteMathurin: Good shout! This would require a serious refactoring anyway

test/system/test_misc.py
RemDelaporteMathurin: We don't _need_ them but it's nice to have some prints when he tests fail

test/system/test_misc.py
RemDelaporteMathurin: I haven't but I'll see if I can parametrise this test

festim/generic_simulation.py
KulaginVladimir: Indeed. In fact, there are two loops that go through exports. I don't think they are needed, but some refactoring can be done elsewhere.

KulaginVladimir (approved):
I suppose everything is ok.

Issue #829, opened 2024-07-30
`TotalSurface` and `AbsorbedHydrogen` raise warnings in cartesian


RemDelaporteMathurin:
We noticed in https://github.com/festim-dev/FESTIM-workshop/pull/67 that the derived quantities `TotalSurface` and `AbsorbedHydrogen` raised a warning even when used in cartesian meshes.

`AbsorbedHydrogen` should work with any system of coordinates.
`TotalSurface` shouldn't raise a warning in cartesian.

We should add them to:

https://github.com/festim-dev/FESTIM/blob/40cdd22f0832d6bcc8f4ca2172d2965dfba2d5ab/festim/generic_simulation.py#L360-L379

PR #832, opened 2024-07-30
allowed_meshes property for derived quantities

RemDelaporteMathurin:
## Proposed changes

Fix for #829 

This PR proposes to change the approach for knowing if certain quantities are allowed with various systems of coordinates (cartesian, cylindrical, spherical).

Instead of having a long list of allowed quantities for each mesh type in `Simulation.initialise` - which was very hard to maintain -, each quantity now has a property called `allowed_meshes` that is a list of mesh types that this property can be used with.
By default, the three mesh types are allowed.

When inheriting from the parent class, it is possible to override this property.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/exports/derived_quantities/average_surface.py
+++ b/festim/exports/derived_quantities/average_surface.py
@@ -29,6 +29,10 @@ class AverageSurface(SurfaceQuantity):
         super().__init__(field=field, surface=surface)
 
     @property
+    def allowed_meshes(self):
+        return ["cartesian"]
+
+    @property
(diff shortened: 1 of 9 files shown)


RemDelaporteMathurin:
@jhdark would you mind giving this a look? also, if we agree to go with this system, then you would have to add the same changes in #410

jhdark (approved):
looks good to me

PR #836, opened 2024-07-31
Fixed Simulation.exports setter

### LINKED ISSUE #833: [BUG] `Simulation.exports` should always be a list
**Describe the bug**
FESTIM expects a list for the exports attribute of `Simulation`.

However, no error is raised when users give something else.

**To Reproduce**

Run on main:

```python
import numpy as np
import festim as F

my_model = F.Simulation()
my_model.mesh = F.MeshFromVertices(vertices=np.linspace(0, 1, num=100))
my_model.materials = F.Material(id=1, D_0=1, E_D=0, Q=2)
my_model.T = F.Temperature(value=700 + 1000 * F.x)
    absolute_tolerance=1e-10, relative_tolerance=1e-10, transient=False, soret=True
)

flux_left = F.SurfaceFlux(field=0, surface=1)
flux_right = F.SurfaceFlux(field=0, surface=2)
my_model.exports = F.DerivedQuantities([flux_left, flux_right])

my_model.initialise()
my_model.run()

print(flux_left.data)
print(flux_right.data)
```

Produces an empty list for data of derived quantities (they are just ignored).

```
Defining initial values
Defining variational problem
Defining source terms
Defining boundary conditions
Solving steady state problem...
Solved problem in 0.00 s
[]
[]
```

**Expected behavior**
Either the exports setter should convert the user input to a list, or an error should be raised

RemDelaporteMathurin:
## Proposed changes

Fixes #833 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -154,12 +154,12 @@ class Simulation:
     def exports(self, value):
         if value is None:
             self._exports = festim.Exports([])
+        elif isinstance(value, (festim.Export, festim.DerivedQuantities)):
+            self._exports = festim.Exports([value])
         elif isinstance(value, festim.Exports):
             self._exports = value
         elif isinstance(value, list):
             self._exports = festim.Exports(value)
-        elif isinstance(value, festim.Export):
-            self._exports = festim.Exports([value])
         else:
             raise TypeError(
                 "accepted types for exports are list, festim.Export or festim.Exports"
--- a/festim/generic_simulation.py
+++ b/festim/generic_simulation.py
@@ -162,7 +162,7 @@ class Simulation:
             self._exports = festim.Exports(value)
         else:
             raise TypeError(
-                "accepted types for exports are list, festim.Export or festim.Exports"
+                "accepted types for exports are list, festim.DerivedQuantities, festim.Export or festim.Exports"
             )
 
     @property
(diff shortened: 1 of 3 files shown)


RemDelaporteMathurin:
@KulaginVladimir would you mind giving this a quick look? I think this is a result of these classes now inheriting from `list`

festim/generic_simulation.py
```diff
@@ -154,12 +154,12 @@ def exports(self):
...
+            self._exports = festim.Exports([value])
         elif isinstance(value, festim.Exports):
             self._exports = value
         elif isinstance(value, list):
             self._exports = festim.Exports(value)
-        elif isinstance(value, festim.Export):
-            self._exports = festim.Exports([value])
         else:
             raise TypeError(
                 "accepted types for exports are list, festim.Export or festim.Exports"
```
KulaginVladimir, line 165: This line should be changed since derived quantities could be passed, no?

KulaginVladimir (commented):
The fix should work 👌🏻

Unfortunately, I can't check it thoroughly at the moment. As I understand, the problem was only with derived quantities, right?

festim/generic_simulation.py
RemDelaporteMathurin: Yep, i just bumped it up a few lines and changed the type check

KulaginVladimir: I meant the error message. With your fix we can pass `festim.DerivedQunatities` directly to `Simulation.exports`, right?

RemDelaporteMathurin: Yes! Good shout!

KulaginVladimir (approved):

Issue #745, opened 2024-04-08
Refactor `HeatTransferProblem` and `HTransportProblem`


RemDelaporteMathurin:
`HeatTransferProblem` and `HTransportProblem` share a lot of functionality.

We would need a base class (`Problem`?) that has all this shared functionality and have `HeatTransferProblem` and `HTransportProblem` inherit from it.

PR #857, opened 2024-08-01
created shared Problem class for HeatTransferProblem and HTransportProblem

rekomodo:
Solve #745

## Proposed changes 

Created a common `Problem` class to house functionality common to both `HeatTransferProblem` and `HTransportProblem`.

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

`Problem` class has no docstring yet.

### DIFF
--- a/festim/heat_transfer_problem.py
+++ b/festim/heat_transfer_problem.py
@@ -62,8 +62,19 @@ class HeatTransferProblem(ProblemBase):
         self.create_solver()
         self.initialise_exports()
 
-    def define_function_space(self):
-        element_CG = super().define_function_spaces()
+    def define_function_spaces(self):
+        """Creates the function space of the model, creates a mixed element if
+        model is multispecies. Creates the main solution and previous solution
+        function u and u_n. Create global DG function spaces of degree 0 and 1
+        for the global diffusion coefficient"""
+
+        degree = 1
+        element_CG = basix.ufl.element(
+            basix.ElementFamily.P,
+            self.mesh.mesh.basix_cell(),
+            degree,
+            basix.LagrangeVariant.equispaced,
+        )
 
         self.function_space = fem.functionspace(self.mesh.mesh, element_CG)
 
--- a/festim/heat_transfer_problem.py
+++ b/festim/heat_transfer_problem.py
@@ -13,7 +13,12 @@ class HeatTransferProblem(ProblemBase):
         settings=None,
     ) -> None:
         super().__init__(
-            mesh, sources, exports, subdomains, boundary_conditions, settings
+            mesh=mesh,
+            sources=sources,
+            exports=exports,
+            subdomans=subdomains,
+            boundary_conditions=boundary_conditions,
+            settings=settings,
         )
 
         self.initial_condition = initial_condition
--- a/festim/heat_transfer_problem.py
+++ b/festim/heat_transfer_problem.py
@@ -1,7 +1,11 @@
-from .problem import *
+from dolfinx import fem
+from dolfinx.nls.petsc import NewtonSolver
+import basix
+import ufl
+from mpi4py import MPI
+import festim as F
 
-
-class HeatTransferProblem(ProblemBase):
+class HeatTransferProblem(F.ProblemBase):
     def __init__(
         self,
         mesh=None,
--- a/festim/heat_transfer_problem.py
+++ b/festim/heat_transfer_problem.py
@@ -5,6 +5,7 @@ import ufl
 from mpi4py import MPI
 import festim as F
 
+
 class HeatTransferProblem(F.ProblemBase):
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


festim/problem.py
```diff
@@ -0,0 +1,164 @@
...
+from dolfinx.nls.petsc import NewtonSolver
+import basix
+import ufl
+from mpi4py import MPI
+import numpy as np
+import tqdm.autonotebook
+import festim as F
+
+
+class Problem:
```
RemDelaporteMathurin, line 12: Should we rename this `ProblemBase` to understand that this is not a class that users should use but only inherit from?

What I mean is that you cannot use `Problem` on its own.

festim/problem.py
```diff
@@ -0,0 +1,164 @@
...
+
+        degree = 1
+        element_CG = basix.ufl.element(
+            basix.ElementFamily.P,
+            self.mesh.mesh.basix_cell(),
+            degree,
+            basix.LagrangeVariant.equispaced,
+        )
+
+        return element_CG
```
RemDelaporteMathurin, line 61: I understand this is shared functionality across classes. However this method doesn't do what it says in the docstrings (ie. define a function space), it only returns a finite element. Several options:
- have a bit of code duplication in each child class
- convert this into a property `element_CG` so that classes can call `self.element_CG`

Note: it would be beneficial to expose `degree` to users one day. Could we make a note of this and/or a new issue?

festim/problem.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def define_meshtags_and_measures(self):
+        """Defines the facet and volume meshtags of the model which are used
+        to define the measures fo the model, dx and ds"""
+
+        if isinstance(self.mesh, F.MeshFromXDMF):
+            self.facet_meshtags = self.mesh.define_surface_meshtags()
+            self.volume_meshtags = self.mesh.define_volume_meshtags()
```
RemDelaporteMathurin, line 69: I notice here a naming inconsistency (not related to this PR) between `surface_meshtags` and `facet_mesthags`

festim/heat_transfer_problem.py
```diff
@@ -21,21 +12,11 @@ def __init__(
         exports=None,
         settings=None,
     ) -> None:
-        self.mesh = mesh
-        self.subdomains = subdomains or []
+        super().__init__(
+            mesh, sources, exports, subdomains, boundary_conditions, settings
```
RemDelaporteMathurin, line 16: should we use the keywords here `mesh=mesh, sources=sources` etc

This would avoid errors when/if we change the order of the arguments in `Problem` one day.

Same comment for `HTransportProblem`

festim/hydrogen_transport_problem.py
```diff
@@ -1,16 +1,7 @@
...
-import basix.ufl
-import ufl
-from mpi4py import MPI
-import numpy as np
-import tqdm.auto as tqdm
-import festim as F
-
-
-class HydrogenTransportProblem:
+from .problem import *
```
RemDelaporteMathurin, line 1: You should avoid this as much as possible.

```suggestion
from festim import Problem
```

festim/heat_transfer_problem.py
```diff
@@ -1,16 +1,7 @@
...
-import ufl
-from mpi4py import MPI
-from dolfinx.mesh import meshtags
-import numpy as np
-import tqdm.autonotebook
-import festim as F
-
-
-class HeatTransferProblem:
+from .problem import *
```
RemDelaporteMathurin, line 1:
```suggestion
from festim import Problem
```

RemDelaporteMathurin (changes requested):
Very nice refactoring @rekomodo much needed!! 

Thanks for this first contribution to FESTIM.

I have a few first comments.

festim/problem.py
rekomodo: yeah, sounds good. I added a short docstring too

festim/problem.py
rekomodo: I think it's best to have a bit of code duplication then. As for exposing degree, I can do that in a small pull request after this, I'll add a note.

festim/problem.py
rekomodo: added a TODO for this

festim/hydrogen_transport_problem.py
rekomodo: getting this error when I try to do this

```
ImportError: cannot import name 'ProblemBase' from partially initialized module 'festim' (most likely due to a circular import) (/home/rekomodo/festim/FESTIM/festim/__init__.py)
```

I'm guessing that we can't avoid it here since we're trying to import the module to define the module

RemDelaporteMathurin (approved):
Thanks!

PR #865, opened 2024-08-02
Details for contribution workflow

RemDelaporteMathurin:
## Proposed changes

This just adds a few steps to the contribution workflow.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

### DIFF
--- a/docs/source/devguide/index.rst
+++ b/docs/source/devguide/index.rst
@@ -24,6 +24,8 @@ And/or contribute to the source code by:
 Contributing to the code
 ------------------------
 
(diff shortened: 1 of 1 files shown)


jhdark (commented):
This is certainly an improvement from what was there previously. However, as much as I understand that technically you don't need a fenics environment to contribute, I still think we should mention this, and that one should run tests locally before asking for a review etc.

RemDelaporteMathurin:
> and that one should run tests locally before asking for a review etc.

Well in that case, you _need_ an environment to contribute

Issue #723, opened 2024-03-27
Cylindrical and spherical versions of derived quantities


RemDelaporteMathurin:
Following what was done in #689 we need an implementation of the cylindrical and spherical versions of
- [x] TotalVolume #939
- [x] TotalSurface #939 
- [x] AverageVolume #820 
- [x] AverageSurface #820 

## 2D axisymmetric
In 2D axisymmetric, the surface element ds is
ds = r dr dtheta or ds = r dz dtheta
In both cases, in fenics integrating a function f is `assemble(f * r * ds)` where r is the radius.

dV = r dr dtheta dz
Integrating a function f in fenics is therefore `assemble(f * r * dx)` where r is the radius.

## 1D spherical
The surface element in 1D spherical is:
dS_r = r^2 sin(theta) dtheta dphi
The integral of f on a spherical surface is therefore
integral(f dS_r) = integral(f r^2 sin(theta) dtheta dphi)
                        = (phi2 - phi1) * (-cos(theta2) + cos(theta1)) * f r^2
The volume element is 

dV = r^2 sin(theta) dr dtheta dphi
the integral of f on a spherical volume is therefore

integral(f dV) = (phi2 - phi1) * (-cos(theta2) + cos(theta1)) * integral(f r^2 dr)

phi is the polar angle and theta is the azimuthal angle

jhdark:
@gabriele-ferrero If you wanna do the TotalVolume and TotalSurface ones I can do the Average ones

RemDelaporteMathurin:
@jhdark @gabriele-ferrero I "assigned" you to this issue. Let me know if you need help and if you _want_ to have this in 1.3

RemDelaporteMathurin:
@rekomodo is having a go at implementing the remaining classes

RemDelaporteMathurin:
@jhdark can we close this?

KulaginVladimir:
Thus, `HydrogenFlux` and `ThermalFlux` don't need updates?

RemDelaporteMathurin:
> Thus, `HydrogenFlux` and `ThermalFlux` don't need updates?

Oh I didn't think of this.... I guess right now we can just restrict these _convenience_ classes to cartesian. But we could make this clearer in the docs

PR #882, opened 2024-09-01
Total Volume for Cylindrical and Spherical meshes

rekomodo:
## Proposed changes

Implements TotalVolume derived quantities for both Cylindrical and Spherical meshes as per #723 .

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/exports/derived_quantities/total_volume.py
+++ b/festim/exports/derived_quantities/total_volume.py
@@ -1,5 +1,6 @@
 from festim import VolumeQuantity
 import fenics as f
+import numpy as np
 
 
 class TotalVolume(VolumeQuantity):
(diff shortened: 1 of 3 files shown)


festim/exports/derived_quantities/total_volume.py
RemDelaporteMathurin, line 1: For these new classes you will need to override the `export_unit` property.

The units of TotalVolumeCylindrical and TotalVolumeSpherical are always `H` (since it's a 3D equivalent).

For `TotalVolume` this property is defined here:
https://github.com/festim-dev/FESTIM/blob/9dfb219ded55fc2617abe0704e94f2a6db71fbf9/festim/exports/derived_quantities/total_volume.py#L38-L50

festim/exports/derived_quantities/total_volume.py
```diff
@@ -58,3 +59,128 @@ def title(self):
...
+
+    @polar_range.setter
+    def polar_range(self, value):
+        if value[0] < -np.pi or value[1] > np.pi:
+            raise ValueError("Polar range must be between - pi and pi")
+        self._polar_range = value
+
+    @property
+    def allowed_meshes(self):
+        return ["spherical"]
```
RemDelaporteMathurin, line 176: This line isn't tested. Could we add a simple test that checks that the only allowed mesh is `spherical`?

festim/exports/derived_quantities/total_volume.py
```diff
@@ -58,3 +59,128 @@ def title(self):
...
+        theta_0, theta_1 = self.azimuth_range
+        return (theta_1 - theta_0) * f.assemble(
+            self.function * self.r * self.dx(self.volume)
+        )
+
+
+class TotalVolumeSpherical(TotalVolume):
+    """
+    Computes the total value of a field in a given volume
+    int(f r dx)
```
RemDelaporteMathurin, line 120:
```suggestion
    int(f r**2 dx)
```

Btw, this should include the angular coverages too! Same comment for the Cylindrical version

festim/exports/derived_quantities/total_volume.py
```diff
@@ -58,3 +59,128 @@ def title(self):
...
+        return ["spherical"]
+
+    def compute(self):
+        phi_0, phi_1 = self.azimuth_range
+        theta_0, theta_1 = self.polar_range
+
+        return (
+            (phi_1 - phi_0)
+            * (theta_1 - theta_0)
+            * f.assemble(self.function * self.r**2 * self.dx(self.volume))
```
RemDelaporteMathurin, line 185: This is not correct. There should be a `cos` somewhere (see [this](https://github.com/festim-dev/FESTIM/blob/9dfb219ded55fc2617abe0704e94f2a6db71fbf9/festim/exports/derived_quantities/surface_flux.py#L261-L292)). 

Consider the expression of the volume element in spherical coordinates ([link](https://en.wikipedia.org/wiki/Spherical_coordinate_system)):
image

test/unit/test_exports/test_derived_quantities/test_total_volume.py
```diff
@@ -46,6 +47,127 @@ def test_minimum(self):
...
+    c = 2 * r
+
+    my_total.dx = dx
+    my_total.function = c
+    my_total.r = f.Expression("x[0]", degree=1)
+
+    az0, az1 = azimuth_range
+    th0, th1 = polar_range
+
+    expected_value = f.assemble((az1 - az0) * (th1 - th0) * c * r**2 * dx(volume))
```
RemDelaporteMathurin, line 143: This is **not** the expected value, see my comment above

RemDelaporteMathurin (changes requested):
Hi @rekomodo thanks for this contribution it will be much needed!

I made a few comments. The expression for spherical is wrong and needs fixing

RemDelaporteMathurin:
Closing as replaced by #939

Issue #819, opened 2024-07-26
`TXTExport` produces duplicates


KulaginVladimir:
As highlighted by @rekomodo on [discourse](https://festim.discourse.group/t/double-exports-in-txtexport/58), `TXTExport` produces duplicates in the output file, what may confuse a user. 

Some suggestions from @RemDelaporteMathurin to improve this behaviour:

> I guess the internal FESTIM process should be:
>
> 1. if there is only one material (or no conservation of chemical potential) then project to CG instead of DG
> 2. If there are interfaces, since TXTExport is only used in 1D, project to DG but only keep the duplicates where there are interfaces (based > on the materials border argument)

RemDelaporteMathurin:
I think to start with we could at least make sure that values in the txt file are sorted by x

KulaginVladimir:
@RemDelaporteMathurin 
> If there are interfaces, since TXTExport is only used in 1D, project to DG but only keep the duplicates where there are interfaces (based > on the materials border argument)

Based on your recent [example ](https://github.com/festim-dev/FESTIM-workshop/blob/main/tasks/task13.ipynb), such filtering might be more complex.

KulaginVladimir:
@RemDelaporteMathurin 

> If there are interfaces, since TXTExport is only used in 1D, project to DG but only keep the duplicates where there are interfaces (based > on the materials border argument)

if vertices of a mesh do not match the interfaces (borders), shall we try to keep duplicates near the interfaces?  

Say:
```python
model.materials = [
    F.Material(
        id=1,
        D_0=1,
        E_D=0,
        S_0=1,
        E_S=0,
        borders=[0,0.5]
    )
    F.Material(
        id=2,
        D_0=1,
        E_D=0,
        S_0=2,
        E_S=0,
        borders=[0.5,1.0]
    )
]

model.mesh = F.MeshFromVertices(
    np.linspace(0, 1, 30)
)
``` 

The closest vertices to the interface at `x=0.5` are `x=0.48275862` and `x=0.51724138`

RemDelaporteMathurin:
1. We shouldn't condone this behaviour from users. Of course FESTIM would run but we should maybe throw a warning "vertices don't match interfaces" or something
2. if users still decide to do this for whatever reason, then we only keep the duplicate where the "`fenics`" interface is. I guess it's the closest vertex.

So in your example, we could do something along the lines of:

```
for vertex in vertices:
    if not isclose(vertex, interface_x):
[...]

Issue #863, opened 2024-08-02
TXT export: store data in memory and write only at last timestep


RemDelaporteMathurin:
@jhdark suggested that instead of writing the TXT at each timestep, we could store the data in memory and write everything at the last time step.

We could have a user argument to choose the behaviour.

@KulaginVladimir what do you think?

KulaginVladimir:
> We could have a user argument to choose the behaviour.

Makes sense.

KulaginVladimir:
> We could have a user argument to choose the behaviour.

Do we need to write data at each timestep? 

In the current implementation, `TXTExport` writes data -> reads old data and append new -> writes data, and so on. The only benefit I see is that at some timestep we can run out of memory and get an error. However, we nonetherless read the whole dataset at each call of `TXTExport.write()`

RemDelaporteMathurin:
We are working with numpy arrays so it's unlikely that we will run out of memory.

Plus I don't see how not writing the data at each timestep would solve this? Maybe there's something I don't get here

PR #883, opened 2024-09-09
`TXTExport` refactoring

KulaginVladimir:
## Proposed changes

This PR removes deprecated class `TXTExports` and fixes issues #819 and #863.

To-do list:

- [x] Add possibility to write data at the last export time
- [x] Define function space depending on the chemical potential conservation flag
- [x] Add filter method to remove duplicates if `DG` is used
- [x] Tests

## Types of changes

- [x] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted


festim/exports/txt_export.py
```diff
@@ -51,19 +57,58 @@ def is_it_time_to_export(self, current_time):
...
+        # Find indices of first elements in duplicated pairs and mesh borders
+        _, unique_indx = np.unique(x, return_index=True)
+
+        # Combine both arrays of indices
+        combined_indx = np.concatenate([border_indx, unique_indx])
+
+        # Sort unique indices to return a slice
+        combined_indx = sorted(np.unique(combined_indx))
+
+        return data[combined_indx, :]
```
KulaginVladimir, line 101: @RemDelaporteMathurin
Considering older issues with small floats and `np.isclose`, maybe this method might suit for filtering? 

Also, I'm not sure about deleting elements from a data array, thus suggest slicing based on indices of the needed array rows (duplicates near interfaces and single elements from a pair of duplicated otherwise)

festim/exports/txt_export.py
```diff
@@ -51,19 +57,58 @@ def is_it_time_to_export(self, current_time):
...
+
+    def write(self, current_time, final_time, materials, chemical_pot):
+        # create a DG1 functionspace if chemical_pot is True
+        # else create a CG1 functionspace
+        if chemical_pot:
+            V = f.FunctionSpace(self.function.function_space().mesh(), "DG", 1)
+        else:
+            V = f.FunctionSpace(self.function.function_space().mesh(), "CG", 1)
+
+        solution = f.project(self.function, V)
```
RemDelaporteMathurin, line 111: Do we need to do this if `chemical_pot` is False?

festim/exports/txt_export.py
```diff
@@ -51,19 +57,58 @@ def is_it_time_to_export(self, current_time):
...
+
+        # Combine both arrays of indices
+        combined_indx = np.concatenate([border_indx, unique_indx])
+
+        # Sort unique indices to return a slice
+        combined_indx = sorted(np.unique(combined_indx))
+
+        return data[combined_indx, :]
+
+    def write(self, current_time, final_time, materials, chemical_pot):
```
RemDelaporteMathurin, line 103: This function now does a lot more than just "write" (it filters the data and has some logic etc).

Plus it seems like a lot of what is done here could be done only once.
For instance, `chemical_pot` will always be the same throughout the simulation, the materials will never change either!

Maybe we could put all this in a initialise step instead of doing this at each write?

festim/exports/txt_export.py
```diff
@@ -72,62 +117,40 @@ def write(self, current_time, steady):
...
 
-            # if steady or it is the first time to export
-            # write data
+            # create header if steady or it is the first time to export
             # else append new column to the existing file
-            if steady or self._first_time:
-                if steady:
-                    header = "x,t=steady"
+            if final_time is None or self._first_time:
+                if final_time is None:
```
RemDelaporteMathurin, line 123: Just for readability, it would be clearer to keep the `steady` variable here but add

```suggestion
            steady = final_time is None
            if steady or self._first_time:
                if steady:
```

festim/generic_simulation.py
```diff
@@ -503,7 +503,12 @@ def run_post_processing(self):
         self.update_post_processing_solutions()
 
         self.exports.t = self.t
-        self.exports.write(self.label_to_function, self.mesh.dx)
+        self.exports.write(
+            self.label_to_function,
+            self.mesh.dx,
+            self.materials,
+            self.settings.chemical_pot,
```
RemDelaporteMathurin, line 510: I'm always careful when adding parameters like this because if we keep adding them then these functions because very hard to manage...

RemDelaporteMathurin (changes requested):
Hi @KulaginVladimir thanks for taking care of this.

I'm wondering about adding more arguments to the `write` method.

Could we instead store the information in attributes? For example, instead of calling `filter_duplicates` at each `write` we could call it once at the beginning of the simulation (if chemical pot is on) then store `combined_indx` as an attribute (we would need more explicit variable names ofc).

Then, when we call `write` we don't need to pass `materials` and `chemical_pot`.
If `self.combined_indx` is `None` then don't do anything to the data, otherwise filter the data with `self.data = self.data[self.combined_indx, :]` or something like this.

What do you think?

festim/exports/txt_export.py
RemDelaporteMathurin: This looks good to me!

festim/exports/txt_export.py
KulaginVladimir: > Maybe we could put all this in a initialise step instead of doing this at each write?

somewhere [here](https://github.com/festim-dev/FESTIM/blob/a1e3aec1f73f39b339dedcf6946e2b5be13c8f6f/festim/generic_simulation.py#L390-L399)?

RemDelaporteMathurin: Yes

festim/exports/txt_export.py
KulaginVladimir: Maybe no. Please check new attempt

festim/generic_simulation.py
```diff
@@ -381,31 +381,41 @@ def initialise(self):
...
+                #  and material borders
+                project_to_DG = (
+                    self.settings.chemical_pot
+                    or self.settings.traps_element_type == "DG"
+                )
+                export.initialise_TXTExport(
+                    self.mesh.mesh,
+                    project_to_DG,
+                    self.materials,
+                )
```
KulaginVladimir, line 400: 1. I removed the repeating loop `for export in self.exports` below. I don't think there is a needto iterate through it twice durinig initialisation.
2. I temporarly added new method `initialise_TXTExport` (former `filter_duplicates`) that's called once.
3. Considering [this ](https://github.com/festim-dev/FESTIM-workshop/blob/main/tasks/task13.ipynb) task of FESTIM WS, I assume that we should project fields to `DG` if `settings.traps_element_type = "DG"`. I added such a check with the `project_to_DG` variable, but it doesn't help and some information is lost due to filtering. For example, here is how the second figure from the task looks like: 
image
since there are borders, the output data near the boundary looks like this:
<details><summary>Details</summary>
<p>

4.646464646464646964e-01,5.000000000000542899e-01
4.747474747474748069e-01,5.000000000000532907e-01
4.848484848484848619e-01,5.000000000000531797e-01
**4.949494949494949725e-01,5.000000000000528466e-01
4.949494949494949725e-01,5.000000000000526246e-01**
5.050505050505050830e-01,5.000000000000521805e-01
5.151515151515152491e-01,0.000000000000000000e+00
5.252525252525253041e-01,0.000000000000000000e+00

</p>
</details> 

if there are no `material.borders`, then (fig. 4 from the task):
image
and there are no duplicates in the output file.

festim/exports/txt_export.py
```diff
@@ -51,83 +67,101 @@ def is_it_time_to_export(self, current_time):
...
+                    unique_indices.append(indx)
+
+            self._unique_indices = np.array(unique_indices)
+
+        else:
+            # Get list of unique indices as integers
+            self._unique_indices = np.argsort(x_column, axis=0)[:, 0]
+
+        self.data = x_column[self._unique_indices]
+        self.header = "x"
```
KulaginVladimir, line 129: This method now initialises the exported data. `self._unique_indices` allows to get a sorted array without duplicates (or with those near interfaces) from an unsorted array of initial data. Possibly, there is a more efficient way of doing it?

KulaginVladimir (commented):
@RemDelaporteMathurin 
I refactored the filtering method according to your suggestions. However, there are several points I'd like to adress. Please, see my comments below.

festim/generic_simulation.py
RemDelaporteMathurin: There's not much we can do about 3. I'm afraid. That is why I always recommend working with XDMF because there is no alteration of the data.

festim/exports/txt_export.py
```diff
@@ -51,83 +67,101 @@ def is_it_time_to_export(self, current_time):
...
+            if np.isclose(current_time, final_time, atol=0):
+                # write at final time if exports at each timestep
+                return True
+        else:
+            if np.isclose(current_time, self.times[-1], atol=0):
+                # write at final time if exports at specific times
+                return True
+        return False
+
+    def initialise_TXTExport(self, mesh, project_to_DG=False, materials=None):
```
RemDelaporteMathurin, line 84: I think we should call this method something else. Maybe just `initialise` and then add docstrings to explain what it does

RemDelaporteMathurin (changes requested):
This looks good to me! Just some more small comments about documentation, naming...

🚀

festim/exports/txt_export.py
```diff
@@ -51,83 +67,101 @@ def is_it_time_to_export(self, current_time):
...
+
+        # if chemical_pot is True or trap_element_type is DG, get indices of duplicates near interfaces
+        # and indices of the first elements from a pair of duplicates otherwise
+        if project_to_DG:
+            # Collect all borders
+            borders = []
+            for material in materials:
+                if material.borders:
+                    for border in material.borders:
+                        borders.append(border)
```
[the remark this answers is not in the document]
KulaginVladimir: heh, I thought you would suggest the long variant for readability

festim/generic_simulation.py
KulaginVladimir: However, what I showed is the result of filtering. We could filter data only for solute H or I don't know. If it's ok, then I'd maybe add a warning message

festim/exports/txt_export.py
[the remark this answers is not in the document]
RemDelaporteMathurin: Yep, depends what we want: compact vs readable.

Let's keep the readable since the method isn't super long

docs/source/api/exports.rst
```diff
@@ -15,10 +15,6 @@ Exports
     :members:
     :show-inheritance:
 
-.. autoclass:: TXTExports
-    :members:
-    :show-inheritance:
-
```
KulaginVladimir, line 21: Concerning docs, I [built ](https://txtexport-refactoring.readthedocs.io/en/latest/)docs from my distro without errors. Therefore, I don't understand what's the issue

festim/exports/txt_export.py
```diff
@@ -44,90 +62,157 @@ def filename(self, value):
...
+            self.V = f.FunctionSpace(mesh, "DG", 1)
+        else:
+            self.V = f.FunctionSpace(mesh, "CG", 1)
+
+        x = f.interpolate(f.Expression("x[0]", degree=1), self.V)
+        x_column = np.transpose([x.vector()[:]])
+
+        # if project_to_DG is True, get indices of duplicates near interfaces
+        # and indices of the first elements from a pair of duplicates otherwise
+        if project_to_DG:
```
KulaginVladimir, line 139: I thought a bit about filtering, maybe we should allow users to control whether the data will be filtered or not?

festim/exports/txt_export.py
```diff
@@ -44,90 +76,157 @@ def filename(self, value):
...
+            # write if steady
+            return True
+        elif self.times is None:
+            if np.isclose(current_time, final_time, atol=0):
+                # write at final time if exports at each timestep
+                return True
+        else:
+            if np.isclose(current_time, self.times[-1], atol=0):
+                # write at final time if exports at specific times
+                return True
```
RemDelaporteMathurin, line 123:
```suggestion
        elif np.isclose(current_time, self.times[-1], atol=0):
            # write at final time if exports at specific times
            return True
```

festim/exports/txt_export.py
```diff
@@ -44,90 +76,157 @@ def filename(self, value):
...
+        return False
+
+    def initialise(self, mesh, project_to_DG=False, materials=None):
+        """
+        Initialises TXTExport. Depending on the project_to_DG flag, defines a function space (DG1 or CG1)
+        for projection of the exported field. After that, an unsorted array of mesh vertices is created for export.
+        The array is then used to obtain indices of sorted elements for the data export.
+
+        .. note::
+            If DG1 is used and filter flag is True, the duplicated vertices in the array are filtered except those near interfaces,
```
RemDelaporteMathurin, line 133:
```suggestion
            If DG1 is used and ``filter`` flag is True, the duplicated vertices in the array are filtered except those near interfaces,
```

test/unit/test_exports/test_txt_export.py
```diff
@@ -111,33 +111,45 @@ def test_sorted_by_x(self, my_export, function, mesh):
...
         current_time = 1
         my_export.function = function
+        my_export.filter = filter
         my_export.initialise(mesh, project_to_DG, materials)
         my_export.write(
             current_time=current_time,
             final_time=None,
         )
+        print(my_export._unique_indices)
         assert len(my_export.data) == export_len
```
RemDelaporteMathurin, line 153: Instead of a print you can do the following:

```suggestion
        assert len(my_export.data) == export_len, f"Wrong export length. Unique indices: {my_export._unique_indices}"
```

RemDelaporteMathurin (approved):
Hi @KulaginVladimir very sorry for being so late on this... This looks good! A couple of very minor comments, feel free to merge!

test/unit/test_exports/test_txt_export.py
KulaginVladimir: Didn't know about such a functionality, but I forgot to delete this print-statement

PR #893, opened 2024-10-08
Deprecate `MeshFromRefinements`

### LINKED ISSUE #855: Deprecate `MeshFromRefinements`
I think it is time to deprecate `MeshFromRefinements`.

It is a very old class that have been replaced by `MeshFromVertices` in pretty much all demos and examples.

We should:
- Remove `MeshFromRefinements`
- Remove the tests
- Remove the documentation

I don't think we need to worry about backwards compatibility. I don't know a single user who uses this class.

If for some reason someone really needs it, please let us know here so that we can assist you.

RemDelaporteMathurin:
Fixes #855


KulaginVladimir (approved):
Looks fine!

PR #909, opened 2024-10-31
Fix solver options

RemDelaporteMathurin:
## Proposed changes

I realised that when users provided PETSC options for any `F.ProblemBase` class, they were just ignored.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

### DIFF
--- a/src/festim/hydrogen_transport_problem.py
+++ b/src/festim/hydrogen_transport_problem.py
@@ -149,6 +149,7 @@ class HydrogenTransportProblem(problem.ProblemBase):
         settings=None,
         exports=None,
         traps=None,
+        petsc_options=None,
     ):
         super().__init__(
             mesh=mesh,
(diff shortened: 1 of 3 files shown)


RemDelaporteMathurin:
@jorgensd to test this is there a way to catch a PETSC warning (I'd like to add a `foo` parameter and check that PETSC complains)

RemDelaporteMathurin:
I added `"options_left"` by default but it doesn't raise an error, just prints more stuff:

```
#PETSc Option Table entries:
-nls_solve_coucoucou 3 # (source: code)
-nls_solve_ksp_type preonly # (source: code)
-nls_solve_pc_factor_mat_solver_type mumps # (source: code)
-nls_solve_pc_type lu # (source: code)
-options_left # (source: code)
#End of PETSc Option Table entries
WARNING! There are options you set that were not used!
WARNING! could be spelling mistake, etc!
There is one unused database option. It is:
Option left: name:-nls_solve_coucoucou value: 3 source: code
```

Without it it just prints:

```
WARNING! There are options you set that were not used!
WARNING! could be spelling mistake, etc!
There is one unused database option. It is:
Option left: name:-nls_solve_coucoucou value: 3 source: code
```

Issue #920, opened 2024-11-14
Bug in milestones for stepsize


RemDelaporteMathurin:
When the time stepping is about to overstep a milestone, instead of assigning the time difference `next_milestone - current_time` to the stepsize, we assign the actual `next_milestone`.

https://github.com/festim-dev/FESTIM/blob/0769a6f497922026623ff649df048df89fd3132d/src/festim/stepsize.py#L138-L144

Consider the following MWE to reproduce the bug:

```python
import festim as F

milestones = [1000]
current_dt_value = 10
current_time = 999

stepsize = F.Stepsize(
    initial_value=current_dt_value ,
    growth_factor=1,  # don't relax or shrink
    cutback_factor=1,
    target_nb_iterations=4,
    milestones=milestones,
)

new_value = stepsize.modify_value(value=current_dt_value, nb_iterations=2, t=current_time)
print(new_value)   # should be 1 but instead is 1000 (ie the value of the milestone)
```

@kaelyndunnell do you think you can push a fix?

PR #922, opened 2024-11-15
Fix timestep bug

kaelyndunnell:
## Proposed changes

Fixes bug in timestep implementation in accordance with issue #920.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/test/test_stepsize.py
+++ b/test/test_stepsize.py
@@ -185,7 +185,7 @@ def test_overshoot_milestone(nb_its=1, target=4):
         value=current_value, nb_iterations=nb_its, t=0.9
     )
 
-    expected_value = 1.3
+    expected_value = 1.3 - current_value
 
     assert new_value == expected_value
 
--- a/test/test_stepsize.py
+++ b/test/test_stepsize.py
@@ -181,9 +181,7 @@ def test_overshoot_milestone(nb_its=1, target=4):
     my_stepsize.milestones = [1.3]
 
     current_value = 100000
-    new_value = my_stepsize.modify_value(
-        value=current_value, nb_iterations=nb_its, t=1
-    )
+    new_value = my_stepsize.modify_value(value=current_value, nb_iterations=nb_its, t=1)
 
     expected_value = 1.3 - 1
 
--- a/test/test_stepsize.py
+++ b/test/test_stepsize.py
@@ -175,17 +175,17 @@ def test_overshoot_milestone(nb_its=1, target=4):
     """
 
     my_stepsize = F.Stepsize(initial_value=0.1)
-    my_stepsize.growth_factor = 5.0
+    my_stepsize.growth_factor = 1
     my_stepsize.target_nb_iterations = target
 
     my_stepsize.milestones = [1.3]
 
-    current_value = 0.5
+    current_value = 100000
     new_value = my_stepsize.modify_value(
-        value=current_value, nb_iterations=nb_its, t=0.9
+        value=current_value, nb_iterations=nb_its, t=1
     )
 
-    expected_value = 1.3 - current_value
+    expected_value = 1.3 - 1
 
     assert new_value == expected_value
 
--- a/test/test_stepsize.py
+++ b/test/test_stepsize.py
@@ -169,19 +169,19 @@ def test_next_milestone(milestones, current_time, expected_value):
     assert expected_value == next_milestone
 
 
-def test_overshoot_milestone(nb_its=1, target=4):
+def test_overshoot_milestone():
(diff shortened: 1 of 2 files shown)


test/test_stepsize.py
```diff
@@ -185,7 +185,7 @@ def test_overshoot_milestone(nb_its=1, target=4):
         value=current_value, nb_iterations=nb_its, t=0.9
     )
 
-    expected_value = 1.3
+    expected_value = 1.3 - current_value
```
RemDelaporteMathurin, line 188: For this test, you don't even need a growth factor > 1. Try this:

growth factor = 1

curent_dt_value = 100000  very large

current_time = 1  # next timestep will overshoot the milestone of 1.3

expected_value = 1.3 - current_time # you need to fix this !!!

RemDelaporteMathurin (changes requested):
Your test is wrong as you have the wrong expected value

RemDelaporteMathurin (changes requested):
Thanks! CI is passing.

It's failing on the nightly build of FEniCS but we'll wait a bit to see if that gets resolved.

Could you format your code please?

jorgensd:
It’s failing because of the test added by @garth-wells in https://github.com/FEniCS/dolfinx/pull/3502/files 
I am not sure why it shouldn’t be allowed to make a mixed element with a single element.

RemDelaporteMathurin (approved):

garth-wells:
> It’s failing because of the check added by @garth-wells in https://github.com/FEniCS/dolfinx/pull/3502/files I am not sure why it shouldn’t be allowed to make a mixed element with a single element.

Mixed elements do not have a well-defined value shape, whereas 'base' sub-elements of a mixed element do. An element should be what it is. If it's mixed it should be mixed, if it's not mixed it should not be mixed.

jorgensd:
> > It’s failing because of the check added by @garth-wells in https://github.com/FEniCS/dolfinx/pull/3502/files I am not sure why it shouldn’t be allowed to make a mixed element with a single element.
> 
> Mixed elements do not have a well-defined value shape, whereas 'base' sub-elements of a mixed element do. An element should be what it is. If it's mixed it should be mixed, if it's not mixed it should not be mixed.
@garth-wells 
I still don't see why you shouldn't be allowed to do `mixed_element([single_element])`.
This is a perfectly valid use-case for more generalized problems where one can have `M` number of species in a problem, where 1 species is a special case. It would be silly to have to have dual code paths for handling this in any application code.

garth-wells:
@RemDelaporteMathurin can you share an example of where you rely on a single (base) element to behave like a mixed element?

jorgensd:
> @RemDelaporteMathurin can you share an example of where you rely on a single (base) element to behave like a mixed element?

let’s say you want to support having multiple diffusion equations co-located in space (but with different function spaces. Then you would have a mixed element ([el_0,el_1]). In other parts of the domain (with submesh involved, you might only have a single species. One doesn’t one to treat those differently.

RemDelaporteMathurin:
> @RemDelaporteMathurin can you share an example of where you rely on a single (base) element to behave like a mixed element?

@garth-wells we do not _need_ a single (base) element to behave like a mixed element per se. It would just simplify the logic for us. For example, in the `HydrogenTransportProblem` class, this is how we assign functions (solutions, test functions, previous solutions, etc) to "species":

https://github.com/festim-dev/FESTIM/blob/effcbac86ba0620da1388dcee4e8424ce33939db/src/festim/hydrogen_transport_problem.py#L485-L511

We have this check `if self.multispecies:` in multiple places to check if there are more than one element and if so, treat it as a mixed element.

This is how we do it in the `HTransportProblemDiscontinuous` class (using a mixed domain approach):

https://github.com/festim-dev/FESTIM/blob/effcbac86ba0620da1388dcee4e8424ce33939db/src/festim/hydrogen_transport_problem.py#L990-L1021

Here we can have mixed element with only one element and we don't have to have these checks everywhere.

This is an example for assigning functions to species but we have other places in the code where this happens, for instance assigning BCs:

https://github.com/festim-dev/FESTIM/blob/effcbac86ba0620da1388dcee4e8424ce33939db/src/festim/hydrogen_transport_problem.py#L537-L586

Assigning ICs:

https://github.com/festim-dev/FESTIM/blob/effcbac86ba0620da1388dcee4e8424ce33939db/src/festim/hydrogen_transport_problem.py#L610-L642


Having everything as mixed elements would greatly simplify the logic on our side.

PR #937, opened 2025-01-13
Printing messages in parallel runs

jhdark:
## Proposed changes

Due to the way printing in python has been used within festim means that print messages are shown on every node, which in practice means printing the message repeatedly by the number of processors defined. For instance running the following script:

```python
import festim as F
import numpy as np

my_model = F.Simulation(
    mesh=F.MeshFromVertices(vertices=np.linspace(0, 1, num=1000)),
    materials=F.Material(id=1, D_0=1, E_D=0),
    temperature=300,
    settings=F.Settings(
        absolute_tolerance=1e-10,
        relative_tolerance=1e-10,
        transient=False,
    ),
)
my_model.initialise()
my_model.run()
``` 
running this in parallel using `mpirun -np 4 python mwe.py` with 4 processors would produce the following:

```
Defining initial values
Defining variational problem
Defining initial values
Defining variational problem
Defining initial values
Defining variational problem
Defining initial values
Defining variational problem
Defining source terms
Defining source terms
Defining source terms
Defining source terms
Defining boundary conditions
Solving steady state problem...
Defining boundary conditions
Solving steady state problem...
Defining boundary conditions
Solving steady state problem...
Defining boundary conditions
Solving steady state problem...
Solved problem in 0.00 s
Solved problem in 0.00 s
Solved problem in 0.00 s
Solved problem in 0.00 s
```

With these changes, the printing message is only done on the master process (or rank 0 process). Meaning printed messages will only appear once, no matter the number of processors used.

resulting in:
```
Defining initial values
Defining variational problem
Defining source terms
Defining boundary conditions
Solving steady state problem...
Solved problem in 0.00 s
```

making printed messaged much cleaner and easier for users to understand.

@KulaginVladimir don't suppose you know of anyway we could test this? I imagine it may have to be at the github actions level in order to run the test scripts in parallel. Which itself could open up a new can of worms as we should really be testing in parallel too.

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/festim/meshing/mesh_from_xdmf.py
+++ b/festim/meshing/mesh_from_xdmf.py
@@ -1,5 +1,5 @@
 import fenics as f
-from festim import Mesh
+from festim import Mesh, festim_print
 
 
 class MeshFromXDMF(Mesh):
@@ -43,8 +43,9 @@ class MeshFromXDMF(Mesh):
         f.XDMFFile(self.boundary_file).read(surface_markers, "f")
         surface_markers = f.MeshFunction("size_t", mesh, surface_markers)
 
-        if f.MPI.comm_world.rank == 0:
-            print("Succesfully load mesh with " + str(len(volume_markers)) + " cells")
+        festim_print(
+            "Succesfully load mesh with " + str(len(volume_markers)) + " cells"
+        )
 
         self.volume_markers = volume_markers
         self.surface_markers = surface_markers
--- a/festim/meshing/mesh_from_xdmf.py
+++ b/festim/meshing/mesh_from_xdmf.py
@@ -43,6 +43,8 @@ class MeshFromXDMF(Mesh):
         f.XDMFFile(self.boundary_file).read(surface_markers, "f")
         surface_markers = f.MeshFunction("size_t", mesh, surface_markers)
 
-        print("Succesfully load mesh with " + str(len(volume_markers)) + " cells")
+        if f.MPI.comm_world.rank == 0:
+            print("Succesfully load mesh with " + str(len(volume_markers)) + " cells")
+
         self.volume_markers = volume_markers
         self.surface_markers = surface_markers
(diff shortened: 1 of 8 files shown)


KulaginVladimir:
Hi @jhdark ! 

> @KulaginVladimir don't suppose you know of anyway we could test this?

Sorry, but I don't. Maybe, we could search for a way somewhere?

By the way, all Error/Warning messages are printed by each node when running in parallel, no?

jhdark:
Yes this is true, but error messages are exceptional cases, so might not be completely necessary.

RemDelaporteMathurin:
I agree with @jhdark errors shouldn't occur in a normal situation so that is not worth rewriting eveyrthing.

RemDelaporteMathurin (commented):
@jhdark should we add a function 

```python
def festim_print(msg):
    if MPI.comm_word.rank == 0:
        print(msg)
```

jhdark:
Can do, might make the script a little easier to read. And just apply it to the changes that have been made?

RemDelaporteMathurin:
> Can do, might make the script a little easier to read. And just apply it to the changes that have been made?

Yes

RemDelaporteMathurin (approved):

festim/meshing/mesh_from_xdmf.py
```diff
@@ -43,6 +43,9 @@ def define_markers(self):
         f.XDMFFile(self.boundary_file).read(surface_markers, "f")
         surface_markers = f.MeshFunction("size_t", mesh, surface_markers)
 
-        print("Succesfully load mesh with " + str(len(volume_markers)) + " cells")
+        festim_print(
+            "Succesfully load mesh with " + str(len(volume_markers)) + " cells"
+        )
```
KulaginVladimir, line 48: I think that this message should be printed with `print`, since the mesh cells are not always partitioned equally between the cpus.

PR #940, opened 2025-01-27
New method to handle and convert user inputs to FEniCS objects

jhdark:
## Statement of need

Many classes within festim have a function called either `create_value_fenics` or `create_value` which is there to convert a user given input and converts it to a relevant fenics object depending on what has been provided. However, many of these functions  have a lot of code duplication. Furthermore, in an effort to allow users to access given inputs, these same classes typically have the `value` attribute, then the `value_fenics`, and in some cases, an associated fem.Expression to update time dependent values within the function at each iteration.
 
## Proposed changes

This change introduces a new class within helpers called `Value` which will have one argument, input_value, given by the users and remains unformatted should the user wish to retrieve it. Then 3 further attributes: ufl_expression, fenics_interpolation_expression and fenics_object.

For the implementation, classes which have value as an argument, automatically convert the user given value to a Value class object with the given value as the `input_value` argument. The Value class itself then has two additional properties: `time_dependent` and `temperature_dependent` depending on the value given by the user. 

The class has two functions: convert_input_value which creates one of 3 acceptable FEniCS objects: fem.Constant, ufl.core.expr.Expr or fem.Function, depending on the input given by the user. In some cases such as with sources and fluxes, callable inputs given need only be passed the temperature or time or space as needed to create a ufl expression, and so an additional optional argument: up_to_ufl_expr is provided for this case. All the functions used in this conversion function are also provided in helpers if there are cases where just the function may be needed.

For the moment I have just updated the tests for `test_h_transport_problem.py` and `test_heat_transfer_problem.py`   to show that this could work for both types of problems, however I haven't delved into the multi-material classes yet to see how compatible this may be.

@RemDelaporteMathurin  I know you didn't want to automatically convert user given inputs, but this way they are still retrievable, what do you think?

Also very open to changing any of the naming of functions/attributes etc.

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments


src/festim/helpers.py
```diff
@@ -12,16 +15,265 @@ def as_fenics_constant(
...
+        value=value,
+        function_space=function_space,
+        mesh=mesh,
+        t=t,
+        temperature=temperature,
+    )
+    fenics_object = fem.Function(function_space)
+    fenics_object.interpolate(fenics_interpolation_expression)
+
+    return fenics_interpolation_expression, fenics_object
```
RemDelaporteMathurin, line 89: I don't think these functions are very useful. `as_fenics_interp_expr_and_function` basically calls `as_fenics_interpolation_expression` and adds two lines...

`as_fenics_interpolation_expression` calls `as_ufl_expression` and adds one line

jhdark: Oh yeah, meant to merge these two actually, totally forgot

src/festim/helpers.py
```diff
@@ -17,11 +21,257 @@ def as_fenics_constant(
...
+        value: the callable to convert
+        mesh: the mesh of the domain
+        t: the time
+        temperature: the temperature
+
+    Returns:
+        The mapped function
+    """
+
+    arguments = value.__code__.co_varnames
```
jorgensd, line 53:
```suggestion
    # Extract the input variable names in the callable function `value`
    arguments = value.__code__.co_varnames
```
I think `as_mapped_function` needs some more documentation, as this is fairly advanced interpreting of code.
What about something along the lines of:
```python
# Call  `value(*args)` with the supplied values `t` and `temperature` if the variable names in `*args` matches `t` or `T`. If `x` in `args` insert a `SpatialCoordinate`.

Example:
    def my_func(x, T):
         return x[0] + T
    new_obj = as_mapped_function(my_func, mesh=mesh, temperature=10)
    new_obj := ufl.SpatialCoordinate(mesh)[0] + 10
```
The type hints here should also reflect that `t`, and `temperature` is optional, and you should handle if `t` is None, but is an input to your function, i.e. what happens if:
```python
   def my_func(T):
        return T**2
   as_mapped_function(my_func, mesh=mesh)
```

# Given a function `lambda t, x, T:  value(t, x, T)` return

src/festim/helpers.py
```diff
@@ -17,11 +21,257 @@ def as_fenics_constant(
...
+        | ufl.core.expr.Expr
+        | fem.Function
+    )
+
+    ufl_expression: ufl.core.expr.Expr
+    fenics_interpolation_expression: fem.Expression
+    fenics_object: fem.Function | fem.Constant | ufl.core.expr.Expr
+
+    def __init__(self, input_value):
+        (self,)
```
jorgensd, line 138:
```suggestion
```

src/festim/helpers.py
```diff
@@ -17,11 +21,257 @@ def as_fenics_constant(
...
+                "Value must be a float, int, fem.Constant, np.ndarray, fem.Expression,"
+                f" ufl.core.expr.Expr, fem.Function, or callable not {value}"
+            )
+
+    @property
+    def time_dependent(self) -> bool:
+        """Returns true if the value given is time dependent"""
+        if self.input_value is None:
+            return False
+        if isinstance(self.input_value, fem.Constant | ufl.core.expr.Expr):
```
jorgensd, line 180: What do you imply by time-dependent here? Is it only if it is a `Callable`-function with `t` as an input argument? as expression and constants can be time-dependent, as they depend on variables that could be updated over time.

src/festim/helpers.py
```diff
@@ -17,11 +21,257 @@ def as_fenics_constant(
...
+        mesh: dolfinx.mesh.Mesh = None,
+        t: fem.Constant = None,
+        temperature: fem.Function | fem.Constant | ufl.core.expr.Expr = None,
+        up_to_ufl_expr: bool = False,
+    ):
+        """Converts a user given value to a relevent fenics object depending
+        on the type of the value provided
+
+        Args:
+            mesh (dolfinx.mesh.Mesh): the mesh of the domain
```
jorgensd, line 213: Same comment regarding mesh (not needed as input)

src/festim/helpers.py
jhdark: Yeah its just there to detect if t is given in the arguments of the callable, do you know of a way to detect if a `fem.Constant` or `fem. Expression` is time dependent?

RemDelaporteMathurin: I don't think there's a way to do this, FEniCS has no concept of time.

jorgensd: We could add some overloaded version of constant that we annotate this in

src/festim/helpers.py
```diff
@@ -17,11 +22,252 @@ def as_fenics_constant(
...
+        elif callable(value):
+            self._input_value = value
+        else:
+            raise TypeError(
+                "Value must be a float, int, fem.Constant, np.ndarray, fem.Expression,"
+                f" ufl.core.expr.Expr, fem.Function, or callable not {value}"
+            )
+
+    @property
+    def time_dependent(self) -> bool:
```
RemDelaporteMathurin, line 172:
```suggestion
    def explicit_time_dependent(self) -> bool:
```

Based on discussions with Jorgen we should rename this

src/festim/helpers.py
jorgensd: We decided that this should be called `explicit_time_dependency`.

RemDelaporteMathurin:
I'll release an alpha version before merging this if you don't mind

PR #941, opened 2025-02-06
`SurfaceKinetics` minor update

### LINKED ISSUE #925: Remove lambda_abs in `SurfaceKinetics`
I think $\lambda_\mathrm{abs}$ is redundant [here](https://github.com/festim-dev/FESTIM/blob/d5bb806810351f2639b0ddaab78e9aa12b594b49/festim/boundary_conditions/fluxes/surface_kinetics.py#L133-L154) and makes the BC less flexible. 

We can redefine $k_\mathrm{bs}$ as: $k_\mathrm{bs}\lambda_\mathrm{abs}\rightarrow k_\mathrm{bs}$, so $k_\mathrm{bs}$ could be named as the rate constant of transition with units of $[m s^{-1}]$.

KulaginVladimir:
## Proposed changes

This PR fixes #925 in order to remove the hard-coded parameter $\lambda_{abs}$. With this change, the definition of $k_{bs}$ in theory (as a rate constant) will be correct. 

Additionally, `J_vs`, `k_bs`, and `k_sb` now accept `mobile` cocnentration by default allowing to account for fluxes that might depend on the concentration of subsurface H atoms (see [Baskes 1980](https://www.sciencedirect.com/science/article/abs/pii/0022311580901178), [Richards 1988](https://www.sciencedirect.com/science/article/abs/pii/0022311588903339), [Pisarev 1997](https://www.sciencedirect.com/science/article/abs/pii/S0022311597002018#preview-section-references)). 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/test/unit/test_boundary_conditions.py
+++ b/test/unit/test_boundary_conditions.py
@@ -620,14 +620,14 @@ def test_create_form_surf_kinetics():
     """
 
     # build
-    def k_sb(cs, T, prm1, prm2):
-        return 2 * T + cs**2 + prm1 - prm2
+    def k_sb(T, cs, cm, prm1, prm2):
+        return 2 * T + cs**2 / cm + prm1 - prm2
 
-    def k_bs(cs, T, prm1, prm2):
-        return 2 * T + 3 * cs + prm1 + prm2
+    def k_bs(T, cs, cm, prm1, prm2):
+        return 2 * T + 3 * cs + cm + prm1 + prm2
 
-    def J_vs(cs, T, prm1, prm2):
-        return 2 * T + 1
+    def J_vs(T, cs, cm, prm1, prm2):
+        return 2 * T + 5 * cm - 3 * cs
 
     lambda_IS = 1
     n_surf = 1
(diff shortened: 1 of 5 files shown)


test/unit/test_boundary_conditions.py
RemDelaporteMathurin, line 1: IN this file I see that you changed the order of the arguments in functions `k_sb`, `k_bs`, etc. `cs` used to be the first argument and now it's the second one. However, later in lines 681 the temperature was given first. That should've made the tests fail no?

KulaginVladimir: indeed, I'll check

RemDelaporteMathurin:
@KulaginVladimir can you rebase on main to bring the changes from #942

test/unit/test_boundary_conditions.py
RemDelaporteMathurin: Sorry if there was some confusion. Let me clarify, this is the previous version (on main):

Here you create these functions:

https://github.com/festim-dev/FESTIM/blob/03ed7d54465eb720af2058b11ad42807388cdf27/test/unit/test_boundary_conditions.py#L623-L630

Later in the test you use these functions but you see that T and cs are swaped:

https://github.com/festim-dev/FESTIM/blob/03ed7d54465eb720af2058b11ad42807388cdf27/test/unit/test_boundary_conditions.py#L681-L683

KulaginVladimir: Right, I made a mistake with parameter names in functions that confuses. However, the same functions are used and parameters (`T`, `adsorbed`, ...) are passed in the same order to these functions to create the form in `SurfaceKinetics`. The test compares the final forms which are equal in both cases, so there is no difference as long as lines 681-683 are kept unchanged. Am I wrong?

RemDelaporteMathurin: ok I think i understand thanks!

PR #949, opened 2025-02-26
Export average surface temperature

kaelyndunnell:
## Proposed changes

Adds export for average surface temperature.

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

## Further comments


src/festim/exports/surface_temperature.py
```diff
@@ -0,0 +1,102 @@
+import csv
+from dolfinx import fem
+import ufl
+import festim as F
+
+class SurfaceTemperature:
```
RemDelaporteMathurin, line 6: I would be explicit an name it `AverageSurfaceTemperature`

RemDelaporteMathurin: This should inherit from `SurfaceQuantity`

```suggestion
class SurfaceTemperature(F.SurfaceQuantity):
```

src/festim/exports/surface_temperature.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @property
+    def surface(self):
+        return self._surface
+
+    @surface.setter
+    def surface(self, value):
+        if not isinstance(value, (int, F.SurfaceSubdomain)) or isinstance(value, bool):
+            raise TypeError("surface should be an int or F.SurfaceSubdomain")
+        self._surface = value
```
RemDelaporteMathurin, line 53: This is duplicated from `SurfaceQuantity`. Can be removed when this class inherits from it

src/festim/exports/surface_temperature.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @filename.setter
+    def filename(self, value):
+        if value is None:
+            self._filename = None
+        elif not isinstance(value, str):
+            raise TypeError("filename must be of type str")
+        elif not value.endswith(".csv") and not value.endswith(".txt"):
+            raise ValueError("filename must end with .csv or .txt")
+        self._filename = value
```
RemDelaporteMathurin, line 43: Can be removed if inherits from `SurfaceQuantity`

src/festim/exports/surface_temperature.py
```diff
@@ -0,0 +1,102 @@
...
+        if self.filename is not None:
+            if self._first_time_export:
+                header = ["t(s)", f"{self.title}"]
+                with open(self.filename, mode="w+", newline="") as file:
+                    writer = csv.writer(file)
+                    writer.writerow(header)
+                self._first_time_export = False
+            with open(self.filename, mode="a", newline="") as file:
+                writer = csv.writer(file)
+                writer.writerow([t, self.value])
```
RemDelaporteMathurin, line 102: Can be removed if inherits from `SurfaceQuantity`

src/festim/hydrogen_transport_problem.py
```diff
@@ -786,6 +793,8 @@ def update_time_dependent_values(self):
             if source.temperature_dependent:
                 source.update(t=t)
 
+    surface_temp_processed = False
```
RemDelaporteMathurin, line 796: What is this? the indentation is odd, is this a leftover?

src/festim/hydrogen_transport_problem.py
```diff
@@ -818,6 +827,17 @@ def post_processing(self):
...
                     export.write(t=float(self.t))
+
+            elif isinstance(export, exports.SurfaceTemperature):
+                export.compute(self.ds)  # compute surface temp
+
+                export.t.append(float(self.t))  # update export time
+
+                # if filename given write export data to file
+                if export.filename is not None:
+                    export.write(t=float(self.t))
```
RemDelaporteMathurin, line 838: These lines are not tested. That's because by design, it's impossible to use `SurfaceTemperature` in combination with `HTransportProblem`

RemDelaporteMathurin (changes requested):
Hi @kaelyndunnell thanks for this, this is very much needed!

However I think it should be implemented differently:

- It doesn't make sense to ask users to provide `temperature_field` as an argument of `SurfaceTemperature` since they don't have access to the fenics object (`fem.Constant` or `fem.Temperature`) before calling `initialise` and since exports are defined _before_ calling `initialise`. Instead, I believe the class should not have this argument at all, and `temperature_field` is set as an attribute only inside `HTransportProblem`

```python
for export in self.exports:
    if isinstance(export, SurfaceTemperature):
        export.temperature_field = self.temperature_fenics
```

- `SurfaceTemperature` shares a lot of functionality with `SurfaceQuantity`: it should inherit from it
- I'm thinking by looking at this we should revisit the design of these quantities @JonathanGSDUFOUR , I'll open an issue to describe my idea but here's the rough pitch

```python
class SurfaceQuantity:
    def __init__(self, surface, filename):
        self.surface = surface
        self.filename = filename

        self.field = None   # this is no longer a user argument but only an attribute
     
    ...

class AverageSurfaceConcentration(SurfaceQuantity):
    def __init__(self, species, surface, filename):
        self.species = species
        super.__init__(surface, filename)
```

Inside `HTransportProblem`

```python
def initialise_exports(self):
    for export in self.exports:
        if not hasattr(export, "species"):  # appropriate fenics functions are already stored in F.Species objects
            export.field = self.temperature_fenics 
```

src/festim/exports/surface_temperature.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    def __init__(self, surface, filename: str = None) -> None:
+        self.surface = surface
+        self.filename = filename
+
+        self.temperature_field = None
+        self.t = []
+        self.data = []
+        self._first_time_export = True
```
RemDelaporteMathurin, line 29: I would rather have use the super class init for consistency

```suggestion
    def __init__(self, surface, filename: str = None) -> None:

        self.temperature_field = None
        super().__init__(field=None, surface=surface, filename=filename)
```

src/festim/hydrogen_transport_problem.py
```diff
@@ -415,6 +418,9 @@ def initialise_exports(self):
                 export.D = D
                 export.D_expr = D_expr
 
+            if isinstance(export, exports.AverageSurfaceTemperature):
+                export.temperature_field = self.temperature_fenics
```
RemDelaporteMathurin, line 422: This should be tested!

test/test_surface_temperature.py
```diff
@@ -0,0 +1,191 @@
...
+    with pytest.raises(
+        TypeError, match="surface should be an int or F.SurfaceSubdomain"
+    ):
+        my_model = F.HydrogenTransportProblem(
+            mesh=F.Mesh1D(np.linspace(0, 6.0, 10000)), temperature=500
+        )
+        my_model.define_temperature()
+        F.AverageSurfaceTemperature(
+            surface="1",
+        )
```
RemDelaporteMathurin, line 191: These tests aren't need as it tests the functionality in the parent class `SurfaceQuantity`

test/test_surface_temperature.py
```diff
@@ -0,0 +1,191 @@
...
+    my_model.initialise_exports()
+
+    # RUN
+    for i in range(3):
+        my_model.t.value += dt.value
+        my_model.update_time_dependent_values()
+
+    my_export = F.AverageSurfaceTemperature(surface=dummy_surface)
+    my_export.temperature_field = my_model.temperature_fenics
+    my_export.compute(ds)
```
RemDelaporteMathurin, line 51: Good, but we also need to test the full integration with `HydrogenTransportProblem` ie. just do 

```
my_model.exports = [....]
my_model.initialise()
my_model.run()

# test value of export after run
```

src/festim/hydrogen_transport_problem.py
RemDelaporteMathurin: See comment below about integration test

src/festim/hydrogen_transport_problem.py
RemDelaporteMathurin, line 1: Can we integrate the average surface temperature with `HeatTransferProblem` class please?

RemDelaporteMathurin (changes requested):

src/festim/hydrogen_transport_problem.py
RemDelaporteMathurin: actually it seems like it's not supported yet... ignore this

src/festim/exports/surface_temperature.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Attributes:
+        temperature_field (fem.Constant or fem.Function): the temperature field
+        surface (int or festim.SurfaceSubdomain): the surface subdomain
+        filename (str): name of the file to which the surface temperature is exported
+        t (list): list of time values
+        data (list): list of average temperature values on the surface
+    """
+
+    def __init__(self, surface, filename: str = None) -> None:
```
jhdark:
```suggestion
    Args:
        surface: the surface subdomain
        filename: name of the file to which the average surface temperature is exported

    Attributes:
        temperature_field: the temperature field
        surface (int or festim.SurfaceSubdomain): the surface subdomain
        filename (str): name of the file to which the surface temperature is exported
        t (list): list of time values
        data (list): list of average temperature values on the surface
    """
    
    surface: int | SurfaceSubdomain
    filename: str | None
    
    temperature_field: fem.Constant | fem.Function


    def __init__(self, surface, filename: str = None) -> None:
```
We could use more dynamic type hinting here like whats shown here, might need to have some additional imports in the file.

src/festim/exports/surface_temperature.py
```diff
@@ -0,0 +1,51 @@
...
+
+        surface_integral = fem.assemble_scalar(
+            fem.form(temperature_field * ds(self.surface.id))
+        )  # integral over surface
+
+        surface_area = fem.assemble_scalar(fem.form(1 * ds(self.surface.id)))
+
+        self.value = surface_integral / surface_area  # avg temp
+
+        self.data.append(self.value)
```
jhdark: This might not be needed if we inherit from `AverageSurface`

jhdark (changes requested):
Just some small doc fixes from me, I think this class should probably inherit from `AverageSurface` then might be able make it simpler. 

Going back to this is making me realize we need to have a more dev team encompassing discussion on how these derived are handled in general. But this one can probably still go through for now

RemDelaporteMathurin:
This is now very stale and would deserve and completely new look. Closing until we go back to this.

PR #950, opened 2025-02-26
Add VTX Temperature Export

kaelyndunnell:
## Proposed changes

Add VTX Temperature Export for HydrogenTransportProblem Class.

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

## Further comments


src/festim/hydrogen_transport_problem.py
```diff
@@ -369,6 +370,24 @@ def initialise_exports(self):
         a string, find species object in self.species"""
 
         for export in self.exports:
+            if isinstance(export, exports.VTXTemperatureExport):
+                if isinstance(self.temperature_fenics, (fem.Function, fem.Expression)):
+                    temperature_field = self.temperature_fenics
+                else:
+                    mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 1, 1)
+                    V = dolfinx.fem.FunctionSpace(mesh, ("CG", 1))
```
RemDelaporteMathurin, line 378: Why create a unit square mesh everytime??? what if it's a 1D simulation?

You can use the mesh of the simulation directly:

```suggestion
                    V = dolfinx.fem.FunctionSpace(self.mesh, ("CG", 1))
```

src/festim/hydrogen_transport_problem.py
```diff
@@ -369,6 +370,24 @@ def initialise_exports(self):
...
+                    temperature_field = dolfinx.fem.Function(V)
+                    temperature_field.interpolate(lambda x: np.full(x.shape[1], self.temperature_fenics))
```
RemDelaporteMathurin, line 380: Why `x.shape[1]`?

src/festim/hydrogen_transport_problem.py
```diff
@@ -786,6 +805,8 @@ def update_time_dependent_values(self):
             if source.temperature_dependent:
                 source.update(t=t)
 
+    surface_temp_processed = False
```
RemDelaporteMathurin, line 808: what is this?

src/festim/hydrogen_transport_problem.py
RemDelaporteMathurin, line 380: I understand that this handles the case where `self.temperature_fenics` is a `fem.Constant`. What if this constant is updated in time? Then the newly created `fem.Function` will not be updated, correct? can we test this?

If that's the case, we probably need to add something to `post_processing` to update it in time

RemDelaporteMathurin (changes requested):
Hi Kaelyn, again very much needed. Can you update your branch with `fenicsx`?

```
git fetch upstream
git merge upstream/main
git push
```

src/festim/hydrogen_transport_problem.py
kaelyndunnell: On this note, I'm not sure.  I will say that tests done with hisp-for-iter produced temp vtx files that correctly had the time dependency

Issue #900, opened 2024-10-30
`np.max` not safe in parallel


RemDelaporteMathurin:
Note for future: this is not safe in parallel

_Originally posted by @RemDelaporteMathurin in https://github.com/festim-dev/FESTIM/pull/878#discussion_r1822516886_
            

Should be

```python
self.value = solution.function_space.mesh.comm.allreduce(val, op=MPI.MAX)
```

PR #953, opened 2025-02-26
Find max value in parallel safely

jhdark:
## Proposed changes

fix for #900 

use MPI to find max value

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/src/festim/exports/maximum_volume.py
+++ b/src/festim/exports/maximum_volume.py
@@ -30,7 +30,7 @@ class MaximumVolume(VolumeQuantity):
         solution = self.field.solution
         indices = self.volume.locate_subdomain_entities(solution.function_space.mesh)
 
-        MPI.COMM_WORLD.barrier()
+        solution.function_space.mesh.comm.barrier()
         self.value = solution.function_space.mesh.comm.allreduce(
             np.max(self.field.solution.x.array[indices]), op=MPI.MAX
         )
--- a/src/festim/exports/maximum_volume.py
+++ b/src/festim/exports/maximum_volume.py
@@ -30,7 +30,6 @@ class MaximumVolume(VolumeQuantity):
         solution = self.field.solution
         indices = self.volume.locate_subdomain_entities(solution.function_space.mesh)
 
-        solution.function_space.mesh.comm.barrier()
         self.value = solution.function_space.mesh.comm.allreduce(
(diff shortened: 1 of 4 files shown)


src/festim/exports/maximum_volume.py
```diff
@@ -27,5 +27,7 @@ def compute(self):
         """
         solution = self.field.solution
         indices = self.volume.locate_subdomain_entities(solution.function_space.mesh)
-        self.value = np.max(self.field.solution.x.array[indices])
+        max_value = max(self.field.solution.x.array[indices])
+        MPI.COMM_WORLD.barrier()
```
RemDelaporteMathurin, line 31: 1. I believe we still want to use numpy.max instead of native max here for performance purposes, no?
2. what does `barrier()` do?

RemDelaporteMathurin (commented):
I guess we'll need this for [maxinum surface](https://github.com/festim-dev/FESTIM/blob/828844cd83b02321d2dbd6a3c2d2060693931541/src/festim/exports/maximum_surface.py#L31) too

src/festim/exports/maximum_volume.py
jhdark: 1. Yes sorry, I had just copied an example I had found. Now just using numpy within the `allreduce' function instead

2. It seems the barrier function forces all processes to reach the barrier before any of them can continue execution, however for this case it may not be necessary as all the processors will be following the same logical flow. Do you have any thoughts on this @jorgensd ?

RemDelaporteMathurin: Plus shouldn't it be `solution.function_space.mesh.comm.barrier()` instead of comm world?

jhdark: I guess so

RemDelaporteMathurin (approved):
LGTM, @jorgensd do you know if the `barrier` call is required?

jorgensd:
> LGTM, @jorgensd do you know if the `barrier` call is required?

You do not need the barrier as you are using all reduce.

a barrier would be necessary if say only one process does something (or a subset), and the others have to wait for them to finish. I almost exclusively use this whenever I create meshes with gmsh, but not very often otherwise. 
It can at times be useful when profiling code as well, to ensure that timings are started in a synchronized fashion

Issue #908, opened 2024-10-31
Bug in `locate_subdomain_entities`


RemDelaporteMathurin:
https://github.com/festim-dev/FESTIM/blob/013937a860d72252436118dabcbf75b80ebc5cce/src/festim/subdomain/volume_subdomain.py#L32-L43

This will always return all the cells of the mesh, not the subdomain cells!!

We should use meshtags.find()

RemDelaporteMathurin:
I guess this works with the mixed domain approach since each subdomain has its own submesh.... this will not work with `HydrogenTransportMaterial`

RemDelaporteMathurin:
Actually running this with HydrogenTransportMaterial with several materials crashes:


```python
from mpi4py import MPI

import dolfinx
import dolfinx.fem.petsc
import numpy as np

import festim as F


def generate_mesh(n=20):

    def bottom_boundary(x):
        return np.isclose(x[1], 0.0)

    def top_boundary(x):
        return np.isclose(x[1], 1.0)

    def half(x):
        return x[1] <= 0.5 + 1e-14

    mesh = dolfinx.mesh.create_unit_square(
        MPI.COMM_WORLD, n, n, dolfinx.mesh.CellType.triangle
    )

    # Split domain in half and set an interface tag of 5
    gdim = mesh.geometry.dim
    tdim = mesh.topology.dim
    fdim = tdim - 1
    top_facets = dolfinx.mesh.locate_entities_boundary(mesh, fdim, top_boundary)
    bottom_facets = dolfinx.mesh.locate_entities_boundary(mesh, fdim, bottom_boundary)
    num_facets_local = (
        mesh.topology.index_map(fdim).size_local
        + mesh.topology.index_map(fdim).num_ghosts
    )
    facets = np.arange(num_facets_local, dtype=np.int32)
    values = np.full_like(facets, 0, dtype=np.int32)
    values[top_facets] = 1
    values[bottom_facets] = 2

    bottom_cells = dolfinx.mesh.locate_entities(mesh, tdim, half)
    num_cells_local = (
        mesh.topology.index_map(tdim).size_local
        + mesh.topology.index_map(tdim).num_ghosts
    )
    cells = np.full(num_cells_local, 4, dtype=np.int32)
    cells[bottom_cells] = 3
    ct = dolfinx.mesh.meshtags(
        mesh, tdim, np.arange(num_cells_local, dtype=np.int32), cells
    )
[...]

PR #981, opened 2025-04-18
Fix locate entities

RemDelaporteMathurin:
## Proposed changes

Fix for #908 + other bug found in the creation of D_global

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/src/festim/exports/maximum_surface.py
+++ b/src/festim/exports/maximum_surface.py
@@ -26,16 +26,17 @@ class MaximumSurface(SurfaceQuantity):
         Computes the maximum value of the field on the defined surface
         subdomain, and appends it to the data list
         """
-        solution = self.field.solution
+        solution = self.field.post_processing_solution
         entities = self.facet_meshtags.find(self.surface.id)
-        V = solution.function_space
+        if isinstance(solution, dolfinx.fem.Function):
+            V = solution.function_space
+        else:
+            V = self.field.sub_function_space
         mesh = V.mesh
-        # mesh.topology.create_connectivity(mesh.topology.dim - 1, mesh.topology.dim - 1)
+
         dofs = dolfinx.fem.locate_dofs_topological(
             V=V, entity_dim=mesh.topology.dim - 1, entities=entities
         )
 
-        self.value = solution.function_space.mesh.comm.allreduce(
-            np.max(self.field.solution.x.array[dofs]), op=MPI.MAX
-        )
+        self.value = mesh.comm.allreduce(np.max(solution.x.array[dofs]), op=MPI.MAX)
         self.data.append(self.value)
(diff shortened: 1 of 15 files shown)


src/festim/exports/maximum_surface.py
```diff
@@ -27,11 +27,15 @@ def compute(self):
...
         solution = self.field.solution
-        indices = self.surface.locate_boundary_facet_indices(
-            solution.function_space.mesh
+        entities = self.facet_meshtags.find(self.surface.id)
+        V = solution.function_space
+        mesh = V.mesh
+        # mesh.topology.create_connectivity(mesh.topology.dim - 1, mesh.topology.dim - 1)
+        dofs = dolfinx.fem.locate_dofs_topological(
+            V=V, entity_dim=mesh.topology.dim - 1, entities=entities
         )
```
jhdark, line 36: This wont work in multispecies cases as the solution is an index in this case, not an fem.function. maybe entities could be an attribute of the SurfaceSubdomain or VolumeSubdomain?

src/festim/exports/minimum_surface.py
```diff
@@ -27,11 +27,16 @@ def compute(self):
         subdomain, and appends it to the data list
         """
         solution = self.field.solution
-        indices = self.surface.locate_boundary_facet_indices(
-            solution.function_space.mesh
+
+        entities = self.facet_meshtags.find(self.surface.id)
+        V = solution.function_space
+        mesh = V.mesh
+        # mesh.topology.create_connectivity(mesh.topology.dim - 1, mesh.topology.dim - 1)
```
jhdark, line 34:
```suggestion
```

jhdark (changes requested):

src/festim/exports/maximum_surface.py
RemDelaporteMathurin: Hm I managed to make this work in a multispecies case

RemDelaporteMathurin: Ah no sorry hang on

RemDelaporteMathurin: So:

It works with multiple species using `HydrogenTransportProblem`

Derived quantities are not supported with `HydrogenTransportProblemDiscontinuous` yet anyway #980

jhdark (commented):
Could maybe add some tests for the multispecies cases?

RemDelaporteMathurin:
> Could maybe add some tests for the multispecies cases?

I did: https://github.com/festim-dev/FESTIM/blob/2387a82ac7ad56b2c83a4df5f3ef1c6eed6daebd/test/system_tests/test_misc.py#L91-L98

jhdark (approved):
will be improved in #982

Issue #988, opened 2025-04-29
Support for alternative diffusivities


jhdark:
FESTIM should be able to support different types of diffusivities and accept fem.Functions if given rather than just as floats or D_0 and E_D. 

Guess this could link in with #742

Issue #742, opened 2024-04-06
D_0 and E_D optional in F.Material


jhdark:
`D_0` and `E_D` will also need to be optional arguments, as when running HeatTransferProblem, these values won't be required.

But to prevent user error, we could incorporate tests to ensure in `HydrogenTransportProblem`s and `HeatTransferProblem`s that the necessary arguments are given in Material.

PR #989, opened 2025-04-30
Allow functions to given as diffusion coefficient

jhdark:
## Proposed changes

Fix for #988 and #742 

With this change the `D_0` and `E_D` arguments for the `festim.Material` class will be optional introducing a new parameter `D` where users can pass an `fem.Function` object instead of defining the individual components of fickian diffusion coefficient. This can allow for more complex diffusion regimes to be used as long as it is evaluated beforehand. 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/src/festim/material.py
+++ b/src/festim/material.py
@@ -86,6 +86,11 @@ class Material:
         self.solubility_law = solubility_law
         self.D = D
 
+        if self.D_0 and self.D:
+            raise ValueError(
+                "D_0 and D cannot be set at the same time. Please set only one of them."
+            )
+
     @property
     def D(self):
         return self._D
--- a/src/festim/material.py
+++ b/src/festim/material.py
@@ -213,7 +213,7 @@ class Material:
         else:
             raise TypeError("E_K_S must be either a float, int or a dict")
 
-    def get_diffusion_coefficient(self, mesh, temperature, species=None):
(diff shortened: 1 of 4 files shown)


src/festim/material.py
```diff
@@ -76,19 +84,34 @@ def __init__(
         self.heat_capacity = heat_capacity
         self.name = name
         self.solubility_law = solubility_law
+        self.D = D
```
RemDelaporteMathurin, line 87: Can we add a check that a user cannot provide both `D_0` and `D` at the same time?

test/test_material.py
```diff
@@ -198,3 +199,33 @@ def test_raises_TypeError_when_E_D_is_not_correct_type():
...
+    function"""
+
+    V = fem.functionspace(test_mesh.mesh, ("Lagrange", 1))
+    D = fem.Function(V)
+    D.x.array[:] = 2.0
+
+    my_mat = F.Material(D=D)
+    D_out = my_mat.get_diffusion_coefficient()
+
+    assert isinstance(D_out, fem.Function)
```
RemDelaporteMathurin, line 231: Could you do the following instead?

```python
assert D_out == D
```
Since in theory they should be the same object.

RemDelaporteMathurin (approved):

PR #995, opened 2025-06-03
Parrallel assemble with scifem

### LINKED ISSUE #896: `dolfinx.fem.assemble_scalar` is not parallel safe
@jorgensd realised that this is not parallel safe and would return the wrong result in parallel.

https://github.com/festim-dev/FESTIM/blob/2677c3a31c62a91b940bbb1beb1dc18d53e3d2c4/festim/exports/derived_quantities/average_volume.py#L46-L48

Would be nice to have a demonstration of this

RemDelaporteMathurin:
Fixes #896


RemDelaporteMathurin:
@jhdark can I merge this?

jhdark (approved):

Issue #991, opened 2025-05-29
Expose order of finite element


jhdark:
Right now in FESTIM2 the element order is hard coded to 1, users should be able to control this when setting up a FESTIM model.

https://github.com/festim-dev/FESTIM/blob/bedbb20a6db9df49a3d8670e02382ae04dd68f21/src/festim/hydrogen_transport_problem.py#L542

RemDelaporteMathurin:
I think what we would do here is allow this to be an optional argument of `Settings`

```python

my_model.settings = F.Settings(.... , mobile_concentrations_order=2)
```

PR #997, opened 2025-06-04
Added functionality to specify degree element order

ck768:
## Proposed changes

Users can specify the degree order when setting up a FESTIM model. This argument is optional, default is set to 1.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)
- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted
- [x] Pytest succesful

## Example usage: 

    atol=1e10,
    rtol=1e-10,
    max_iterations=30,
    final_time=50,
   element_degree=2
)

### DIFF
--- a/src/festim/hydrogen_transport_problem.py
+++ b/src/festim/hydrogen_transport_problem.py
@@ -531,7 +531,7 @@ class HydrogenTransportProblem(problem.ProblemBase):
         D.interpolate(D_expr)
         return D, D_expr
 
-    def define_function_spaces(self,element_degree=1):
+    def define_function_spaces(self, element_degree=1):
         """Creates the function space of the model, creates a mixed element if
         model is multispecies. Creates the main solution and previous solution
         function u and u_n. Create global DG function spaces of degree 0 and 1
@@ -1141,7 +1141,9 @@ class HydrogenTransportProblemDiscontinuous(HydrogenTransportProblem):
                 "initial conditions not yet implemented for discontinuous"
             )
 
-    def define_function_spaces(self, subdomain: _subdomain.VolumeSubdomain, element_degree=1):
+    def define_function_spaces(
+        self, subdomain: _subdomain.VolumeSubdomain, element_degree=1
+    ):
         """
         Creates appropriate function space and functions for a given subdomain (submesh)
         based on the number of species existing in this subdomain. Then stores the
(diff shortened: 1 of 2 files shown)


ck768:
Solves #991

RemDelaporteMathurin (changes requested):
Thanks @ck768 for this first contribution! 

Could you also propagate this change to `HydrogenTransportProblemDiscontinuous` which is the multi-material version of the hydrogen problem class?

https://github.com/festim-dev/FESTIM/blob/d308842eefcfabecc21b708e7886d98ee5d6af7d/src/festim/hydrogen_transport_problem.py#L1170-L1176

RemDelaporteMathurin:
Also, it seems like the CI is failing, are the tests passing locally?

```
python -m pytest .
```

RemDelaporteMathurin:
@ck768 I see why some tests are failing. Now whenever we call `define_function_spaces` the `.settings` attribute needs to be set because that's where we take the degree from.

Now the straightforward way of fixing that is to add a `settings` attribute in all the failing tests.

Otherwise, it may be better to instead have `element_degree` as an argument of the `define_function_spaces()` method

```python
def define_function_space(self, element_degree=1):
    ...
```

Then inside `initialise()` we simply have

```python
def initialise(self):
    ...
    self.define_function_space(element_degree=self.settings.element_degree)
```

That way you don't have to modify all the tests and it's a better design altogether

jhdark:
Yeah seems the CI is failing because in many tests, settings aren't given to the problem object. 

Just wondering if this argument should live in settings, maybe within Mesh class instead? Even though they are different, they're more closely related than this an settings, which is more for solver and transient parameters. And, in building the function spaces it already gets the `basix_cell` from the mesh, so would make it more compatible.

What do you think @RemDelaporteMathurin ?

RemDelaporteMathurin:
I would keep it in the settings instead of mesh and keep the mesh object for anything geometry related

ck768:
Hmm, that's interesting. I ran the tests locally and I had initial errors that I needed to install some Python modules, but after  I installed them, it said the tests were successful. I've been running the line
`python -m pytest .`
in my cloned and forked repo (on the fenicsx branch). Is there a better way I should test the CI?

ck768:
Figured out where the problem was while running local tests (thanks @jhdark )! I changed `define_function_spaces` to now take in the argument, and all tests seem successful now!

PR #1010, opened 2025-07-04
Fix for `MaximumSurface` and `MinimumSurface`

### LINKED ISSUE #983: [BUG] MaximumSurface requires surface_markers
**Describe the bug**

Running a FESTIM (v1.4) simulation with the `MaximumSurface` export produces the following error:
```
MaximumSurface.compute() missing 1 required positional argument: 'surface_markers'
``` 

A similar error arises with `MinimumSurface`. It seems that another if-clause should be added [here](https://github.com/festim-dev/FESTIM/blob/84905040764f87ebfdc9616779d3ed4735e2fbf0/festim/exports/derived_quantities/derived_quantities.py#L127-L143) to pass the `surface_markers`. 

**To Reproduce**

Run the MWE below:

```python
import festim as F
import numpy as np

model = F.Simulation()

model.materials = F.Material(id=1, D_0=1, E_D=0)

model.mesh = F.MeshFromVertices(np.linspace(0, 1, 10))

model.T = 1

model.settings = F.Settings(
    transient=True, absolute_tolerance=1e10, relative_tolerance=1e-10, final_time=1
)

model.dt = F.Stepsize(0.1)

model.exports = F.DerivedQuantities([F.MaximumSurface(field=0, surface=1)])

model.initialise()
model.run()

```

KulaginVladimir:
## Proposed changes

Thanks to @jorgensd, this PR should fix #983.  

I tested the fix locally for the cases of parallel computation. 

Let me know if the tests should be modified or there is a need to test the fix for `mpirun` via `pytest-mpi`.  

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/test/unit/test_exports/test_derived_quantities/test_maximum_surface.py
+++ b/test/unit/test_exports/test_derived_quantities/test_maximum_surface.py
@@ -2,6 +2,7 @@ from festim import MaximumSurface
 import fenics as f
 import numpy as np
 import pytest
+from dolfin import MPI
 
 
 @pytest.mark.parametrize("field,surface", [("solute", 1), ("T", 2)])
@@ -27,9 +28,10 @@ class TestCompute:
 
     c = f.interpolate(f.Expression("x[0]", degree=1), V)
 
-    left = f.CompiledSubDomain("x[0] < 0.5")
-    surface_markers = f.MeshFunction("size_t", mesh, 1, 1)
-    left.mark(surface_markers, 2)
+    surface_markers = f.MeshFunction("size_t", mesh, 0, 0)
+
+    right = f.CompiledSubDomain("near(x[0], 1) && on_boundary")
+    right.mark(surface_markers, 1)
 
     dx = f.Measure("dx", domain=mesh, subdomain_data=surface_markers)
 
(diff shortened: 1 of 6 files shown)


test/unit/test_exports/test_derived_quantities/test_maximum_surface.py
```diff
@@ -38,17 +40,23 @@ class TestCompute:
...
-        expected = np.max(self.c.vector().get_local()[subd_dofs])
+        local_dofs = entity_closure_dofs < len(self.c.vector().get_local())
+        if len(local_dofs) == 0:
+            local_max = -np.inf
+        else:
+            local_max = np.max(
+                self.c.vector().get_local()[entity_closure_dofs[local_dofs]]
+            )
+
+        expected = MPI.max(self.mesh.mpi_comm(), local_max)
```
RemDelaporteMathurin, line 58: Since here, the function is simply $y=x$, we know what the expected maximum is without having to compute it. It would make this test more straightforward

test/unit/test_exports/test_derived_quantities/test_minimum_surface.py
```diff
@@ -40,15 +42,20 @@ class TestCompute:
...
-        expected = np.min(self.c.vector().get_local()[subd_dofs])
+        local_dofs = entity_closure_dofs < len(self.c.vector().get_local())
+        if len(local_dofs) == 0:
+            local_min = np.inf
+        else:
+            local_min = np.min(
+                self.c.vector().get_local()[entity_closure_dofs[local_dofs]]
+            )
+
+        expected = MPI.min(self.mesh.mpi_comm(), local_min)
```
RemDelaporteMathurin, line 58: same comment as above

RemDelaporteMathurin (changes requested):
Thanks @KulaginVladimir for this fix! Just a couple of minor comments

PR #1012, opened 2025-07-13
`PointValue` in parallel

KulaginVladimir:
## Proposed changes

This PR allows using the `PointValue` export in parallel. 

In the current FESTIM version, running a simulation with `PointValue` and `mpirun` results in `RuntimeError`:
```
*** Error:   Unable to evaluate function at point.
*** Reason:  The point is not inside the domain. Consider calling "Function::set_allow_extrapolation(true)" on this Function to allow extrapolation.
*** Where:   This error was encountered inside Function.cpp.
*** Process: 0
``` 

The fix is based on the [solution](https://fenicsproject.discourse.group/t/problem-with-evaluation-at-a-point-in-parallel/1188/6) provided on the FEniCS discourse. I checked the fix on 2 machines, but yet it might need additional testing. 

The dedicated test is made with `ipyparallel` according to [this FESTIM2 test](https://github.com/festim-dev/FESTIM/blob/0a262f04039fad1160a66947b83ddd89290da223/test/test_mesh.py#L175-L186). Let me know if it should be modified.

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Black formatted

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Install dependencies
       shell: bash -l {0}
       run: |
-        pip install pytest pytest-cov
+        pip install pytest pytest-cov ipyparallel
 
     - name: Run tests
       shell: bash -l {0}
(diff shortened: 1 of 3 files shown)


KulaginVladimir:
@RemDelaporteMathurin 

I've managed to increase the coverage. Though the test now generates `.coverage.mpi_*` files when running only `pytest test/`.

Issue #1035, opened 2025-10-15
[BUG] Exporting fields in multi-species


jhdark:
**Describe the bug**

When running multi-species cases, it seems the post processing solution that was passed to the writer is not being updated?
 
So the functions are showing all 0's, however we can see from the print statement at the end that the post-processing solution is being updated. So not sure whats going on here.

**To Reproduce**
```python
import numpy as np
import festim as F

my_model = F.HydrogenTransportProblem()
my_model.mesh = F.Mesh1D(vertices=np.linspace(0, 1, num=50))

D = F.Species("D")
T = F.Species("T")
my_model.species = [D, T]

my_mat = F.Material(
    D_0={D: 3e-3, T: 6e-3},
    E_D={D: 0, T: 0},
)

vol = F.VolumeSubdomain1D(id=1, borders=[0, 1], material=my_mat)
surf_left = F.SurfaceSubdomain1D(id=1, x=0)
surf_right = F.SurfaceSubdomain1D(id=2, x=1)
my_model.subdomains = [vol, surf_left, surf_right]


my_model.temperature = 500

my_model.boundary_conditions = [
    F.FixedConcentrationBC(subdomain=surf_left, value=100, species=D),
    F.FixedConcentrationBC(subdomain=surf_left, value=150, species=T),
]

my_model.exports = [
    F.VTXSpeciesExport(filename="D.bp", field=D, subdomain=vol),
    F.VTXSpeciesExport(filename="T.bp", field=T, subdomain=vol),
]

my_model.settings = F.Settings(transient=False, atol=1e-10, rtol=1e-10)

my_model.initialise()
my_model.run()

print(D.post_processing_solution.x.array)
print(T.post_processing_solution.x.array)
```

jhdark:
Tried without the subdomain arg and produces the same result

PR #1038, opened 2025-10-20
Update post-processing solution in multispecies cases

jhdark:
## Proposed changes

fix for #1035 

At the moment trying to use `VTXSpeciesExport` is not working properly in multispecies cases as when opening the files, it will show 0 everywhere. I believe this is because the `post_processing_solution` passed to the writer attribute of the export class is not being updated each time step, instead it is being replaced. Therefore the `VTXWriter` is just writing the empty function created at the initialisation point.  

https://github.com/festim-dev/FESTIM/blob/32d138d9966e39ba436e4c4978b1a9dfcdff7425/src/festim/hydrogen_transport_problem.py#L439

This makes things tricky as I can't see good way to test if the function given to the `VTXWriter` is updating or not as it doesn't keep the function as an attribute, just going straight to making the cpp object. Because at the moment we are testing if the `post_processing_solution` is updating, which it is. @jorgensd do you have any suggestions on how this could be done better or how we can test the function given to the writer?
 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Black formatted

## Further comments

### DIFF
--- a/src/festim/hydrogen_transport_problem.py
+++ b/src/festim/hydrogen_transport_problem.py
@@ -1143,8 +1143,13 @@ class HydrogenTransportProblemDiscontinuous(HydrogenTransportProblem):
         # check that all species have a list of F.VolumeSubdomain as this is
         # different from F.HydrogenTransportProblem
         for spe in self.species:
+            if not spe.subdomains:
+                raise ValueError(
+                    f"Species {spe.name} must have a list of subdomains defined "
+                    "in 'subdomains' attribute for discontinuous problem"
+                )
             if not isinstance(spe.subdomains, list):
-                raise TypeError("subdomains attribute should be list")
+                raise TypeError("subdomains attribute in Species should be list")
 
         self.define_meshtags_and_measures()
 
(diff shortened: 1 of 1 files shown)


jorgensd:
> This makes things tricky as I can't see good way to test if the function given to the `VTXWriter` is updating or not as it doesn't keep the function as an attribute, just going straight to making the cpp object. Because at the moment we are testing if the `post_processing_solution` is updating, which it is. @jorgensd do you have any suggestions on how this could be done better or how we can test the function given to the writer?
> ##

Testing the writer is really painful, as one would need Paraview to parse the data. If/when Paraview becomes pip installable (https://www.kitware.com/paraview-6-0-and-vtk-9-5-better-together/) we can come up with a test.

jorgensd:
There is a slightly better way (I believe) to go about this @jhdark.
See the post I just made at:
https://fenicsproject.discourse.group/t/is-there-any-examples-about-using-pyvista-to-visualize-scalar-solution-in-dynamic-mixed-element-problem/18286/2?u=dokken
i.e. collapse the space, then create a function in that space (for outputting), store the map and assign data directly, then you don't get extra overhead on collapsing each time.

PR #1042, opened 2025-10-27
Only used mixed elements

jhdark:
## Proposed changes

In dolfinx v0.9.0, it became possible to have a mixed function space with a single element - something we can take advantage of, as it can simplify some logic within FESTIM. Currently, we only use a mixed function space if we have a multispecies case. 

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Black formatted

## Further comments

### DIFF
--- a/src/festim/hydrogen_transport_problem.py
+++ b/src/festim/hydrogen_transport_problem.py
@@ -924,12 +924,16 @@ class HydrogenTransportProblem(problem.ProblemBase):
             if advec_term.velocity.explicit_time_dependent:
                 advec_term.velocity.update(t=t)
 
+    def update_post_processing_solution(self):
+        """Updates the post-processing solution of each species"""
+
+        for spe in self.species:
+            spe.post_processing_solution = spe.sub_function.collapse()
+
     def post_processing(self):
         """Post processes the model"""
 
-        # update post-processing for mixed function space
-        for spe in self.species:
-            spe.post_processing_solution = spe.sub_function.collapse()
+        self.update_post_processing_solution()
 
         if self.temperature_time_dependent:
             # update global D if temperature time dependent or internal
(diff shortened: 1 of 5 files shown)


jorgensd:
> @jorgensd Do you foresee any issue with using only mixed element function spaces used like this? Just simplifies some logic for us.

I'll try to have a look this week.
In general, if not every sub space is tightly coupled (i.e. every sub components in the matrix is non-zero) mixedfunctionspace is way better for memory handling).
Here is an illustration:
```python
from mpi4py import MPI

import dolfinx.fem.petsc
import basix.ufl
import ufl

N = 10
mesh = dolfinx.mesh.create_unit_cube(MPI.COMM_WORLD, N, N, N)
el = basix.ufl.element("P", mesh.basix_cell(), degree=1)


def mix_function_space(mesh, el, num_spaces: int):
    V = dolfinx.fem.functionspace(mesh, el)
    return ufl.MixedFunctionSpace(*[V.clone()] * num_spaces)


def mixed_element(mesh, el, num_spaces: int):
    return dolfinx.fem.functionspace(mesh, basix.ufl.mixed_element([el] * num_spaces))


def block_element(mesh, el, num_spaces: int):
    return dolfinx.fem.functionspace(
        mesh, basix.ufl.element("P", mesh.basix_cell(), degree=1, shape=(num_spaces,))
    )


num_spaces = [1, 2, 4, 8, 16, 32]

nz_allocated = {mix_function_space: [], mixed_element: [], block_element: []}

for ns in num_spaces:
    for const, eb in [
        (mix_function_space, True),
        (mixed_element, False),
        (block_element, False),
    ]:
        W = const(mesh, el, ns)
        u = ufl.TrialFunctions(W)
        v = ufl.TestFunctions(W)
        a = 0
        # for i in range(ns):
        #     for j in range(ns):
        #         if i <= j:
        #             a += ufl.inner(u[i], v[j]) * ufl.dx
        a = sum(ufl.inner(u[i], v[i]) * ufl.dx for i in range(ns))
        if eb:
            a = ufl.extract_blocks(a)
        A = dolfinx.fem.petsc.assemble_matrix(dolfinx.fem.form(a))
        A.assemble()
        info = A.getInfo()
        print(ns, A.getSizes(), info["nz_allocated"])

        A.destroy()
        nz_allocated[const].append((ns, info["nz_allocated"]))
nz_to_label = {
    mix_function_space: "MixedFunctionSpace",
    mixed_element: "MixedElement",
    block_element: "BlockElement",
}
nz_to_tick = {
    mix_function_space: "o",
    mixed_element: "s",
    block_element: "^",
}

for const, nzs in nz_allocated.items():
    ns, nzs = zip(*nzs)
    plt.plot(ns, nzs, label=nz_to_label[const], marker=nz_to_tick[const])
plt.yscale("log")
plt.xlabel("Number of spaces")
plt.ylabel("Nonzeros allocated")
plt.legend()
plt.grid()
plt.savefig("nz_diag_allocated.png", dpi=400)
```
nz_diag_allocated
This above is the scenario with no coupling.
If there is say lower-diagonal coupling, you get:
nz_allocated

PR #1047, opened 2025-11-17
Fix Profile1DExport storing identical profiles by appending view inst…

### LINKED ISSUE #1046: [BUG]FESTIM v2.0b Profile1DExport stores identical profiles at different times (missing .copy(), always appending the same array view)
**Describe the bug**
```Profile1DExport``` does not correctly store 1D profiles at different times.
Even when the solution changes (verified in ParaView), the entries in ```Profile1DExport.data``` are identical for all requested times.

**To Reproduce**
```import festim as F
import numpy as np
import ufl

# simple 1D diffusion with changing profile
L = 100
vertices = np.linspace(0, L, 1000)

model = F.HydrogenTransportProblem()

H = F.Species("H")
model.species = [H]

model.mesh = F.Mesh1D(vertices)

material = F.Material(D_0=1.0, E_D=0.0)
vol = F.VolumeSubdomain1D(id=1, borders=[0, L], material=material)
left = F.SurfaceSubdomain1D(id=1, x=0)

model.subdomains = [vol, left]

model.temperature = 500

model.boundary_conditions = [F.FixedConcentrationBC(species=H, subdomain=left, value=0)]

# initial condition: step function
initial_concentration = lambda x: ufl.conditional(x[0] < 10, 1.0, 0.0)
model.initial_conditions = [
    F.InitialConcentration(value=initial_concentration, species=H, volume=vol)
]

profile_times = [1.0, 10.0]

profile = F.Profile1DExport(field=H, subdomain=vol, times=profile_times)
model.exports = [profile]

model.settings = F.Settings(atol=1e-5, rtol=1e-5, final_time=10.0)
model.settings.stepsize = F.Stepsize(
    initial_value=0.01,
    growth_factor=1.1,
    cutback_factor=0.5,
    target_nb_iterations=10,
    milestones=profile_times,
)


model.initialise()
model.run()

print("export times:", profile.t)
print("max absolute diff:", np.max(np.abs(profile.data[1] - profile.data[0])))
```
**Screenshots**

Image

hhy2022:
## Proposed changes

Fix ```Profile1DExport``` storing identical profiles by appending view instead of copy

This PR fixes a bug in ```Profile1DExport``` where profile data were stored as views of the same underlying array rather than independent copies. As a result, all exported profiles at different timesteps ended up containing identical values, which corresponding to the last computed timestep, regardless of the requested export times.

This bug was discussed in Issue #1046
Fixes #1046.

## Types of changes

- [x] New tests

## Checklist

- [x] Black formatted

## Further comments

This fix is small but crucial for all workflows relying on Profile1DExport, including:
- Transient diffusion problems
- Post-processing using stored spatial profiles
- Validation tasks requiring comparison between profiles at different times

### DIFF
--- a/test/test_profile1Dexport.py
+++ b/test/test_profile1Dexport.py
@@ -2,50 +2,50 @@ import festim as F
 import numpy as np
 import ufl
 
-# simple 1D diffusion with changing profile
-L = 100
-vertices = np.linspace(0, L, 1000)
 
-model = F.HydrogenTransportProblem()
+def test_profile1Dexport():
+    L = 100
+    vertices = np.linspace(0, L, 1000)
 
-H = F.Species("H")
-model.species = [H]
+    model = F.HydrogenTransportProblem()
 
-model.mesh = F.Mesh1D(vertices)
+    H = F.Species("H")
+    model.species = [H]
 
-material = F.Material(D_0=1.0, E_D=0.0)
-vol = F.VolumeSubdomain1D(id=1, borders=[0, L], material=material)
-left = F.SurfaceSubdomain1D(id=1, x=0)
+    model.mesh = F.Mesh1D(vertices)
 
-model.subdomains = [vol, left]
+    material = F.Material(D_0=1.0, E_D=0.0)
+    vol = F.VolumeSubdomain1D(id=1, borders=[0, L], material=material)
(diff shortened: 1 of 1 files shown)


hhy2022:
Closing since it was fixed by #1042

PR #1078, opened 2026-02-26
Various fixes in IO 

jorgensd:
Read data on grid from checkpoint, then interpolate to mesh used in simulation.
- Could be modified if we change how the checkpointing is done.

- Fix surfaceflux post-processing, it uses the wrong grid compared to the integration measure.

- Add importorskip for ipyparallel if not installed


RemDelaporteMathurin (approved):
Thank you so much for this @jorgensd 

I see now that we might have been a bit careless here. How would you do the checkpointing differently to avoid this interpolation?

Maybe we should read the mesh of the simulation from the checkpoint file too

PR #1081, opened 2026-03-02
new OutflowBC

### LINKED ISSUE #1080: Correct advection term
https://github.com/festim-dev/FESTIM/blob/b09927e62424982e6283387a558d3d95e36f6471/src/festim/hydrogen_transport_problem.py#L872-L874

Should be 

```
advection_term = -ufl.inner(vel * conc, ufl.grad(v))*self.dx(adv_term.subdomain_id)
```

jhdark:
### Summary
When including an advection term in a FESTIM simulation, an additional term may be required from integration by parts: + (v·n) u w ds.

## Type of Change

- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation


src/festim/exports/advection_flux.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Args:
+        field: species for which the advective flux is computed
+        surface: surface subdomain
+        filename: name of the file to which the advective flux is
+            exported
+        velocity_field: velocity field for which the advective flux is computed
+
+    Attributes:
+        see `festim.SurfaceFlux`
```
RemDelaporteMathurin: well there's also `velocity_field` which doesn't exist in `SurfaceFlux`

RemDelaporteMathurin (changes requested):
a few comments on my end

jhdark:
Superseded by #1231, which adds the same `+ (v·n) u w ds` term as an `OutflowBC`, arrived at from the conservative-form side, along with Soret and electromigration on the same base class. I plan to close this in favour of that PR.

Before closing I went through this PR feature by feature to check nothing gets lost.

### Covered by #1231

- **Conservative advection form.** Every drift term there is assembled in divergence form (`drift.drift_form`), so the `v·grad(c)` → `div(c v)` change here is superseded rather than dropped.
- **`OutflowBC`.** Same boundary term, same file (`boundary_conditions/outflow_bc.py`), wired into both the continuous and the discontinuous problem paths, and additionally to the boundary of a codim-1 manifold.
- **Advective contribution to the reported flux.** `SurfaceFlux` now returns the total `-D grad(c)·n + c v·n`, so removing the "advection terms are not currently accounted for" warning is correct there for the same reason it was here.
- **Mass-balance testing.** `test_flux_conservation_advection` here is covered by `test_every_atom_that_enters_leaves` (counterflow), `test_inventory_is_conserved_under_drift`, `test_surface_flux_accounts_for_drift`, and the two `test_outflow_bc_*` tests, plus an MMS case with a non-solenoidal velocity.

### Not carried over, worth porting

- **The same-mesh short-circuit in `VelocityField.convert_input_value`.** This PR uses the given velocity function directly when it already lives on the problem's mesh (`vel.function_space.mesh == function_space.mesh`) instead of interpolating it. #1231 always interpolates into a CG1 vector space. This is the only thing here with no equivalent there, and it is the same concern as my DG0 comment on that PR, so I will follow it up there rather than keep this one open for it.

### Deliberate differences

- **`AdvectionFlux` export dropped**, with the advective part folded into `SurfaceFlux` instead. One consequence to be aware of: with drift terms present there is then no export that reports the *diffusive-only* flux, since `SurfaceFlux` always includes drift once the problem sets its `drift_velocity`. Raising that on #1231.
- **`OutflowBC` takes a single species** there rather than a list, which is consistent with `ParticleFluxBC` and answers @RemDelaporteMathurin's comment above about the inconsistency.
- **The velocity is taken from the drift terms** acting on that species in the adjacent volume rather than passed to the BC, so the BC is a no-op on a surface where no drift acts. This is the better interface: the problem already knows the velocity, and a time-dependent one is updated through the drift term.

PR #1131, opened 2026-04-29
Landing page makeover

RemDelaporteMathurin:
## Type of Change

- [x] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation


KulaginVladimir:
@RemDelaporteMathurin, Looks awesome!

Some optional suggestions:
1. In the carousel with FESTIM examples, _Permeation through a multi-layer membrane_ does not look as visually impressive or as easy to understand as, for example, the monoblock simulations. The _Tritium permeator probe_ example also seems a bit overcomplicated, with many small subfigures. The image could perhaps be simplified for the gallery.
2. I think the icons are too small. For example, the Python icon is barely distinguishable, although their size of course depends on whether they need to fit in a single row.
3. In dark mode, the text in the panels lacks contrast with the background, but it looks fine on hover.
image

RemDelaporteMathurin:
> @RemDelaporteMathurin, Looks awesome!
> 
> Some optional suggestions:
> 
> 1. In the carousel with FESTIM examples, _Permeation through a multi-layer membrane_ does not look as visually impressive or as easy to understand as, for example, the monoblock simulations. The _Tritium permeator probe_ example also seems a bit overcomplicated, with many small subfigures. The image could perhaps be simplified for the gallery.
> 2. I think the icons are too small. For example, the Python icon is barely distinguishable, although their size of course depends on whether they need to fit in a single row.
> 3. In dark mode, the text in the panels lacks contrast with the background, but it looks fine on hover.
> 
> image

Thanks @KulaginVladimir I've updated the page based on your comments

hhy2022:
Thanks for the great work on this! The page is looking really clean overall. A few small thoughts and observations that might be worth considering:

1. Dark/light mode toggle
Switching to dark mode seems to take two clicks, while switching back to light mode only takes one. Not sure if this is intentional, it might feel a bit more consistent if both behaved the same way?

2. Institution logos
The logos are grayscale by default and turn to color on hover, which is a really nice interaction. That said, defaulting to grayscale might feel a little less vibrant than color. Maybe default to color and use grayscale as the hover state?

3. Hero visualization
The content seems centered on the page, leaving quite a bit of empty space on the left and right sides. This becomes a bit more noticeable on larger screens.

4. BibTeX copy button
Nice idea to have the BibTeX in a collapsible block. A one-click "Copy" button could be a small but useful addition for researchers.

5. "Trusted by researchers worldwide"
Maybe we could also add something like "Used by 30+ institutions across 15 countries"?

6. Image caption
The caption under the hero visualization is fairly small and low-contrast, so it's easy to miss. Slightly larger font size or higher contrast could help.

RemDelaporteMathurin:
> 1. Dark/light mode toggle
>    Switching to dark mode seems to take two clicks, while switching back to light mode only takes one. Not sure if this is intentional, it might feel a bit more consistent if both behaved the same way?

This is because there are three modes. Dark, Light, and "system" which matches the operating system setting.



> 2\. Institution logos
>     The logos are grayscale by default and turn to color on hover, which is a really nice interaction. That said, defaulting to grayscale might feel a little less vibrant than color. Maybe default to color and use grayscale as the hover state?

This was deliberate as having too many different colours is a bit much in my opinion

> 3\. Hero visualization
>     The content seems centered on the page, leaving quite a bit of empty space on the left and right sides. This becomes a bit more noticeable on larger screens.

I'm afraid that's a pydata scheme setting I haven't been able to play with too much

> 4\. BibTeX copy button
>     Nice idea to have the BibTeX in a collapsible block. A one-click "Copy" button could be a small but useful addition for researchers.

good shout, I'll see what I can do

> 5\. "Trusted by researchers worldwide"
>     Maybe we could also add something like "Used by 30+ institutions across 15 countries"?

Yes we could and that would put our impact into perspective, although that requires maintenance and update which I'm not too keen on for this. Happy to hear others' thoughts

> 6\. Image caption
>     The caption under the hero visualization is fairly small and low-contrast, so it's easy to miss. Slightly larger font size or higher contrast could help.

While this is supposed to be second order elements, I can try and make it pop a bit more if others agree

RemDelaporteMathurin:
@hhy2022 the last commit added the `.. code-bock::` RST directive which actually has a lot more features (like syntax colouring) including a copy button. Thanks for the suggestion! 

I didn't know we could easily mix `.. raw:: html` and other directives like that!

image

PR #1138, opened 2026-04-30
Custom derived quantity

### LINKED ISSUE #1130: Custom derived quantities
It would be nice for users to be able to define custom derived quantities like this:

```python
def my_func(**kwargs):
    D = kwargs["D"]
    c_h1 = kwargs["H_1"]
    c_h2 = kwargs["H_2"]
    n = kwargs["n"]

    return -D * ufl.dot(ufl.grad(c_h1 + c_h2), n)


def retention(**kwargs):
    D = kwargs["D"]["H_1"]
    D_0 = kwargs["D_0"]
    c_h1 = kwargs["H_1"]
    c_h2 = kwargs["H_2"]
    n = kwargs["n"]

    return c_h1 + c_h2


my_custom_quantity = F.CustomQuantity(expr=retention, subdomain=vol1, title="Retention", filename="retention.txt")
```

RemDelaporteMathurin:
### Summary
This PR adds `festim.exports.CustomQuantity` which allows users to compute any derived quantity on a volume or surface subdomain.

Other changes:
- `VolumeQuantity` and `SurfaceQuantity` now both inherit from `DerivedQuantity` (all three are abstract classes)
- This allows to deduplicate code like `write()` and `filename()`
- added facet_meshtags and volume mesthtags attributes to `MaximumVolume` (and friends)
- I had to modify the way we do the "D_global" thing in the continuous problem to make sure that `kwargs["D"]` was updated when the temperature was time dependent

#### Usage
```python
import numpy as np
import festim as F

material = F.Material(D_0=1, E_D=0)
volume = F.VolumeSubdomain(id=1, material=material)
surface = F.SurfaceSubdomain(id=1, locator=lambda x: np.isclose(x[1], 1))


# cA + cB
def total_concentration(**kwargs):
    return kwargs["A"] + kwargs["B"]

quantity = F.CustomQuantity(
    expr=total_concentration,
    subdomain=volume,
    title="Total quantity",
)

# -D grad(c_A) . n
surface_quantity = F.CustomQuantity(
    expr=lambda **kwargs: -kwargs["D_A"] * ufl.dot(
        ufl.grad(kwargs["A"]), kwargs["n"]
    ),
    subdomain=surface,
    title="Surface flux",
)
```

### Related Issues
Fixes #1130 

## Type of Change

- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation


src/festim/exports/custom_quantity.py
```diff
@@ -0,0 +1,131 @@
...
+        Q = \int_{\Gamma} q\,\mathrm{d}\Gamma
+
+    over the selected surface subdomain.
+
+    The expression returned by ``expr`` is treated as an integrand and assembled over
+    the chosen subdomain.
+
+    .. math::
+
+        Q = \int_{\Omega} q\,\mathrm{d}\Omega
```
jhdark, line 94: These math blocks render on the docs page, but not with the VSC hover/tooltip function. Maybe this is fine? Otherwise, we could just have the equations in Unicode?

src/festim/exports/custom_quantity.py
```diff
@@ -0,0 +1,131 @@
...
+
+    where ``q`` is the UFL expression returned by ``expr`` and ``\Omega`` is either a
+    surface or a volume subdomain.
+    """
+
+    def __init__(
+        self,
+        expr: Callable,
+        subdomain: SurfaceSubdomain | VolumeSubdomain,
+        title: str = "Custom Quantity",
```
jhdark, line 104: Maybe we could make this a required arg instead? happy either way

src/festim/exports/custom_quantity.py
```diff
@@ -0,0 +1,131 @@
...
+    ) -> None:
+        super().__init__(filename=filename)
+        self.expr = expr
+        self.subdomain = subdomain
+        self._title = title
+        self.ufl_expr = None
+
+    @property
+    def title(self):
+        return self._title
```
jhdark, line 115: Is it worth doing any type checking here, not sure if it would break things down the line if people gave anything other than a string?

jhdark: Actually looking down the line, the value is put into an fstring so wont be a problem

jhdark (approved):
Looks good to me, is it worth adding some testing too to make sure things break when trying to define an volume/surface quantity without giving a title/compute functions? Although this isnt FESTIM functionality just the python abstract class.

src/festim/hydrogen_transport_problem.py
```diff
@@ -493,38 +498,52 @@ def initialise_exports(self):
...
+                # NOTE we need to change our D_global approach
+                D_kwargs = {
+                    f"D_{sp.name}": self._species_to_D_global[sp] for sp in self.species
+                }
+                kwargs.update(D_kwargs)
+                kwargs["D"] = {sp.name: D_kwargs[f"D_{sp.name}"] for sp in self.species}
+                if len(self.species) == 1:
+                    kwargs["D"] = kwargs[f"D_{self.species[0].name}"]
+                kwargs["x"] = ufl.SpatialCoordinate(self.mesh.mesh)
+                export.ufl_expr = export.expr(**kwargs)
```
jhdark, line 547: We could add in time dependence here too

src/festim/exports/custom_quantity.py
RemDelaporteMathurin: I think we have this in other places too. I didn't know they don't render on the VSC (pylance?) tooltip. That can be a separate issue that we fix (or not) globally

src/festim/exports/custom_quantity.py
RemDelaporteMathurin: I personally never use titles for derived quantities (even the ones given by default), don't know about the others. That's also because I pretty much never write derived quantities to csv. 

For users using derived quantities like me, i would find it a bit constraining to have to add a title even if you never use it. The only downside of having this default title is that if a user exports two custom quantities to a csv then they would both be written as "Custom Quantity" in the header.

Two solutions:

- raise a warning if two quantities have the same name (i'd do it in another PR)
- dynamically add an index to the title if there are several (ie. "Custom Quantity_1", "Custom Quantity_2")

What do you think?

jhdark: Tbf, you can give different filenames per quantity too, so maybe not an issue, think its an issue you'd run into once and then change. Dynamic naming would be better, but only if this rare scenario happens, and it would add more logic. Let's just leave it and see what issues we run into, if any

PR #1158, opened 2026-05-18
Weakly enforced Dirichlet BCs

### LINKED ISSUE #1153: Option to impose Dirichlet BC's weakly
We could allow users to define the Dirichlet BC weakly, but they would need to also provide a penalty term for this as it would use Niche method

RemDelaporteMathurin:
### Summary
This PR adds the possibility to weakly enforce dirichlet BCs.

### Related Issues
Fixes #1153 

### Motivation and Context
Sometimes, it is necessary to weakly impose a Dirichlet BC (eg. flux conservation). So far FESTIM only allowed strongly enforced BCs but it wasn't too complicated to weakly enforce it in the variational formulation thanks to @jorgensd tutorials.

## Type of Change

- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation

### DIFF
--- a/.github/workflows/ci_docker.yml
+++ b/.github/workflows/ci_docker.yml
@@ -20,8 +20,7 @@ jobs:
 
       - name: Install local package and dependencies
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install .[test] --no-build-isolation
+          python -m pip install .[test]
 
       - name: Overload adios4dolfinx
         if: ${{ matrix.container_version == 'nightly' }}
--- a/.github/workflows/ci_docker.yml
+++ b/.github/workflows/ci_docker.yml
@@ -20,7 +20,7 @@ jobs:
 
       - name: Install local package and dependencies
         run: |
-          pip install .[test] --no-build-isolation
+          python -m pip install .[test] --no-build-isolation
 
       - name: Overload adios4dolfinx
         if: ${{ matrix.container_version == 'nightly' }}
--- a/.github/workflows/ci_docker.yml
+++ b/.github/workflows/ci_docker.yml
@@ -20,7 +20,7 @@ jobs:
 
       - name: Install local package and dependencies
         run: |
-          pip install .[test]
+          pip install .[test] --no-build-isolation
 
       - name: Overload adios4dolfinx
         if: ${{ matrix.container_version == 'nightly' }}
--- a/.github/workflows/ci_docker.yml
+++ b/.github/workflows/ci_docker.yml
@@ -20,6 +20,7 @@ jobs:
 
       - name: Install local package and dependencies
         run: |
+          python -m pip install --upgrade pip
           python -m pip install .[test] --no-build-isolation
 
       - name: Overload adios4dolfinx
--- a/.github/workflows/ci_docker.yml
+++ b/.github/workflows/ci_docker.yml
@@ -20,6 +20,7 @@ jobs:
 
       - name: Install local package and dependencies
         run: |
+          python -m pip install scipy  # needed in scifem, can be removed when scifem adds it to their dependencies
           python -m pip install .[test]
 
       - name: Overload adios4dolfinx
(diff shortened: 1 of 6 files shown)


.github/workflows/ci_docker.yml
```diff
@@ -20,7 +20,8 @@ jobs:
 
       - name: Install local package and dependencies
         run: |
-          pip install .[test]
+          python -m pip install --upgrade pip
```
jhdark: Whats this for?

src/festim/boundary_conditions/dirichlet_bc.py
```diff
@@ -138,6 +144,47 @@ def update(self, t: float):
...
+        mesh = ds.ufl_domain()
+        n = ufl.FacetNormal(mesh)
+        h = ufl.Circumradius(mesh)  # FIXME this doesn't work for rectangles
+        alpha = self.penalty
+        assert alpha is not None, (
+            "Penalty parameter must be given for weakly enforced Dirichlet BCs"
+        )
+        assert self.value_fenics is not None, (
+            "value_fenics must be defined for weakly enforced Dirichlet BCs"
+        )
```
jhdark, line 183: Should this not be done in a setter instead? Only if the `enforce_weakly` is True?

src/festim/boundary_conditions/dirichlet_bc.py
```diff
@@ -138,6 +144,47 @@ def update(self, t: float):
...
+                is applied
+            v: the test function
+            ds: the surface measure
+
+        Returns:
+            the weak formulation
+        """
+        mesh = ds.ufl_domain()
+        n = ufl.FacetNormal(mesh)
+        h = ufl.Circumradius(mesh)  # FIXME this doesn't work for rectangles
```
jhdark, line 176: Will this break for hex meshes or only give the wrong value? Because if it breaks, maybe we could have a mini test in here and give a more descriptive error?

src/festim/boundary_conditions/dirichlet_bc.py
jhdark, line 51: Should add these new parameters to the doc strings

RemDelaporteMathurin:
@jorgensd @jhdark The Docker CI for dolfinx 0.9 doesn't go pas the install and I'm confused as to why. It seems related to some pip PEP8 rules about the license in scifem being `"MIT"` instead of `{ text = "MIT" }` ???? but for some reason that's not required in the 0.10.... I'm keen on ignoring this for now but I thought I'd bring it up with you

jhdark (changes requested):
Looks great, that was a quick turn around, are we going to add this to the discontinuous class too?

.github/workflows/ci_docker.yml
jhdark: Oh never mind, saw the final comment on the PR

RemDelaporteMathurin: I'll revert this cause i don't think it's related to this. See this PR https://github.com/scientificcomputing/scifem/pull/213

src/festim/boundary_conditions/dirichlet_bc.py
RemDelaporteMathurin: It will break. But because there is an open issue for this elswhere I'm hoping we can fix this pretty soon anyway

src/festim/boundary_conditions/dirichlet_bc.py
RemDelaporteMathurin: No not in a setter because then you always have to set say `enforce_weakly` before `alpha` or the opposite because there is no way to set them both at the same time.

This method is only called if `enforce_weakly` is True anyway.

RemDelaporteMathurin (commented):
> are we going to add this to the discontinuous class too?

It has already been added to the discontinuous case.

jorgensd:
> @jorgensd @jhdark The Docker CI for dolfinx 0.9 doesn't go pas the install and I'm confused as to why. It seems related to some pip PEP8 rules about the license in scifem being `"MIT"` instead of `{ text = "MIT" }` ???? but for some reason that's not required in the 0.10.... I'm keen on ignoring this for now but I thought I'd bring it up with you

Try setting a bound on setuptools, like: https://github.com/FEniCS/ufl/blob/main/pyproject.toml#L3

RemDelaporteMathurin:
@jhdark all the tests are now passing after a few fixes to the Docker CI, if you're happy with the changes can we merge this?

PR #1169, opened 2026-06-02
Interface condition diagnostic

### LINKED ISSUE #1137: Tools to diagnose interface condition enforcement
As a user, I would like to be able to export, on an interface (ie. codim 1 submesh), the following quantities:

$c_1/K_{S_1}$, $c_2/K_{S_2}$, and the residual  $c_1/K_{S_1} - c_2/K_{S_2}$ 
(and equivalent for Sievert/Henry interfaces)

These space-dependent quantities would then be exported to VTX for visualisation. 

Alternatively, we should also be able to export the integral of the residual as a derived quantity: $\int_\Gamma \left( c_1/K_{S_1} - c_2/K_{S_2} \right) dS$

Steps to achieve this:

- make a codim 1 submesh of the interface
- make a function space (CG1?)
- create `dolfinx.fem.Function` object each quantity
- create `dolfinx.fem.Expression` (⚠️ can this work if https://github.com/FEniCS/dolfinx/issues/3207 is not resolved?)
- interpolate expression in appropriate `Function` object in the iteration loop + write to file

jhdark:
## Description

Fix for issue #1180 

### Summary
This PR adds an additional export which evaluates the residual of the interface condition, which should confirm whether the penalty term used is high enough to enforce the condition properly.

Here is a small MWE which plots the max value of the residual against the penalty term used in both Sievert-Sieverts and Sieverts-Henry cases:

```python
from mpi4py import MPI

import dolfinx
import numpy as np

import festim as F


def sim(penalty, mat_r_law):
    my_model = F.HydrogenTransportProblemDiscontinuous()

    msh = dolfinx.mesh.create_unit_square(
        MPI.COMM_WORLD,
        10,
        10,
        cell_type=dolfinx.mesh.CellType.triangle,
        diagonal=dolfinx.mesh.DiagonalType.crossed,
    )

    my_mesh = F.Mesh(msh)
    my_model.mesh = my_mesh

    mat_L = F.Material(D_0=1e-03, E_D=0.0, K_S_0=1, E_K_S=0.0, solubility_law="sievert")
    mat_R = F.Material(D_0=1e-03, E_D=0.0, K_S_0=2, E_K_S=0.0, solubility_law=mat_r_law)
    vol_L = F.VolumeSubdomain(id=1, material=mat_L, locator=lambda x: x[0] <= 0.5)
    vol_R = F.VolumeSubdomain(id=2, material=mat_R, locator=lambda x: x[0] >= 0.5)
    left = F.SurfaceSubdomain(id=3, locator=lambda x: np.isclose(x[0], 0.0))
    right = F.SurfaceSubdomain(id=4, locator=lambda x: np.isclose(x[0], 1.0))
    my_model.subdomains = [vol_L, vol_R, left, right]

    H = F.Species("H", subdomains=[vol_L, vol_R])
    my_model.species = [H]

    center_interface = F.Interface(
        id=5, subdomains=[vol_L, vol_R], penalty_term=penalty
    )
    my_model.interfaces = [center_interface]

    my_model.temperature = 100

        F.FixedConcentrationBC(subdomain=left, value=1.0, species=H),
        F.FixedConcentrationBC(subdomain=right, value=0.0, species=H),
    ]

    my_model.settings = F.Settings(atol=1e-10, rtol=1e-10, transient=False)

    my_model.exports = [
        F.VTXSpeciesExport(field=H, filename="mwe_L.bp", subdomain=vol_L),
        F.VTXSpeciesExport(field=H, filename="mwe_R.bp", subdomain=vol_R),
        F.VTXInterfaceResidualExport(
            field=H, filename="mwe_interface.bp", interface=center_interface
        ),
    ]

    my_model.initialise()
    my_model.run()

    max_residual = np.max(np.absolute(my_model.exports[2].function.x.array[:]))

    return max_residual


if __name__ == "__main__":
    penalty_terms = np.geomspace(1e-02, 1e05, num=10)
    mat_laws = ["sievert", "henry"]
    max_values_ss = []
    max_values_sh = []

    for penalty in penalty_terms:
        max_residual = sim(penalty=penalty, mat_r_law="sievert")
        print(
            f"Sievert-Sievert, Penalty term: {penalty:.2e}, Max interface residual: {max_residual:.2e}"
        )
        max_values_ss.append(max_residual)

    for penalty in penalty_terms:
        max_residual = sim(penalty=penalty, mat_r_law="henry")
        print(
            f"Sievert-Henry, Penalty term: {penalty:.2e}, Max interface residual: {max_residual:.2e}"
        )
        max_values_sh.append(max_residual)

    plt.figure()
    plt.plot(penalty_terms, max_values_ss, marker="o", label="Sievert-Sievert")
    plt.plot(penalty_terms, max_values_sh, marker="o", label="Sievert-Henry")
    plt.xscale("log")
    plt.yscale("log")
    plt.xlabel("Penalty term")
    plt.ylabel("Max interface residual")
    ax = plt.gca()
    ax.spines["top"].set_visible(False)
    ax.spines["right"].set_visible(False)
    plt.legend()

    plt.show()

```
image

### Per-timestep logging

The export also reports a summary of the residual to the logger, one line per
interface per timestep. It only appears when **both** are true: a
`VTXInterfaceResidualExport` has been defined, and the log level is `INFO` or
below.

```python
import dolfinx

dolfinx.log.set_log_level(dolfinx.log.LogLevel.INFO)
```

```
[2026-08-13 08:51:36.223] [info] Interface 5 residual (H) t=1.00000e-01 ; max=6.51530e-01 ; avg=6.48719e-01 ; max_relative=9.61151e-03
[2026-08-13 08:51:36.227] [info] Interface 5 residual (H) t=2.00000e-01 ; max=7.76615e-01 ; avg=7.76109e-01 ; max_relative=9.37504e-03
[2026-08-13 08:51:36.230] [info] Interface 5 residual (H) t=3.00000e-01 ; max=7.92985e-01 ; avg=7.92920e-01 ; max_relative=9.27045e-03
```

The same case at `penalty_term=1e4` instead of `1e1`:

```
[info] Interface 5 residual (H) t=3.00000e-01 ; max=7.92962e-04 ; avg=7.92897e-04 ; max_relative=9.27019e-06
```

A penalty term 1000x larger gives a residual 1000x smaller, which is the `1/penalty`
scaling of the penalty method showing up directly in the diagnostic.

`max_relative` divides the largest mismatch by the largest magnitude of the two side
terms. It is the number worth watching, because the absolute residual is dimensional:
`max=8e-2` looks alarming on its own, but at `max_relative=1e-4` the interface
condition is being met perfectly well. It is a single global ratio rather than the
largest pointwise ratio, since the pointwise one is unbounded wherever both sides
approach zero.

The statistics are also available programmatically via
`export.compute_statistics()`, which returns `max`, `mean` and `max_relative`.
Nothing is computed when the log level is above `INFO`, so there is no cost (and no
MPI reduction) for users who do not ask for it.

### `filename` is optional

If all you want is the logged summary, there is no need to write a file:

```python
# residual reported to the logger only, nothing written to disk
F.VTXInterfaceResidualExport(field=H, interface=my_interface)

# residual also written for ParaView, with f_0 and f_1 alongside it
F.VTXInterfaceResidualExport(
    field=H, interface=my_interface, filename="interface_residual.bp"
)
```

`filename` now defaults to `None` and follows `interface` in the signature, so the
two required arguments come first.

### 1D

Supported. The interface of a 1D mesh is a single point, so the residual is one
value rather than a field, and its function space is `DG0` (a point admits no
order-1 element). Everything else, including the statistics and the log line, works
the same. In 1D `max` and `mean` are necessarily the same number, and the `.bp` file
holds a single point, which is where leaving `filename` unset is most useful.

### Also in this PR

Three things that came out of writing the diagnostic:

- **`F.Interface(..., method=...)` now takes effect.** `method_interface` is a
  property on `HydrogenTransportProblemDiscontinuous`, so `hasattr(self, "method_interface")`
  in `initialise()` was always `True`, and every interface had its method overwritten
  with the class default. Selecting Nitsche per interface was silently ignored, and the
  deprecated problem-level attribute was the only way that worked. It also emitted a
  `DeprecationWarning` on every run of a discontinuous problem, even when the user had
  never touched the attribute.
- **The Nitsche method now raises on a mixed Henry/Sievert interface** instead of
  silently solving a different problem. Its jump is `u_0/K_0 - u_1/K_1`, which is only
  the interface condition when both sides share a solubility law. On a mixed interface
  the solve converges cleanly to a solution whose error does not decrease under
  refinement, with nothing to indicate anything is wrong. Use the penalty method for
  those, which is what already supported them.
- **Fix to the residual itself:** a uniform temperature combined with a non-zero
  `E_K_S` raised `Found multiple domains, cannot return just one`. A `fem.Constant`
  carries the mesh it was built on, and it cannot appear in an expression alongside the
  interface-submesh concentrations. It went unnoticed because with `E_K_S == 0` the
  temperature cancels out of `K_S` and the clash never surfaces.

### Known limitations

- Only wired into `HydrogenTransportProblemDiscontinuous`. Adding the export to
  `HydrogenTransportProblem` raises a `NotImplementedError` naming the alternative.
- The residual is computed from the converged solution at each timestep, not during
  the Newton iterations, so it says whether the penalty term is adequate, not whether
  it is hurting convergence.

## Type of Change

- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation

## Breaking Changes

Both follow from `F.Interface(..., method=...)` now actually being honoured, and only
affect users who were already relying on behaviour that did not do what it said.

1. A user who wrote `F.Interface(..., method="nitsche")` was silently getting the
   penalty method. They now get Nitsche, which is what they asked for, but it is a
   change in what their script solves.
2. If that same user also has a mixed Henry/Sievert interface, they now get a
   `NotImplementedError` rather than a silently wrong answer. The fix is to select
   `InterfaceMethod.penalty` explicitly, which is what they were effectively running
   before.


jhdark:
@RemDelaporteMathurin So far, this only interpolates the values of the residual onto a function which lives on the interface, but maybe we could also add something additional to the logger, like obtaining the max absolute value from the function? Let me know what you think

src/festim/exports/vtx.py
```diff
@@ -345,6 +348,180 @@ def check_valid_inputs(self, kwargs: dict):
...
+
+        K_S_0 = self._K_S_0 * np.exp(-self._E_K_S_0 / (_k_B * T))
+        K_S_1 = self._K_S_1 * np.exp(-self._E_K_S_1 / (_k_B * T))
+        left = interface_condition_term(
+            self._u_0.x.array, K_S_0, self._law_0, self._law_1
+        )
+        right = interface_condition_term(
+            self._u_1.x.array, K_S_1, self._law_1, self._law_0
+        )
+        self.function.x.array[:] = right - left
```
RemDelaporteMathurin: i think using `.x.array` for this might be risky since it assumes they ordered wrt dofs in the exact same way, which might not be True.

the proper way to do it is to make a `fem.Expression` object with `u_0/K_s - u_1/K_s` and then `interpolate` it in `self.function`

RemDelaporteMathurin (commented):
- Logging: yes!
- We also need to be able to export not only the residual c/s - c/s but c/s itself

RemDelaporteMathurin:
@jhdark tests seem to fail here

src/festim/exports/vtx.py
```diff
@@ -464,62 +526,104 @@ def initialise(self, temperature_fenics: fem.Constant | fem.Function) -> None:
...
-        u_1 = self.field.subdomain_to_post_processing_solution[subdomain_1]
 
-        self._u_0.interpolate_nonmatching(
-            u_0, self._interface_cells, interpolation_data=self._interp_data_0
+        # Get the concentration from each subdomain on the interface submesh
+        c_0 = self.field.subdomain_to_post_processing_solution[subdomain_0]
+        c_1 = self.field.subdomain_to_post_processing_solution[subdomain_1]
+
+        # FIXME: once we support dolfinx 0.11 we should be able to mix parent and
+        # sub-meshes in the same expression
```
RemDelaporteMathurin: we do support dolfixn 0.11 now!

jhdark: Ok when we start using this feature in 0.11 we could re-do this export

src/festim/hydrogen_transport_problem.py
```diff
@@ -1857,7 +2224,8 @@ def post_processing(self):
                 elif isinstance(export, exports.VTXTemperatureExport):
                     export.writer.write(float(self.t))
                 elif isinstance(export, exports.VTXInterfaceResidualExport):
-                    export.write(float(self.t))
+                    export.update()
```
RemDelaporteMathurin, line 2227: can we add a note here that we wouldn't need to update if we can (I think we can already) mix parent functions in submesh expressions

RemDelaporteMathurin (commented):
Two small comments regarding mixing parent functions/constants in submesh expression now available in dolfinx 0.11

src/festim/exports/vtx.py
RemDelaporteMathurin: is it worth doing it now since we already can? it should simplify the implementation since we don't need either updating or nmm interpolation

jhdark:
@RemDelaporteMathurin Regarding the mixing parent functions in submesh expressions, should we look into this for this PR or simply change the FIXME to be more specific on what would need to be there to replace what's currently implemented?

RemDelaporteMathurin:
> @RemDelaporteMathurin Regarding the mixing parent functions in submesh expressions, should we look into this for this PR or simply change the FIXME to be more specific on what would need to be there to replace what's currently implemented?

i think if it's not too much we should take care of it now

src/festim/exports/vtx.py
```diff
@@ -344,6 +356,371 @@ def check_valid_inputs(self, kwargs: dict):
...
+        self._f_1_interface.interpolate(self.f_1_expr)
+
+    def write(self, t: float) -> None:
+        """Write the current residual to the VTX file, if one was requested.
+
+        Args:
+            t: the current time
+        """
+        if self.writer is not None:
+            self.writer.write(t)
```
RemDelaporteMathurin, line 656: Noticing that this silently fails if `self.writer is None`. 
I would prefer if it failed loudly:


```suggestion
        assert self.writer, "writer attribute needs to be set"
        self.writer.write(t)
```

Issue #1123, opened 2026-04-23
Support for hydride reactions


RemDelaporteMathurin:
We should be able to write custom `Reaction` classes to support hydride processes (nucleation, growth, dissociation).

Work by @ck768 has already begun here: https://github.com/ck768/dolfinx-hydride

RemDelaporteMathurin:
Progress has been made by @ck768 here: https://github.com/ck768/dolfinx-hydride/tree/main/festim

PR #1181, opened 2026-07-09
Reaction re-work

jhdark:
### Summary

This PR reworks `Reaction` into a small class hierarchy so that FESTIM can express
reactions other than Arrhenius trapping/detrapping, and changes how reactions enter
the variational formulation.

**New classes** (all exported in `festim/__init__.py`):

| Class | Net rate |
|---|---|
| `ReactionBase` | arbitrary $R = f(c_i, x, T, t)$ given by a single `reaction_rate` |
| `GenericReaction(ReactionBase)` | mass action, $R = k_1 \prod_i c_i^\text{reactant} - k_2 \prod_j c_j^\text{product}$ |
| `ArrheniusReaction(GenericReaction)` | $k_1, k_2$ built from `k_0`/`E_k` and `p_0`/`E_p` (the old `Reaction`) |
| `DecayReaction(GenericReaction)` | first-order radioactive decay, $R = \lambda  c_i$ |

`Reaction` is kept as a thin deprecated alias of `ArrheniusReaction` emitting a
`DeprecationWarning`.

**Reactions are now expanded into particle sources.** Instead of each problem class
writing reaction terms into the formulation itself, a reaction is unpacked into one
`ParticleSource` per appearance of a species in `reactant`/`product`
(`Reaction.create_sources()`): a sink $-R$ for each reactant, a source $+R$ for each
product. The problem classes gained `convert_reaction_rates_to_fenics_objects()` and
`create_sources_from_reactions()`, which fill `_unpacked_sources` (user sources +
reaction sources), and the existing source loop does the rest. This removes the
duplicated reaction-term code from `HydrogenTransportProblem` and
`HydrogenTransportProblemDiscontinuous`.

**Rate coefficients are now `festim.Value` objects**, so a rate may be a float, a UFL
expression, a `fem.Constant`/`fem.Function`, or a callable of temperature (`T`), time
(`t`), position (`x`) and/or other species concentrations (through `arg_to_species`).
Explicitly time-dependent rates are updated every timestep in
`update_time_dependent_values`.

## Type of Change

- [x] 🔨 Code refactoring (no functional changes, no API changes)
- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation

### `ArrheniusReaction` — unchanged behaviour, new name

```python
import festim as F

mobile = F.Species("mobile")
trapped = F.Species("trapped", mobile=False)
empty_sites = F.ImplicitSpecies(n=1e19, others=[trapped])

my_model.reactions = [
    F.ArrheniusReaction(
        reactant=[mobile, empty_sites],
        product=trapped,
        k_0=1e-7,
        E_k=0.2,
        p_0=1e13,
        E_p=1.2,
        volume=vol,
    )
]
```

$$R = k_0 \, e^{-E_k / k_B T} \, c_\mathrm{m} \, (n - c_\mathrm{t}) - p_0 \, e^{-E_p / k_B T} \, c_\mathrm{t}$$

### `DecayReaction` — new

A first-order radioactive decay, following the mass-action form of `GenericReaction`
with a single rate coefficient built from the half-life $t_{1/2}$:

$$R = \lambda \prod_i c_i, \qquad \lambda = \frac{\ln 2}{t_{1/2}}$$

so for a single decaying species the governing equation picks up

$$\frac{\partial c}{\partial t} = \nabla \cdot (D \nabla c) - \lambda c$$

whose solution in the absence of transport is the expected
$c(t) = c_0 \, e^{-\lambda t} = c_0 \, 2^{-t / t_{1/2}}$.

```python
import festim as F

T = F.Species("T")    # tritium
He = F.Species("He")  # helium-3

my_model.reactions = [
    F.DecayReaction(
        reactant=T,
        half_life=3.888e8,  # ~12.3 years, in the simulation's time unit
        volume=vol,
        product=He,  # optional: omit if the decay product is not tracked
    )
]
```

The reaction is irreversible (no backward rate): each reactant is consumed at rate $R$
and each product, if any, is produced at rate $R$.

### `GenericReaction` — new: any mass-action rate law

Forward and backward coefficients can be floats, UFL expressions, or callables of `T`,
`t` and `x`:

```python
import festim as F

A, B, C = F.Species("A"), F.Species("B"), F.Species("C")

my_model.reactions = [
    F.GenericReaction(
        reactant=[A, B],
        product=C,
        forward_rate=lambda T: 1e-3 * T,  # any function of temperature
        backward_rate=lambda t: 2.0 * (1 - ufl.exp(-t)),  # time-dependent
        volume=vol,
    )
]
```

$$R = k_1(T) \, c_A \, c_B - k_2(t) \, c_C$$

A rate coefficient can also depend on the concentration of another species, declared
through `arg_to_species`:

```python
# a trapping rate that saturates with the local damage concentration c_D
my_model.reactions = [
    F.GenericReaction(
        reactant=[mobile, empty_sites],
        product=trapped,
        forward_rate=lambda T, c_D: k_0 * ufl.exp(-E_k / (F.k_B * T)) * c_D,
        backward_rate=1e13,
        volume=vol,
        arg_to_species={"c_D": damage},
    )
]
```

Equivalently, the mapping can be attached to the rate itself as a `festim.Value` (the
two ways are mutually exclusive, giving both raises a `ValueError`):

```python
forward_rate = F.Value(lambda c_D: k * c_D, species_dependent_value={"c_D": damage})
```

### `ReactionBase` — new: an arbitrary net rate

When the rate is not of mass-action form at all, give the net rate directly:

```python
import festim as F

A, B = F.Species("A"), F.Species("B")

# R = 2 (c_A - c_B): A is consumed at rate R, B produced at rate R
my_model.reactions = [
    F.ReactionBase(
        reaction_rate=lambda c_A, c_B: 2.0 * (c_A - c_B),
        volume=vol,
        reactant=A,
        product=B,
        arg_to_species={"c_A": A, "c_B": B},
    )
]
```


RemDelaporteMathurin:
This should make #1123 easier!

src/festim/reaction.py
```diff
@@ -222,3 +183,168 @@ def get_concentration(species):
...
+        if self.p_0 is not None:
+
+            def backward_rate(temperature):
+                if self.p_0 and self.E_p:
+                    return self.p_0 * exp(-self.E_p / (_k_B * temperature))
+                elif self.p_0:
+                    return self.p_0
+                else:
+                    return 0
+
```
RemDelaporteMathurin, line 784: add a deprecation warning here

src/festim/reaction.py
```diff
@@ -1,81 +1,153 @@
...
-    """A reaction between two species, with a forward and backward rate.
+class GenericReaction:
+    """A generic reaction between one or more reactant species and zero or more
+    product species, taking place within a volume.
+
+    The reaction follows a mass-action net rate
+
+    .. math::
+
+        R = k_1 \\prod_i c_i^{\\text{reactant}} - k_2 \\prod_j c_j^{\\text{product}}
```
RemDelaporteMathurin, line 330: @ck768 is this form enough to account for hydride?

src/festim/reaction.py
```diff
@@ -90,135 +162,334 @@ def reactant(self, value):
...
+        unmapped = species_args - set(value)
+        if unmapped:
+            raise ValueError(
+                f"the reaction rate coefficients depend on {sorted(unmapped)}, "
+                "which must be given in arg_to_species"
+            )
+        # 2. the mapping must not contain a key used by no rate coefficient
+        unused = set(value) - species_args
+        if unused:
+            raise ValueError(
```
RemDelaporteMathurin: Maybe this should raise a warning instead.

src/festim/reaction.py
```diff
@@ -90,135 +162,334 @@ def reactant(self, value):
...
+                f"arg_to_species contains {sorted(unused)}, which is not an "
+                "argument of any reaction rate coefficient"
+            )
+        self._arg_to_species = value
+        # hand the same mapping to each coefficient; each festim.Value ignores the
+        # keys its own callable does not declare (see helpers.as_mapped_function)
+        self.forward_rate.species_dependent_value = value
+        self.backward_rate.species_dependent_value = value
```
RemDelaporteMathurin: if `forward_rate` (resp. backward) is given as a `Value` already, it may have a `species_dependent_value` mapping already given by the user. In this case, these lines would simply overwrite it.

We could:
- do this but warn the user
- raise an error and ask the user to remove one of the two (preferred)
- set a priority + warn the user

src/festim/reaction.py
```diff
@@ -90,135 +162,334 @@ def reactant(self, value):
...
+        # (HydrogenTransportProblemDiscontinuousChangeVar, its u*K_S substitution)
+        # and can be removed together with the reaction_term arguments feeding it
+        # once that class is dropped
+        overrides = overrides or [None] * len(species)
+        concentrations = []
+        for spe, override in zip(species, overrides, strict=True):
+            if override is not None:
+                concentrations.append(override)
+            elif self._mixed_domain:
+                concentrations.append(spe.concentration_submesh(self.volume))
```
RemDelaporteMathurin, line 509: `.concentration_submesh`? what is this? what's the difference with `spe.subdomain_to_solution`? or is it specific to the `ChangeOfVariable` class?

RemDelaporteMathurin: ok I looked it up and it seems to be so that it works with `ImplicitSpecies`

src/festim/hydrogen_transport_problem.py
```diff
@@ -811,6 +813,24 @@ def convert_source_input_values_to_fenics_objects(self):
...
+                        temperature=self.temperature_fenics,
+                        subdomain=reaction.volume,
+                        up_to_ufl_expr=True,
+                    )
+
+    def create_sources_from_reactions(self):
+        """Expand each reaction into one volumetric particle source per
+        participating species and add them to the sources."""
+        for reaction in self.reactions:
+            self.sources += reaction.create_sources()
```
RemDelaporteMathurin: do we want to add it to `self.souces` or do we want to mirror the `SurfaceReactionBC` behaviour where we have a `unpacked_bcs` object?

src/festim/hydrogen_transport_problem.py
```diff
@@ -2363,6 +2343,13 @@ def initialise(self):
 
         super().initialise()
 
+    def create_sources_from_reactions(self):
+        # this problem uses a change of variable (u * K_S) for mobile species, so
+        # reactions cannot be expressed as plain particle sources and are instead
+        # handled directly in add_reaction_term. This override will be removed
+        # when this deprecated problem class is dropped.
+        pass
```
RemDelaporteMathurin: do we need to keep this?

RemDelaporteMathurin: Yes we do because this function is called in the parent Problem class

test/system_tests/test_1_mobile_1_trap.py
RemDelaporteMathurin, line 1: Can you add a specific test testing `GenericReaction` (not arrhenius)

RemDelaporteMathurin (changes requested):

src/festim/reaction.py
ck768: No, I don't think the HNGD equations are mass-action rates. Dissolution and nucleation are related to how far the concentration is out of equilibrium, and nucleation is super non-linear using JMAK kinetics

Source (nucleation, dissolution) ~ K*(C - equilibrium)

RemDelaporteMathurin: is this a _source_ or a _reaction_? the difference being that there are reactants/products in a reaction

RemDelaporteMathurin: @ck768 with this PR, the reaction rates (`k`) can be whatever function of temperature, space, time, and concentration we want!

ck768: ah gotcha, that makes sense. So if the transport equation (for example dissolution) is:

dc_mobile/dt = K_dissolution * (c_mobile - threshold_dissolution),

this generic reaction class would K_dissolution, which depends on some arbitrary values/temp/etc/?

RemDelaporteMathurin: yes K_dissolution can be anything but the _form_ of the equation is still a bit rigid.

here the dissolution "reaction" is: hydride -> mobile
but the rate/source is not K_dissolution * c_hydride right?

src/festim/reaction.py
```diff
@@ -509,6 +516,105 @@ def __repr__(self) -> str:
...
+
+    @half_life.setter
+    def half_life(self, value):
+        if not isinstance(value, float | int) or isinstance(value, bool):
+            raise TypeError(f"half_life must be a float, not {type(value).__name__}")
+        if value <= 0:
+            raise ValueError(f"half_life must be positive, not {value}")
+        self._half_life = value
+        # decay constant lambda = ln(2) / t_1/2, kept in sync with half_life
+        self.forward_rate = ufl.ln(2) / value
```
RemDelaporteMathurin: I guess here the `forward_rate` could simply be a property

```python
@property
def forward_rate(self):
   return ufl.ln(2) / self._half_life
```

src/festim/reaction.py
RemDelaporteMathurin: @jhdark i think, based on what @ck768 said that currently this "generic reaction" class cannot handle a reaction rate which is $ R = k * (c1 - c2) $ so we would need an even more meta-class that can take _any_ reaction rate (not only any $k$). 

- `Meta` (should be `GenericReaction`) : $R = f(c_i, x, T, t)$
- `AnotherName` (current `GenericReaction`): $ R = f(c_i, x, T, t) * c_A * c_B - g(c_i, x, T, t) * c_C * c_D $
- `ArrheniusReaction` : ${f, g} = $ arrhenius laws


Note here that the difference between `Meta` and a normal `Source` is that the reaction "unpacks" the source terms to the reactants and products

src/festim/reaction.py
```diff
@@ -1,81 +1,163 @@
...
-        E_k: float,
         volume: VolumeSubdomain,
-        product: Union[_Species, list[_Species]] | None = [],
-        p_0: float | None = None,
-        E_p: float | None = None,
+        reactant: Species | ImplicitSpecies | list[Species | ImplicitSpecies],
+        product: Species | list[Species] | None,
+        forward_rate: float
+        | int
+        | np.ndarray
```
RemDelaporteMathurin: Can it really be an array?

src/festim/reaction.py
```diff
@@ -1,81 +1,163 @@
...
+        | Callable
+        | fem.Constant
+        | fem.Expression
+        | ufl.core.expr.Expr
+        | fem.Function
+        | Value,
+        backward_rate: float
+        | int
+        | np.ndarray
```
RemDelaporteMathurin: Can it really be an array?

src/festim/reaction.py
```diff
@@ -1,81 +1,163 @@
...
-        self.volume = volume
+        self.forward_rate = forward_rate
+        self.backward_rate = backward_rate
+        # a species mapping may be attached directly to a rate coefficient Value
+        # (its species_dependent_value); remember that so the arg_to_species setter
+        # can refuse to receive the same information twice
+        self._rate_has_own_species_map = bool(
+            self.forward_rate.species_dependent_value
+            or self.backward_rate.species_dependent_value
+        )
```
RemDelaporteMathurin: Make this a property to avoid potential sync issues between `_rate_has_own_species_map` and the rates

src/festim/reaction.py
RemDelaporteMathurin: You resolved this but i still see it as a ValueError instead of a warning

RemDelaporteMathurin (changes requested):

src/festim/reaction.py
```diff
@@ -1,81 +1,430 @@
...
+            volume = F.VolumeSubdomain(id=1, material=material)
+
+            A = F.Species("A")
+            B = F.Species("B")
+
+            # an arbitrary rate R = 2 * (c_A - c_B): A is consumed and B produced
+            reaction = F.ReactionBase(
+                reaction_rate=lambda c_A, c_B: 2.0 * (c_A - c_B),
+                volume=volume,
+                species_to_stoichiometry={A: -1, B: 1},
```
RemDelaporteMathurin: any particular reason for stepping out of the `reactant` and `product` arguments?

src/festim/reaction.py
```diff
@@ -1,81 +1,430 @@
...
+        consumed = [(s, c) for s, c in self.species_to_stoichiometry.items() if c < 0]
+        produced = [(s, c) for s, c in self.species_to_stoichiometry.items() if c > 0]
+        lhs, rhs = side(consumed), side(produced)
+        return f"{lhs} --> {rhs}" if rhs else f"{lhs} -->"
+
+    def reaction_term(self) -> ufl.core.expr.Expr:
+        """The net reaction rate ``R`` as a ufl expression.
+
+        ``reaction_rate`` must already be converted to a fenics object (done by
+        ``HydrogenTransportProblem.convert_reaction_rates_to_fenics_objects``).
```
RemDelaporteMathurin, line 284: I feel in this case we need to have an assertion

```python
assert self.reaction_rate.hasattr("fenics_object"), "fenics_object has not been defined"
```

src/festim/hydrogen_transport_problem.py
```diff
@@ -350,9 +350,9 @@ def initialise(self):
     def create_implicit_species_value_fenics(self):
         """For each implicit species, create the value_fenics."""
         for reaction in self.reactions:
-            for reactant in reaction.reactant:
-                if isinstance(reactant, _species.ImplicitSpecies):
-                    reactant.create_value_fenics(
+            for spe in reaction.species:
+                if isinstance(spe, _species.ImplicitSpecies):
+                    spe.create_value_fenics(
```
RemDelaporteMathurin, line 355: I don't think this would fail but just wondering why this is needed? ImplicitSpecies should not be a product hence why we only iterate through reactants before

src/festim/hydrogen_transport_problem.py
```diff
@@ -999,15 +1006,15 @@ def update_time_dependent_values(self):
         t = float(self.t)
 
         for reaction in self.reactions:
-            for reactant in reaction.reactant:
-                if isinstance(reactant, _species.ImplicitSpecies):
-                    reactant.update_density(t=t)
+            for spe in reaction.species:
+                if isinstance(spe, _species.ImplicitSpecies):
+                    spe.update_density(t=t)
```
RemDelaporteMathurin, line 1011: same question as abov

src/festim/reaction.py
```diff
@@ -215,85 +482,16 @@ def backward_rate(self, value):
...
-        return (
-            f"{type(self).__name__}({reactants} <--> {products}, "
-            f"{self.forward_rate}, {self.backward_rate})"
-        )
-
-    def __str__(self) -> str:
-        reactants = " + ".join([str(reactant) for reactant in self.reactant])
-        products = " + ".join([str(product) for product in self.product])
-        return f"{reactants} <--> {products}"
+    @ReactionBase.species_to_stoichiometry.getter
```
RemDelaporteMathurin: can you explain that to me please? not familiar with this syntax

src/festim/reaction.py
```diff
@@ -215,85 +482,16 @@ def backward_rate(self, value):
...
+    def species_to_stoichiometry(self):
+        # each reactant is consumed (nu = -1) and each product produced (nu = +1);
+        # derived live from reactant/product so it never desyncs from them
+        stoichiometry = {reactant: -1 for reactant in self.reactant}
+        stoichiometry.update({product: 1 for product in self.product})
+        return stoichiometry
```
RemDelaporteMathurin: I don't think this works as intended

```python
reactant = ["H", "H", "H"]
stoichiometry = {reactant: -1 for reactant in reactant}
print(stoichiometry)
```

RemDelaporteMathurin (changes requested):

RemDelaporteMathurin (approved):

RemDelaporteMathurin:
why weren't the tests triggered?

RemDelaporteMathurin (approved):

PR #1192, opened 2026-07-17
Support for gas enclosures

### LINKED ISSUE #996: Support for enclosures
Some applications require tracking the evolution of the H partial pressure in an enclosure in contact with permeable membranes.

For example, see [this TMAP verification case](https://festim-vv-report.readthedocs.io/en/latest/verification/mes/depleting_source.html).

The way we did it in this case was using an explicit scheme. But thanks to @jorgensd 's `scifem` package, we can use Real elements and include these equations in the variational formulation.

RemDelaporteMathurin:
### Summary
This PR introduces gas enclosures to FESTIM! 🎉 

### Related Issues
Fixes #996 

### Motivation and Context
FESTIM can impose a gas pressure on a surface, but only as *input* — `SievertsBC.pressure`, `HenrysBC.pressure`, and `SurfaceReactionBC.gas_pressure` all accept `float | callable(t)` and nothing else. There is no way to let the pressure of a gas volume *evolve* in response to what permeates through the wall.

This adds that: an `Enclosure` (a 0D gas volume) whose pressure is a genuine unknown, solved implicitly and monolithically with the transport problem, coupled to the solid through the surfaces in contact with it, and optionally exchanging gas with the outside through openings. The reference case is the TMAP verification in issue #996: a slab with a perfect sink on one side and Henry's law on the other, with `dP/dt = φ·A·k·T/V`.

The enabling technology is the **real element** (one global DOF), native in dolfinx 0.11. The alternative — an explicit/staggered update of P between timesteps — is what we are deliberately avoiding; real elements put the pressure ODE directly into the variational form so Newton sees the full coupling.

**Scope: `HydrogenTransportProblemDiscontinuous` only.** The base `HydrogenTransportProblem` is expected to be retired in favour of the Discontinuous class's blocked system, so enclosures are not worth building for it. The base class raises a clear `NotImplementedError` if given enclosures.

**Limits** Right now this PR only includes "flux coupling" between enclosures and surfaces via `SurfaceReactionBC.gas_pressure`. "concentration coupling" via `SievertsBC.pressure`, `HenrysBC.pressure` will probably come in a separate PR for simplicity.

image

## Usage
<details>
  <summary>Click to expand</summary>
  
  ```python
  """Gas enclosure coupled to a slab through a surface reaction.
  
  A 1D slab is initially loaded with hydrogen. Its right surface faces a gas enclosure
  that starts under vacuum. Hydrogen diffuses to that surface, recombines into H2
  (2H <-> H2), and fills the enclosure, so the partial pressure builds up over time.
  
  The pressure is a genuine unknown of the problem, solved together with the transport
  problem, so the reaction sees the pressure it has itself created: as P rises,
  dissociation (kd * P) pushes back against recombination (kr * c^2) until the surface
  reaches equilibrium.
  
  Run with:
  
      python example_enclosure.py
  
  Requires dolfinx >= 0.11 (the enclosure pressure lives in a real function space).
  """
  
  import numpy as np
  from mpi4py import MPI
  
  import dolfinx
  
  import festim as F
  
  # ---------------------------------------------------------------------------
  # Parameters
  # ---------------------------------------------------------------------------
  L = 1e-3  # slab thickness (m)
  AREA = 1e-2  # area of the membrane facing the enclosure (m2)
  VOLUME = 1e-5  # volume of the enclosure (m3)
  TEMPERATURE = 600.0  # K
  
  c_0 = 1e21  # initial concentration in the slab (m-3)
  D_0, E_D = 1e-7, 0.2  # diffusion coefficient
  
  k_r0, E_kr = 1e-26, 0.0  # recombination: 2H -> H2
  k_d0, E_kd = 1e14, 0.0  # dissociation: H2 -> 2H
  
  final_time = 5000.0
  stepsize = 25.0
  
  # ---------------------------------------------------------------------------
  # Model
  # ---------------------------------------------------------------------------
  my_model = F.HydrogenTransportProblemDiscontinuous()
  my_model.mesh = F.Mesh1D(np.linspace(0, L, num=200))
  
  material = F.Material(name="tungsten", D_0=D_0, E_D=E_D)
  slab = F.VolumeSubdomain1D(id=1, borders=[0, L], material=material)
  left = F.SurfaceSubdomain1D(id=1, x=0)
  right = F.SurfaceSubdomain1D(id=2, x=L)
  my_model.subdomains = [slab, left, right]
  
  H = F.Species("H", subdomains=[slab])
  my_model.species = [H]
  my_model.temperature = TEMPERATURE
  
  # The gas species whose partial pressure is the unknown, and the enclosure it lives in.
  # The slab is 1D, so its right surface is a point and carries no area: the area of the
  # membrane facing the enclosure has to be given explicitly.
  H2 = F.GasSpecies(name="H2", initial_pressure=0.0)
  enclosure = F.Enclosure(
      volume=VOLUME,
      species=[H2],
      temperature=TEMPERATURE,
      surfaces={right: AREA},
      name="plenum",
  )
  my_model.enclosures = [enclosure]
  
  # The reaction that couples the two: passing the GasSpecies (rather than a float) is
  # what makes the pressure an unknown instead of an imposed value.
      F.SurfaceReactionBC(
          reactant=[H, H],
          gas_pressure=H2,
          k_r0=k_r0,
          E_kr=E_kr,
          k_d0=k_d0,
          E_kd=E_kd,
          subdomain=right,
      )
  ]
  
  my_model.initial_conditions = [
      F.InitialConcentration(value=c_0, species=H, volume=slab)
  ]
  
  # ---------------------------------------------------------------------------
  # Post-processing
  # ---------------------------------------------------------------------------
  pressure = F.GasPressure(field=H2, filename="enclosure_pressure.csv")
  inventory = F.TotalVolume(field=H, volume=slab)
  my_model.exports = [pressure, inventory]
  
      atol=1e10,
      rtol=1e-10,
      transient=True,
      final_time=final_time,
      stepsize=F.Stepsize(stepsize),
  )
  
  my_model.initialise()
  my_model.run()
  
  # ---------------------------------------------------------------------------
  # Results
  # ---------------------------------------------------------------------------
  # Nothing leaves the system, so every H atom lost by the slab ends up in the gas as
  # half an H2 molecule. This is exact, not approximate: the same expression drives the
  # surface flux and the pressure balance.
  t = np.array(pressure.t)
  P = np.array(pressure.data)
  in_solid = AREA * np.array(inventory.data)
  in_gas = 2.0 * P * VOLUME / (F.k_B_SI * TEMPERATURE)
  total = in_solid + in_gas
  
  print(f"final pressure          : {P[-1]:.4e} Pa")
  print(f"H atoms in slab, start  : {in_solid[0]:.4e}")
  print(f"H atoms in slab, end    : {in_solid[-1]:.4e}")
  print(f"H atoms in gas, end     : {in_gas[-1]:.4e}")
  print(
      f"inventory drift         : {abs(total[-1] - total[0]) / total[0]:.2e} (relative)"
  )
  
  fig, (ax_p, ax_n) = plt.subplots(2, 1, figsize=(7, 7), sharex=True)
  
  ax_p.plot(t, P, color="tab:red")
  ax_p.set_ylabel("H2 partial pressure (Pa)")
  ax_p.set_title("Enclosure filling from a slab through 2H <-> H2")
  ax_p.grid(alpha=0.3)
  
  ax_n.plot(t, in_solid, label="in the slab")
  ax_n.plot(t, in_gas, label="in the enclosure")
  ax_n.plot(t, total, "k--", label="total (conserved)")
  ax_n.set_xlabel("time (s)")
  ax_n.set_ylabel("H atoms")
  ax_n.legend()
  ax_n.grid(alpha=0.3)
  
  fig.tight_layout()
  fig.savefig("enclosure_pressure.png", dpi=150)
  print("\nwrote enclosure_pressure.png and enclosure_pressure.csv")
  
  ```
</details>

## Type of Change

- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation

### DIFF
--- a/docs/source/api/index.rst
+++ b/docs/source/api/index.rst
@@ -7,6 +7,10 @@ API reference
     :members:
     :show-inheritance:
 
+.. automodule:: festim.enclosure
+    :members:
+    :show-inheritance:
+
 .. automodule:: festim.exports
     :members:
     :show-inheritance:
--- a/src/festim/__init__.py
+++ b/src/festim/__init__.py
@@ -9,6 +9,7 @@ except Exception:
 
 R = 8.314462618  # Gas constant J.mol-1.K-1
 k_B = 8.6173303e-5  # Boltzmann constant eV.K-1
+k_B_SI = 1.380649e-23  # Boltzmann constant J.K-1
 
 from .advection import AdvectionTerm, VelocityField
 from .boundary_conditions.dirichlet_bc import (
@@ -24,10 +25,20 @@ from .boundary_conditions.surface_reaction import SurfaceReactionBC
 from .coupled_heat_hydrogen_problem import (
     CoupledTransientHeatTransferHydrogenTransport,
 )
+from .enclosure.enclosure import Enclosure
+from .enclosure.gas_species import GasSpecies
+from .enclosure.openings import (
+    EnclosureConnection,
+    OpeningBase,
+    PrescribedFlowRate,
+    Pump,
+    Reservoir,
+)
 from .exports.average_surface import AverageSurface
 from .exports.average_volume import AverageVolume
 from .exports.custom_quantity import CustomQuantity
 from .exports.derived_quantity import DerivedQuantity
+from .exports.gas_pressure import GasPressure
 from .exports.maximum_surface import MaximumSurface
 from .exports.maximum_volume import MaximumVolume
 from .exports.minimum_surface import MinimumSurface
--- a/src/festim/boundary_conditions/surface_reaction.py
+++ b/src/festim/boundary_conditions/surface_reaction.py
@@ -3,6 +3,7 @@ from dolfinx import fem
 
 from festim import k_B
 from festim.boundary_conditions import ParticleFluxBC
+from festim.enclosure.gas_species import GasSpecies
 
 
 class SurfaceReactionBCpartial(ParticleFluxBC):
@@ -22,7 +23,10 @@ class SurfaceReactionBCpartial(ParticleFluxBC):
 
     Args:
         reactant (list): list of F.Species objects representing the reactants
-        gas_pressure (float, callable): the partial pressure of the product species
+        gas_pressure (float, callable or F.GasSpecies): the partial pressure of
+            the product species. If a F.GasSpecies is given, the pressure is an
+            unknown of the problem and the reaction feeds the mass balance of
+            the enclosure that species belongs to.
         k_r0 (float): the pre-exponential factor of the forward reaction rate
         E_kr (float): the activation energy of the forward reaction rate (eV)
         k_d0 (float): the pre-exponential factor of the backward reaction rate
@@ -53,7 +57,10 @@ class SurfaceReactionBCpartial(ParticleFluxBC):
     def create_value_fenics(self, mesh, temperature, t: fem.Constant):
         kr = self.k_r0 * ufl.exp(-self.E_kr / (k_B * temperature))
         kd = self.k_d0 * ufl.exp(-self.E_kd / (k_B * temperature))
-        if callable(self.gas_pressure):
+        if isinstance(self.gas_pressure, GasSpecies):
+            # the pressure is an unknown of the problem, living in an enclosure
+            gas_pressure = self.gas_pressure.solution
+        elif callable(self.gas_pressure):
             gas_pressure = self.gas_pressure(t=t)
         else:
             gas_pressure = self.gas_pressure
@@ -95,7 +102,10 @@ class SurfaceReactionBC:
 
     Args:
         reactant (list): list of F.Species objects representing the reactants
-        gas_pressure (float, callable): the partial pressure of the product species
+        gas_pressure (float, callable or F.GasSpecies): the partial pressure of
+            the product species. If a F.GasSpecies is given, the pressure is an
+            unknown of the problem and the reaction feeds the mass balance of
+            the enclosure that species belongs to.
         k_r0 (float): the pre-exponential factor of the forward reaction rate
         E_kr (float): the activation energy of the forward reaction rate (eV)
         k_d0 (float): the pre-exponential factor of the backward reaction rate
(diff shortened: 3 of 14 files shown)


RemDelaporteMathurin:
FWIW here's what Opus 4.8 had to say when I asked it if there would be a performance gain in defining the real elements on a submesh:

> Measured, and the answer is no — here's why, with the scaling that settles it:
> 
> 
> n=   8 | vol_dofs=  81 | surf_facets=  8 | nnz(dF_P/du)=  17
> n=  16 | vol_dofs= 289 | surf_facets= 16 | nnz(dF_P/du)=  33
> n=  32 | vol_dofs=1089 | surf_facets= 32 | nnz(dF_P/du)=  65
> n=  64 | vol_dofs=4225 | surf_facets= 64 | nnz(dF_P/du)= 129
> That's the coupling block ∂(pressure residual)/∂(concentration). As the mesh refines 8→64, the volume DOFs grow ~quadratically (81→4225, ×4 per step) but the pressure coupling grows linearly (17→129, ×2 per step) — it tracks the number of surface DOFs, not volume DOFs. The pressure row is already sparse: it touches only the contact-surface DOFs, never the interior. It is not a dense row over the whole mesh.
> 
> The reason is structural: in dolfinx the sparsity pattern of a form comes from the integration measure, not from which mesh the space is defined on. The pressure residual integrates over ds(surface.id), so the real DOF couples only to the DOFs on those facets — that's already the minimum possible. A facet submesh can't make it sparser than surface-only, because it already is.
> 
> So on the three things that drive cost:
> 
> DOF count — a real space has exactly 1 global DOF whether it lives on the parent mesh or a facet submesh. No change.
> Coupling / LU fill-in — already surface-local sparse, as measured. No change.
> Assembly — the forms iterate only their integration entities (ds facets, or dx for surfaceless enclosures); the real q is just a constant coefficient. The parent mesh doesn't force any extra traversal. No change.
> If anything, a facet submesh adds a small amount of entity_maps bookkeeping (the real space would be on yet another mesh), so it's marginally more overhead, not less.
> 
> Verdict: the submesh route is a code-structure choice — it matches your scifem demo's style and keeps the pressure "attached" to its surface conceptually — but there's no performance case for it. I'd downgrade that # NOTE to say exactly that, so nobody later chases it expecting a speedup.

RemDelaporteMathurin:
Reviewed yesterday with @jhdark

Issue #1191, opened 2026-07-16
Only use mixed formulation method


RemDelaporteMathurin:
It's annoying to have to juggle between the mixed formulation class (`Disctontinuous`) and the standard one. We could just use mixed-formulation all the time

PR #1195, opened 2026-07-18
Fix Nitsche weak Dirichlet BC formulation

RemDelaporteMathurin:
### Summary
The Nitsche terms in DirichletBCBase.weak_formulation omitted the diffusion coefficient from the consistency and symmetry terms, making the scheme consistent only when D == 1. At D = 0.01 the L2 convergence rate dropped to ~0.85 and the error varied by ~30x with the penalty parameter. The penalty term also carried a negative sign.

- weak_formulation now takes D and uses the symmetric Nitsche form
- the numerical flux (-D grad(u).n + alpha/h (u - g)), the quantity the discrete scheme actually conserves, is exposed as its own method for reuse by anything consuming the flux through a weakly enforced surface
- both call sites resolve D from the material of the volume subdomain owning the surface; the base class gains volume_subdomain_of_surface for that, mirroring surface_to_volume in the discontinuous class

The existing MMS test passes with the bugs present (D ~= 0.12 with alpha/h ~= 15000 brute-forces u -> g), so add tests that do not: a convergence-rate test at D_0 = 0.01 and a penalty-insensitivity test. Both now give order 2.000 and errors identical to 4 significant figures.

+ now the global penalty scales with both the mesh size (`h`) and the diffusivity (`D`) making the penalty more robust.

### Related Issues
Related to #996 

### Motivation and Context
While working on #1192 , Claude noticed there was an error in how we implemented the weak dirichlet BC as the diffusion coefficient was missing from it. This basically means the penalty has to be increased A LOT in order for the error to be low enough. Convergence was not great though

This is a problem for the enclosure support as we'll rely on weakly enforced Dirichlet BCs/.

## Type of Change

- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation

### DIFF
--- a/src/festim/boundary_conditions/dirichlet_bc.py
+++ b/src/festim/boundary_conditions/dirichlet_bc.py
@@ -151,27 +151,30 @@ class DirichletBCBase:
         elif self.bc_expr is not None:
             self.value_fenics.interpolate(self.bc_expr)
 
-    def weak_formulation(
+    def numerical_flux(
         self,
         u: fem.Function | ufl.indexed.Indexed,
-        v: ufl.argument.Argument | ufl.indexed.Indexed,
-        ds: ufl.Measure,
+        D: ufl.core.expr.Expr | fem.Function | fem.Constant,
+        mesh,
     ) -> ufl.core.expr.Expr:
-        """
-        Returns the Nitsche weak formulation for the BC
-        This follows the dolfinx tutorial
-        https://jsdokken.com/dolfinx-tutorial/chapter1/nitsche.html
+        """Returns the numerical flux leaving the domain through the surface of the BC.
+
+        Nitsche's method only enforces ``u = value`` weakly, so ``u - value`` does not
+        vanish on the boundary and the raw gradient ``-D grad(u).n`` is not the quantity
+        the discrete scheme conserves. The conserved quantity is this numerical flux,
+        which includes the penalty contribution. Use it (rather than the raw gradient)
+        whenever the flux through a weakly enforced boundary feeds another equation, so
+        that the discrete balance is exact.
 
         Args:
             u: the solution function associated to the species for which the BC
                 is applied
-            v: the test function
-            ds: the surface measure
+            D: the diffusion coefficient of the species at this surface
+            mesh: the mesh the surface belongs to
 
         Returns:
-            the weak formulation
+            the numerical flux, positive when leaving the domain
         """
-        mesh = ds.ufl_domain()
         n = ufl.FacetNormal(mesh)
         h = ufl.Circumradius(mesh)  # FIXME this doesn't work for rectangles
         alpha = self.penalty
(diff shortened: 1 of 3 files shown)


src/festim/hydrogen_transport_problem.py
```diff
@@ -192,6 +192,7 @@ def __init__(
         self._temperature_as_function = None
         self._species_to_D_global = None
         self._species_to_D_global_expr = None
+        self._surface_to_volume = None
```
RemDelaporteMathurin, line 202: This is needed to make this fix work with `HydrogenTransportProblem`. Shouldn't be an issue anyway since we want to get rid of this class #1191

src/festim/boundary_conditions/dirichlet_bc.py
```diff
@@ -151,14 +153,54 @@ def update(self, t: float):
         elif self.bc_expr is not None:
             self.value_fenics.interpolate(self.bc_expr)
 
+    def numerical_flux(
+        self,
+        u: fem.Function | ufl.indexed.Indexed,
+        D: ufl.core.expr.Expr | fem.Function | fem.Constant,
+        mesh,
+    ) -> ufl.core.expr.Expr:
```
jhdark, line 161: What is the purpose of this new class? Its not used anywhere else, it seems to just be covering two of the 3 parts for this formulation too?

I know we've been talking about wanting to handle numerical fluxes more directly to make adding influencing factors to the flux easier. Is this a move towards that?

src/festim/boundary_conditions/dirichlet_bc.py
```diff
@@ -151,14 +153,54 @@ def update(self, t: float):
...
+        """
+        n = ufl.FacetNormal(mesh)
+        h = ufl.Circumradius(mesh)  # FIXME this doesn't work for rectangles
+        alpha = self.penalty
+        assert alpha is not None, (
+            "Penalty parameter must be given for weakly enforced Dirichlet BCs"
+        )
+        assert self.value_fenics is not None, (
+            "value_fenics must be defined for weakly enforced Dirichlet BCs"
+        )
```
jhdark, line 188: I know this has just been grandfathered in, but it seems like a error more for devs than users, as users should never interact with this parameter?

jhdark (changes requested):

src/festim/boundary_conditions/dirichlet_bc.py
RemDelaporteMathurin: Correct. But it's good to have imo

Makes the reason of a potential failure more explicit

src/festim/boundary_conditions/dirichlet_bc.py
RemDelaporteMathurin: It's used below in the dirichlet form. It helps making the Nitsche formulation clearer and it's also a bit of future proofing as when we add more diffusion processes like Soret we will just have to change this method.

More importantly, it is used in the Dirichlet coupling with enclosures in the other PR (can't ref cause on my phone) where we use this numerical flux instead of the simple gradient to ensure mass balance

Issue #1174, opened 2026-06-19
Update to dolfinx 0.11


RemDelaporteMathurin:
@jhdark @jorgensd we probably need to wait for scifem to support 0.11 before doing the switch

jorgensd:
> @jhdark @jorgensd we probably need to wait for scifem to support 0.11 before doing the switch

It should support 0.11? Anything specific you are thinking about?

jhdark:
@jorgensd just struggling to even make the conda env atm:

<img width="761" height="224" alt="Image" src="https://github.com/user-attachments/assets/e54b367e-816e-4385-8713-82b744236fcf" />

jorgensd:
Progress at: https://github.com/conda-forge/scifem-feedstock/pull/44#issuecomment-4755506433

PR #1203, opened 2026-07-21
Replace scifem function with new library function

jorgensd:
Replaces the scifem function with the functionality introduced in DOLFINx v0.11


RemDelaporteMathurin (approved):
Thanks for this @jorgensd we missed it in #1174

PR #1204, opened 2026-07-21
Bump minor version of scifem. Use --no-build-isolation for installation

jorgensd:
Bump minor version of scifem to get scipy as dependency.
Use `--no-build-isolation` for installation of scifem.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ description = "Finite element simulations of hydrogen transport"
 readme = "README.md"
 requires-python = ">=3.9"
 license = "Apache-2.0"
-dependencies = ["tqdm", "scifem>=0.20.2", "io4dolfinx"]
+dependencies = ["tqdm", "scifem>=0.20.0", "io4dolfinx"]
 classifiers = [
     "Natural Language :: English",
     "Topic :: Scientific/Engineering",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ description = "Finite element simulations of hydrogen transport"
 readme = "README.md"
 requires-python = ">=3.9"
 license = "Apache-2.0"
-dependencies = ["tqdm", "scifem>=0.2.13", "io4dolfinx"]
+dependencies = ["tqdm", "scifem>=0.20.2", "io4dolfinx"]
 classifiers = [
     "Natural Language :: English",
     "Topic :: Scientific/Engineering",
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -4,7 +4,7 @@ on: [pull_request, push]
 jobs:
   benchmark:
     runs-on: ubuntu-latest
-    container: dolfinx/dolfinx:v0.10.0
+    container: dolfinx/dolfinx:v0.11.0
 
     steps:
       - name: Checkout code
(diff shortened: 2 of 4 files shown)


RemDelaporteMathurin (approved):
thanks seems like it solves the build issue thank you!

RemDelaporteMathurin:
oops spoke too soon, seems like the conda ci is now failing https://github.com/festim-dev/FESTIM/actions/runs/29855101324/job/88717339559?pr=1204

jhdark:
> oops spoke too soon, seems like the conda ci is now failing https://github.com/festim-dev/FESTIM/actions/runs/29855101324/job/88717339559?pr=1204

Do we need to also install scikit?

pyproject.toml
```diff
@@ -17,7 +17,7 @@ description = "Finite element simulations of hydrogen transport"
 readme = "README.md"
 requires-python = ">=3.9"
 license = "Apache-2.0"
-dependencies = ["tqdm", "scifem>=0.2.13", "io4dolfinx"]
+dependencies = ["tqdm", "scifem>=0.20.2", "io4dolfinx"]
```
jhdark:
```suggestion
dependencies = ["tqdm", "scifem>=0.20.0", "io4dolfinx"]
```

I got it working on another PR branch by just reducing the floor for this to allow it to work with v0.10

jorgensd:
I'm more worried that several tests fail on dolfinx nightly..

jhdark:
> I'm more worried that several tests fail on dolfinx nightly..

I've been looking into this, seems to do with how we have been testing things. But there seems to be a change in the ordering of arrays, so its no longer guaranteed that the left most node is the `u.x.array[0]` for instance. But nonetheless we could fix that in a different PR. At least scifem installs now ha

jorgensd:
> > I'm more worried that several tests fail on dolfinx nightly..
> 
> I've been looking into this, seems to do with how we have been testing things. But there seems to be a change in the ordering of arrays, so its no longer guaranteed that the left most node is the `u.x.array[0]` for instance. But nonetheless we could fix that in a different PR. At least scifem installs now ha

Did you assume that u.x.array[0] was the degree of freedom at 0.0?
That has never been a safe assumption.
The change likely has to do with: https://github.com/FEniCS/dolfinx/pull/4286
which massively improves `V.sub(i).collapse()` as seen in: https://github.com/FEniCS/dolfinx/pull/4286#issuecomment-5011316510

jhdark:
> > > I'm more worried that several tests fail on dolfinx nightly..
> > 
> > 
> > I've been looking into this, seems to do with how we have been testing things. But there seems to be a change in the ordering of arrays, so its no longer guaranteed that the left most node is the `u.x.array[0]` for instance. But nonetheless we could fix that in a different PR. At least scifem installs now ha
> 
> Did you assume that u.x.array[0] was the degree of freedom at 0.0? That has never been a safe assumption. The change likely has to do with: [FEniCS/dolfinx#4286](https://github.com/FEniCS/dolfinx/pull/4286) which massively improves `V.sub(i).collapse()` as seen in: [FEniCS/dolfinx#4286 (comment)](https://github.com/FEniCS/dolfinx/pull/4286#issuecomment-5011316510)

Yes exactly, we shouldnt have been testing this way in the first place, laziness on our part.

PR #1216, opened 2026-07-30
Codim: base functionality

RemDelaporteMathurin:
## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔨 Code refactoring (no functional changes, no API changes)
- [ ] 📝 Documentation update
- [ ] ✅ Test update (adding missing tests or correcting existing tests)
- [ ] 🔧 Build/CI configuration change

## Testing

- [ ] All existing tests pass locally (`pytest`)
- [ ] I have added new tests that prove my fix is effective or that my feature works

## Code Quality Checklist

- [ ] My code follows the code style of this project (Ruff formatted: `ruff format .`)
- [ ] My code passes linting checks (`ruff check .`)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas

## Documentation

- [ ] I have updated the documentation accordingly (if applicable)
- [ ] I have added docstrings to new functions/classes following the project conventions

Issue #1208, opened 2026-07-23
Co-dimensional problems coupling 


ee-nn:
The idea would be to couple together a $n$-dimensional transport formulation on a subdomain to an $n-1$-dimensional transport formulation on an interface to capture trapping & diffusion. See [Remi's gist](https://gist.github.com/RemDelaporteMathurin/d1a678b6b7439e339c8471e97cd31a39) for a trapping example (no diffusion) and [this FEniCs discourse thread](https://fenicsproject.discourse.group/t/coupled-problem-with-codim-1-submesh-wrong-derivative/19688/3) for a potential implementation of the required manifold derivatives. 

Edit 7/27/26: my MWE issues with using `ufl.dx(domain=submesh)` have been fixed thanks the workaround script that @jorgensd provided, which provides the necessary modifications to the `assemble_residual` and `assemble_jacobian` functions

Edit 7/28/26: Issues with extending this to multiple submeshes also should be fixed

Edit 7/29/26: Scratch work/MWEs related to this development can be found in [this folder](https://github.com/ee-nn/FESTIM/tree/multi-dimensional-coupling/scratch_work) of this [dev branch](https://github.com/ee-nn/FESTIM/tree/multi-dimensional-coupling)

RemDelaporteMathurin:
Beyond the technicality of the implementation, I think it would be worth thinking about how it would look like from the user side of things.

Is it a new `VolumeSubdomain` class? how would user detail what type of coupling they want ,etc

ee-nn:
@RemDelaporteMathurin Thanks for the pointer, I do think having an `InterfaceSubdomain` class that inherits from `VolumeSubdomain` could be a potential implementation. More thoughts/details to follow!

RemDelaporteMathurin:
this sort of codimensional coupling is not restricted to interfaces between two subdomains though. You can think of something similar for a 2D boundary coupled to a 3D domain for instance where species are allowed to diffuse _along_ the surface

ee-nn:
I forgot about that application, in that case I agree that inheritance from `VolumeSubdomain` would be too restrictive

RemDelaporteMathurin:
I think we clearly have a new "Subdomain" object here and it's closer to a volume subdomain than a surface subdomain in the classical sense because we can _solve_ a transport equation on it. Whereas a classical `SurfaceSubdomain` is only used for boundary conditions/exports definition.

[...]

PR #1224, opened 2026-08-05
Nested codim coupling

ee-nn:
## Description

First attempt at adding in capability for nested co-dimensional domains. Helps with issue #1208 and adds on top of functionality in PR #1216 

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📝 Documentation update


.gitignore
```diff
@@ -113,6 +113,8 @@ venv.bak/
 *.xdmf
 *.h5
 *.txt
+*.zip
+*.png
```
ee-nn, line 117: Just realized these were added from my earlier work, might be nice to have though?

RemDelaporteMathurin:
As discussed on Slack, I think it would be preferable to hold off this PR a little bit to avoid increasing the complexity of this already big jump too fast. 

My understanding is that such nested codim 1 problems are needed for specific transport at triple junctions and/or GB-specific properties (accounting for misalignment for instance).

Issue #1225, opened 2026-08-08
[BUG] Minimum and Maximum exports fail for the discontinuous problem


ee-nn:
**Describe the bug**
Using any of the extrema quantities on `HydrogenTransportProblemDiscontinuous` -- `MaximumVolume,`, `MaximumSurface`, `MinimumVolume` or `MinimumSurface`--leads to the AttributeErrors `'[Minimum/Maximum]Volume' object has no attribute 'volume_meshtags'` and `[Minimum/Maximum]Surface object has no attribute 'facet_meshtags'`, respectively. 

I'll link a PR shortly to provide a fix for this

EDIT: PR is #1226 . Note that this is rather temporary as https://github.com/ee-nn/FESTIM/tree/derived-quantities-revamp will refactor these capabilities after #1216 is released

<details>
<summary> Full traceback for volume </summary>

```bash
Traceback (most recent call last):
  File "~/min_max_export_bug.py", line 71, in <module>
    my_model.run()
    ~~~~~~~~~~~~^^
  File "~/src/festim/hydrogen_transport_problem.py", line 2376, in run
    self.post_processing()
    ~~~~~~~~~~~~~~~~~~~~^^
  File "~/src/festim/hydrogen_transport_problem.py", line 2264, in post_processing
    export.compute()
    ~~~~~~~~~~~~~~^^
  File "~/src/festim/exports/minimum_volume.py", line 32, in compute
    entities = self.volume_meshtags.find(self.volume.id)
               ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'MinimumVolume' object has no attribute 'volume_meshtags'

```

</details>

<details>
<summary> Full traceback for surface </summary>

```bash
 File "~/min_max_export_bug.py", line 73, in <module>
    my_model.run()
    ~~~~~~~~~~~~^^
  File "~/src/festim/hydrogen_transport_problem.py", line 2376, in run
    self.post_processing()
    ~~~~~~~~~~~~~~~~~~~~^^
  File "~/src/festim/hydrogen_transport_problem.py", line 2250, in post_processing
    export.compute()
    ~~~~~~~~~~~~~~^^
  File "~/src/festim/exports/minimum_surface.py", line 32, in compute
    entities = self.facet_meshtags.find(self.surface.id)
               ^^^^^^^^^^^^^^^^^^^
AttributeError: 'MinimumSurface' object has no attribute 'facet_meshtags'

```

</details>

**To Reproduce**
[...]

PR #1226, opened 2026-08-08
bug fix for min/max values not working on discontinuous problem

ee-nn:
## Description

Provides a fix for issue #1225. 

### Summary
Essentially #1225 was the result of two separate problems: 

1. `initialise_exports()` of `HydrogenTransportProblemDiscontinuous` overrides the base method from `HydrogenTransportProblem`, but it never initializes its own `export.volume_meshtags` or `export.facet_meshtags`. Since it doesn't invoke `super()` either, these attributes of `export` aren't carried over to the discontinuous class. 

2. Even with the meshtags attached, `compute()` in Min/MaxVolume and Min/MaxSurface read `self.field.post_processing_solution`, which stays `None` in the discontinuous case because the solution lives per-submesh in `species.subdomain_to_post_processing_solution[subdomain].` 

## Type of Change

- [X] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] 📝 Documentation update

## Testing

Specifically, new system test for the discontinuous case (which is essentially the MWE listed in #1225 - not sure if this is overkill?), and some smaller assertion tests to ensure that `meshtags` are always defined.

## Code Quality Checklist

## Documentation

### DIFF
--- a/src/festim/exports/maximum_surface.py
+++ b/src/festim/exports/maximum_surface.py
@@ -17,28 +17,58 @@ class MaximumSurface(SurfaceQuantity):
 
     Attributes:
         see `festim.SurfaceQuantity`
+        facet_meshtags: the facet meshtags of the mesh the field is defined on
     """
 
-    facet_meshtags: dolfinx.mesh.MeshTags
+    facet_meshtags: dolfinx.mesh.MeshTags | None = None
 
     @property
     def title(self):
         return f"Maximum {self.field.name} surface {self.surface.id}"
 
-    def compute(self):
+    def compute(
+        self,
+        u: dolfinx.fem.Function | None = None,
+        facet_meshtags: dolfinx.mesh.MeshTags | None = None,
+    ):
         """Computes the maximum value of the field on the defined surface subdomain, and
-        appends it to the data list."""
-        solution = self.field.post_processing_solution
-        entities = self.facet_meshtags.find(self.surface.id)
+        appends it to the data list.
+
+        Args:
+            u: the field the maximum is computed from. Defaults to
+                ``self.field.post_processing_solution``
+            facet_meshtags: the facet meshtags used to locate the facets of the
+                surface subdomain. Defaults to ``self.facet_meshtags``. For the
+                discontinuous problem these are the facet meshtags of the submesh
+                the field lives on (``VolumeSubdomain.ft``)
+
+        Raises:
+            ValueError: if no facet meshtags are available
+        """
+        solution = self.field.post_processing_solution if u is None else u
+        meshtags = self.facet_meshtags if facet_meshtags is None else facet_meshtags
+
+        if meshtags is None:
+            raise ValueError(
+                f"facet meshtags are required to compute {self.__class__.__name__}"
+            )
+
         if isinstance(solution, dolfinx.fem.Function):
             V = solution.function_space
         else:
             V = self.field.sub_function_space
         mesh = V.mesh
+        fdim = mesh.topology.dim - 1
 
+        entities = meshtags.find(self.surface.id)
+        mesh.topology.create_connectivity(fdim, mesh.topology.dim)
         dofs = dolfinx.fem.locate_dofs_topological(
-            V=V, entity_dim=mesh.topology.dim - 1, entities=entities
+            V=V, entity_dim=fdim, entities=entities
         )
+        values = solution.x.array[dofs]
 
-        self.value = mesh.comm.allreduce(np.max(solution.x.array[dofs]), op=MPI.MAX)
+        # a process may hold no dof of the surface at all, np.max would then raise
+        local_max = np.max(values) if values.size > 0 else -np.inf
+
+        self.value = mesh.comm.allreduce(local_max, op=MPI.MAX)
         self.data.append(self.value)
--- a/src/festim/exports/maximum_surface.py
+++ b/src/festim/exports/maximum_surface.py
@@ -28,30 +28,24 @@ class MaximumSurface(SurfaceQuantity):
 
     def compute(
         self,
+        facet_meshtags: dolfinx.mesh.MeshTags,
         u: dolfinx.fem.Function | None = None,
-        facet_meshtags: dolfinx.mesh.MeshTags | None = None,
     ):
         """Computes the maximum value of the field on the defined surface subdomain, and
         appends it to the data list.
 
         Args:
-            u: the field the maximum is computed from. Defaults to
-                ``self.field.post_processing_solution``
             facet_meshtags: the facet meshtags used to locate the facets of the
                 surface subdomain. Defaults to ``self.facet_meshtags``. For the
                 discontinuous problem these are the facet meshtags of the submesh
                 the field lives on (``VolumeSubdomain.ft``)
+            u: the field the maximum is computed from. Defaults to
+                ``self.field.post_processing_solution``
 
         Raises:
             ValueError: if no facet meshtags are available
         """
         solution = self.field.post_processing_solution if u is None else u
-        meshtags = self.facet_meshtags if facet_meshtags is None else facet_meshtags
-
-        if meshtags is None:
-            raise ValueError(
-                f"facet meshtags are required to compute {self.__class__.__name__}"
-            )
 
         if isinstance(solution, dolfinx.fem.Function):
             V = solution.function_space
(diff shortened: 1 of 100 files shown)


src/festim/hydrogen_transport_problem.py
```diff
@@ -2135,6 +2135,33 @@ def initialise_exports(self):
...
+                    volume = export.volume
+                    location = f"volume {volume.id}"
+                subdomains = export.field.subdomains
+                if isinstance(subdomains, list) and volume not in subdomains:
+                    raise ValueError(
+                        f"Cannot compute {export.title}: species "
+                        f"{export.field.name} is not defined in the volume subdomain "
+                        f"{volume.id} that {location} belongs to"
+                    )
+
```
ee-nn, line 2155: This just provides some error catching with nice-looking raises so could be deleted if desired

test/system_tests/test_multi_material.py
```diff
@@ -425,3 +425,84 @@ def test_all_cells_are_not_tagged():
...
+    )
+    bot_max = F.MaximumVolume(field=H, volume=bottom_volume)
+    top_min = F.MinimumVolume(field=H, volume=top_volume)
+    top_max = F.MaximumVolume(field=H, volume=top_volume)
+    bot_surf_min = F.MinimumSurface(
+        field=H, surface=bottom_surface, filename=f"{tmpdir}/bot_surf_min.csv"
+    )
+    bot_surf_max = F.MaximumSurface(field=H, surface=bottom_surface)
+    top_surf_min = F.MinimumSurface(field=H, surface=top_surface)
+    top_surf_max = F.MaximumSurface(field=H, surface=top_surface)
```
ee-nn, line 482: Might be a little redundant, perhaps only the bottom domain could be checked?

src/festim/exports/maximum_surface.py
```diff
@@ -17,28 +17,58 @@ class MaximumSurface(SurfaceQuantity):
...
+                the field lives on (``VolumeSubdomain.ft``)
+
+        Raises:
+            ValueError: if no facet meshtags are available
+        """
+        solution = self.field.post_processing_solution if u is None else u
+        meshtags = self.facet_meshtags if facet_meshtags is None else facet_meshtags
+
+        if meshtags is None:
+            raise ValueError(
```
RemDelaporteMathurin: Could be an assertion instead with a message if false

src/festim/exports/maximum_surface.py
```diff
@@ -17,28 +17,58 @@ class MaximumSurface(SurfaceQuantity):
...
+        entities = meshtags.find(self.surface.id)
+        mesh.topology.create_connectivity(fdim, mesh.topology.dim)
         dofs = dolfinx.fem.locate_dofs_topological(
-            V=V, entity_dim=mesh.topology.dim - 1, entities=entities
+            V=V, entity_dim=fdim, entities=entities
         )
+        values = solution.x.array[dofs]
+
+        # a process may hold no dof of the surface at all, np.max would then raise
+        local_max = np.max(values) if values.size > 0 else -np.inf
```
RemDelaporteMathurin, line 65: Good catch

src/festim/hydrogen_transport_problem.py
RemDelaporteMathurin: I think it's good to catch this early

test/system_tests/test_multi_material.py
RemDelaporteMathurin: I think it's ok to look out for everything

test/test_minimum_surface.py
```diff
@@ -42,3 +43,14 @@ def test_minimum_surface_export_compute_1D():
...
+
+def test_minimum_surface_raises_error_when_no_facet_meshtags():
+    """A ValueError is raised when no facet meshtags are available."""
+
+    my_export = F.MinimumSurface(
+        field=F.Species("H"), surface=F.SurfaceSubdomain1D(id=1, x=0)
+    )
+
+    with pytest.raises(ValueError, match="facet meshtags are required"):
+        my_export.compute()
```
RemDelaporteMathurin: If meshtags are always required why are they not a mandatory argument of compute?

RemDelaporteMathurin (changes requested):
Thanks for this @ee-nn !!! A first pass of comments before I go more in depth

test/test_minimum_surface.py
ee-nn: Good point, I can think about this more and update

ee-nn: Pushed an implementation, let me know what you think

src/festim/exports/minimum_volume.py
```diff
@@ -17,29 +17,56 @@ class MinimumVolume(VolumeQuantity):
...
 
     @property
     def title(self):
         return f"Minimum {self.field.name} volume {self.volume.id}"
 
-    def compute(self):
+    def compute(
+        self,
+        u: dolfinx.fem.Function | None = None,
+        volume_meshtags: dolfinx.mesh.MeshTags | None = None,
```
RemDelaporteMathurin, line 34: For a given problem, one would never invoke `compute()` with different volume mesthags and different `u`s right? I think they should be attributes instead and set inside `problem.initialise()` somewhere. (I think we have a `initialise_exports` method)

src/festim/exports/minimum_volume.py
```diff
@@ -17,29 +17,56 @@ class MinimumVolume(VolumeQuantity):
...
+        Args:
+            u: the field the minimum is computed from. Defaults to
+                ``self.field.post_processing_solution``
+            volume_meshtags: the cell meshtags used to locate the cells of the volume
+                subdomain. Defaults to ``self.volume_meshtags``. If neither is
+                available, the minimum is computed over the whole mesh ``u`` is
+                defined on. This is the expected behaviour for the discontinuous
+                problem, where each volume subdomain owns its own submesh.
+        """
+        solution = self.field.post_processing_solution if u is None else u
```
RemDelaporteMathurin, line 48: this could be a property

src/festim/exports/minimum_volume.py
```diff
@@ -17,29 +17,56 @@ class MinimumVolume(VolumeQuantity):
...
+        meshtags = self.volume_meshtags if volume_meshtags is None else volume_meshtags
```
RemDelaporteMathurin, line 49: this should be an attribute

src/festim/exports/minimum_volume.py
RemDelaporteMathurin, line 48: at the beginning of `compute` we should have an assertion to make sure that meshtags have been given

src/festim/exports/minimum_volume.py
```diff
@@ -17,29 +17,56 @@ class MinimumVolume(VolumeQuantity):
...
         else:
             V = self.field.sub_function_space
         mesh = V.mesh
-        mesh.topology.create_connectivity(mesh.topology.dim, mesh.topology.dim)
-        dofs = dolfinx.fem.locate_dofs_topological(
-            V=V, entity_dim=mesh.topology.dim, entities=entities
-        )
 
-        self.value = mesh.comm.allreduce(np.min(solution.x.array[dofs]), op=MPI.MIN)
+        if meshtags is None:
```
RemDelaporteMathurin, line 57: this should maybe be a property too like `self.is_submesh`.

Again, one day we'll get rid of the distinction between discontinous and non-discontinuous #1191

RemDelaporteMathurin (changes requested):

src/festim/exports/minimum_volume.py
ee-nn: This would only be needed for when we are not on a submesh right? If we have a property like `self.is_submesh` as you suggest for the `if meshtags is None` statement, then this would go in the `else` branch of the conditional block, no?

src/festim/exports/minimum_volume.py
ee-nn: Does this only apply to the `volume_meshtag` class and then we keep the `facet_meshtags` argument for the surface?

src/festim/exports/minimum_volume.py
RemDelaporteMathurin: I guess so yes!

RemDelaporteMathurin: or a method rather

src/festim/exports/minimum_volume.py
RemDelaporteMathurin: i think this would apply to both Volume quantities and Surface Quantities (if that's what you're asking)

ee-nn: Ah yeah that's what I was getting at, thanks 🙂

RemDelaporteMathurin (approved):

RemDelaporteMathurin:
Let's merge the stack #1216 first and then fix any conflicts here

ee-nn:
@RemDelaporteMathurin Ok I think this should be good to merge now

PR #1227, opened 2026-08-11
initial fix attempt for the flux conservation bug across internal interfaces

ee-nn:
## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🔨 Code refactoring (no functional changes, no API changes)
- [ ] 📝 Documentation update
- [ ] ✅ Test update (adding missing tests or correcting existing tests)
- [ ] 🔧 Build/CI configuration change

## Testing

- [ ] All existing tests pass locally (`pytest`)
- [ ] I have added new tests that prove my fix is effective or that my feature works

## Code Quality Checklist

- [ ] My code follows the code style of this project (Ruff formatted: `ruff format .`)
- [ ] My code passes linting checks (`ruff check .`)
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas

## Documentation

- [ ] I have updated the documentation accordingly (if applicable)
- [ ] I have added docstrings to new functions/classes following the project conventions

PR #1228, opened 2026-08-11
Codim: fix for the flux conservation bug across internal interfaces

RemDelaporteMathurin:
Migrated from #1227 so the branch lives on `festim-dev/FESTIM` rather than a fork.

Original work and commit authorship by @ee-nn — this PR contains the same commit (`f21898b`), unchanged.

## Description

Fix attempt for the flux conservation bug across internal interfaces in `HydrogenTransportProblemDiscontinuous`:

- Coupling sources on a manifold are integrated on the parent mesh, so their spatial coordinate must come from the parent mesh too — `ufl.SpatialCoordinate` of the manifold's submesh is silently wrong under the `+`/`-` restriction of an interior manifold.
- `ParticleFluxBC.create_value_fenics` now builds its value on the parent mesh instead of the target volume subdomain's submesh.

## Type of Change

## Testing

## Additional Notes

Work in progress — carried over as-is from #1227. Based on `rem/codim-fix-for-0.10` (#1221).

### DIFF
--- a/src/festim/hydrogen_transport_problem.py
+++ b/src/festim/hydrogen_transport_problem.py
@@ -1657,32 +1657,35 @@ class HydrogenTransportProblemDiscontinuous(HydrogenTransportProblem):
             name = f"{species.name}_{subdomain.id}"
             species.subdomain_to_post_processing_solution[subdomain].name = name
 
+    def source_integration_mesh(self, source) -> dolfinx.mesh.Mesh:
+        """The mesh the integral carrying ``source`` is assembled over.
+
+        A source involving only fields of a manifold subdomain is integrated over that
+        manifold's submesh (:meth:`subdomain_measure`); one coupling the manifold to
+        the bulk is a facet integral of the parent mesh (:meth:`coupling_measure`).
+        """
+        if self.is_manifold_self_source(source):
+            return source.volume.submesh
+        if source.volume in self.manifold_to_volumes:
+            return self.mesh.mesh
+        return source.species.subdomain_to_function_space[source.volume].mesh
+
     def convert_source_input_values_to_fenics_objects(self):
         """For each source create the value_fenics."""
         for source in self._unpacked_sources:
-            # create value_fenics for all F.ParticleSource objects
             if isinstance(source, _source.ParticleSource):
-                V = source.species.subdomain_to_function_space[source.volume]
-
-                # a self source on a manifold is integrated over its submesh, so its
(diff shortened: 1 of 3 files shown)


src/festim/hydrogen_transport_problem.py
```diff
@@ -1689,6 +1689,13 @@ def convert_source_input_values_to_fenics_objects(self):
                 else:
                     t = self.t
                     temperature = self.temperature_fenics
+                    if source.volume in self.manifold_to_volumes:
+                        # a coupling source is integrated on the parent mesh, so its
+                        # spatial coordinate has to be the parent mesh's as well --
+                        # ufl.SpatialCoordinate of the manifold's submesh is silently
+                        # wrong under the "+"/"-" restriction of an interior manifold.
+                        # only function_space.mesh is read on the up_to_ufl_expr path
+                        V = dolfinx.fem.functionspace(self.mesh.mesh, ("CG", 1))
```
RemDelaporteMathurin: this is going to create a new functionspace `V` for each source. Do we have a sense if that's ok or not?

RemDelaporteMathurin:
@ee-nn don't forget to add a test to prove that this is a bug

src/festim/hydrogen_transport_problem.py
ee-nn: It works, but you're right that it's probably not the best practice...I can change it

RemDelaporteMathurin: i don't think it's a problem. But you can use `self.V_CG_1`
https://github.com/festim-dev/FESTIM/blob/d40ae63c74ddc229258af78a72c4dc122b6f12e3/src/festim/hydrogen_transport_problem.py#L729

test/system_tests/test_codim1_coupling.py
```diff
@@ -189,6 +189,90 @@ def test_codim1_coupling_2d():
     assert all(r > 1.9 for r in rates_gam), rates_gam
 
 
+@pytest.mark.skipif(MPI.COMM_WORLD.size > 1, reason="serial only for now")
+def test_coupling_source_internal_subdomain():
+    """Emulate the repository MWE to verify codim-1 coupling terms."""
```
RemDelaporteMathurin, line 194: we need a better description since the "repository MWE" is bound to be removed

ee-nn (approved):
Good to go!

PR #1229, opened 2026-08-11
Codim: support for derived quantities

RemDelaporteMathurin:
### Summary

Makes FESTIM's derived quantities work with codimensional (manifold) subdomains. Three
places a quantity can be asked for once a codim-1 subdomain is in the mesh, none of
which the export layer could reach before:

| asked for | before | now |
|---|---|---|
| a **bulk** species on the facets Γ occupies, Γ on the domain boundary | `TypeError` at construction | ✅ converges at rate 2 |
| …same, Γ **interior** | **silently `0.0`** | ✅ exact `4/3` on the closed-form case |
| `TotalVolume` / `AverageVolume` over Γ itself | FFCx `UnboundLocalError: 't'` | ✅ exact |
| anything on Γ's **own boundary** (a codim-2 surface) | `KeyError` | ✅ matches the two-point BVP closed form |

Covers `SurfaceFlux`, `TotalSurface`, `AverageSurface`, `TotalVolume` and
`AverageVolume`. `VTXSpeciesExport` on Γ already worked and is unchanged, as is every
export on an ordinary surface — the codim-0 path is untouched.

```python
gamma = F.VolumeSubdomain(id=2, material=mat, dim=1, locator=...)   # a manifold
outlet = F.SurfaceSubdomain(id=4, dim=0, locator=...)               # its endpoint

exports = [
    F.SurfaceFlux(field=c_bulk, surface=gamma),    # exchange between bulk and manifold
    F.TotalVolume(field=c_gamma, volume=gamma),    # integral along the manifold
    F.SurfaceFlux(field=c_gamma, surface=outlet),  # what leaves at the manifold's end
]
```

### Related Issues

Related to #1208. Builds on the codim-1 support in #1216 — this is "dedicated exports on
Γ", which that PR listed as out of scope.

### Motivation and Context

A codim-1 subdomain is not much use if you cannot measure anything on it. The two
questions users actually have are *how much is exchanged between the bulk and the
manifold* and *how much leaves at the end of the manifold* (the outlet of the 1D fluid
in the pipe case), and neither was answerable.

Two findings are worth flagging to reviewers:

**1. The interior case was a silent wrong answer, not a crash.** `ds(Γ.id)` integrates to
exactly zero over interior facets, so a flux across an interior manifold would have
reported `0.0` with no warning had the type check not stopped it first. Interior Γ now
uses the `dS` coupling measure with the restriction of the side `field` lives on,
reusing `coupling_side` / `restriction_of` / `restrict` unchanged from #1216. This is the
behaviour `test_flux_on_an_interior_manifold_is_not_silently_zero` pins.

**2. Γ's boundary is not a codim-2 integral.** #1216 filed it as "a `ridge` / `dr`
integral, a different problem". It isn't: Γ's boundary is codim-1 *relative to Γ*, so it
is an ordinary exterior-facet `ds` on Γ's submesh — the endpoints of a line in 2D are the
facets of a 1D mesh — and `ufl.FacetNormal` works there. This is the same correction #1216
already made for Dirichlet dof location but did not carry over to assembly.

### Implementation

- `exports/surface_quantity.py` — the `surface` setter accepts a `VolumeSubdomain`, since
  a manifold is passed directly wherever a surface is expected.
- `HydrogenTransportProblem.initialise_exports` — and therefore **rejects** one outright,
  because manifolds exist only in `HydrogenTransportProblemDiscontinuous`. Without this,
  widening the setter would let a volume subdomain's id be looked up in the *facet* tags,
  silently reporting some other surface's value. `ChangeVar` delegates here, so one guard
  covers both.
- `helpers.py` — module-level `restrict(expression, restriction)`, so exports can restrict
  without reaching into the problem. `HydrogenTransportProblemDiscontinuous.restrict` now
  delegates to it.
- `SurfaceFlux` / `TotalSurface` / `AverageSurface` — optional `restriction` kwarg on
  `compute`, applied to the **whole** integrand. That is what makes `-D ∇c · n` pick the
  outward normal of its own side, so the sign convention comes out identical to an
  exterior surface: positive means leaving the subdomain the species lives on, and a flux
  through an interface reads `+J` on one side and `-J` on the other.
- `hydrogen_transport_problem.py` — `unindexed_coupling_measure` (factored out of
  `coupling_measure`), `export_volume_measure`, `_manifold_boundary_measure`, and
  `export_surface_context`, the single place the three cases are decided.
  `initialise_exports` now builds `D` on whichever mesh the integral lives on.

Both new measures carry a meshtag so the exports can index them by subdomain id as they
already do, and both are memoised — DOLFINx requires every integral of a compiled form to
share the *same* `subdomain_data` object, not merely an equal one.

## Type of Change

- [x] 📝 Documentation update
- [x] ✅ Test update (adding missing tests or correcting existing tests)

## Testing

- [x] All existing tests pass locally (`pytest`) — 977 passed, 5 skipped

`test/system_tests/test_codim1_exports.py` (10 tests) and five unit tests in
`test/test_codim1_subdomain.py`. **Every one was confirmed to fail with its own fix
reverted**, which for the interior case means reverting to a literal `0.0` against an
exact `4/3`.

Each asserted value is a closed form, not a regression baseline:

- `test_quantities_on_an_exterior_manifold` — totals and averages against the manufactured
  solution of `test_codim1_coupling.py`.
- `test_flux_on_an_exterior_manifold_matches_the_exchange` — the total flux on Γ is exactly
  zero for that MMS, so it is *also* checked weighted by `cos(πy)` against `D_Ω/2`,
  converging at rate > 1.8. A degenerate zero would otherwise pass with almost any sign or
  measure.
- `test_flux_on_an_interior_manifold_is_not_silently_zero` — the important one.
- `test_interior_manifold_flux_balance` — what Γ takes from one side it gives to the other.
  Passes under the naive implementation (`0 + 0 = 0`) but catches a wrong-side restriction,
  so it and the previous test are complementary.
- `test_quantities_on_the_boundary_of_a_manifold` — value and flux at Γ's endpoint,
  converging at rates 2 and 1.
- `test_outlet_flux_of_a_manifold_carrying_advection` — the pipe outlet against
  `-D dc/dy = v e^Pe / (e^Pe − 1)`, i.e. an endpoint flux whose exact value is *not* zero.
- Four raises: a manifold's own species on its own facets, a codim-0 volume as a surface, a
  locator matching nothing on Γ's boundary, and a volume subdomain as a surface in a problem
  class without manifolds.

Serial only for now, consistent with the rest of the codim-1 suite.

## Code Quality Checklist

- [x] My code passes linting checks (`ruff check .`) — clean on every file touched

## Documentation

New "Exports on a manifold" section in `docs/source/userguide/subdomains.rst` covering the
three positions, the side-inference rule and the sign convention; the limitations list
corrected accordingly. No new doctests (the additions are `code-block`, not `testcode`), and
I was unable to run `sphinx-build` locally — it is not in my environment — so the RST is
unverified beyond review.

## Breaking Changes

None. The `restriction` kwarg defaults to `None`, so every existing `compute` call is
unchanged, and the codim-0 path is byte-identical.

One behaviour *tightens*: passing a `VolumeSubdomain` as a `SurfaceQuantity`'s `surface`
used to be a `TypeError` at construction and is now a `TypeError` at `initialise()` in
problem classes without manifolds. Same outcome, later and with a clearer message.

## Additional Notes

**Known limitations, all pre-existing and none introduced here:**

- **`CustomQuantity` does not work on a manifold**, and is in fact broken in
  `HydrogenTransportProblemDiscontinuous` independently of codimension: it builds its kwargs
  as `{species.name: species.subdomain_to_post_processing_solution[volume] for species in
  self.species}`, which demands *every* species have a solution on the chosen subdomain, so
  any problem where a species does not live everywhere raises `KeyError`. Reproduced with two
  ordinary bulk subdomains and no manifold anywhere. Beyond that fix it also needs an API
  decision: the other exports separate `field` (which side to read) from `surface` (where to
  integrate), and `CustomQuantity` collapses both into one `subdomain`, so "the bulk flux
  across Γ" is not expressible. Worth its own issue.
- **`MinimumSurface` / `MaximumSurface` / `MinimumVolume` / `MaximumVolume`** read
  `field.post_processing_solution` and `self.facet_meshtags`, neither of which
  `HydrogenTransportProblemDiscontinuous` sets — `AttributeError` on `main` with no manifold
  involved. Also worth its own issue.
- **`SurfaceFlux` is diffusive-only**, as everywhere else in FESTIM; it already warns when
  advection terms are present. The pipe-outlet test asserts against the diffusive term only,
  and the docs say so.

### DIFF
--- a/docs/source/userguide/subdomains.rst
+++ b/docs/source/userguide/subdomains.rst
@@ -255,6 +255,55 @@ a float or as a callable of ``x`` and ``t`` rather than as a ready-made
 subdomain rather than sharing one between a reaction on a manifold and a reaction
 elsewhere. Both are raised rather than silently mis-assembled.
 
+Exports on a manifold
+---------------------
+
+Derived quantities can be asked for in three places once a manifold is in the mesh.
+
+**Over the manifold**, for its own species — a volume quantity, with ``volume`` set to
+the manifold. It is integrated over the manifold itself, so a
+:class:`festim.TotalVolume` on a line in a 2D mesh is a line integral:
+
+.. code-block:: python
+
+    F.TotalVolume(field=c_gamma, volume=gamma)
+    F.AverageVolume(field=c_gamma, volume=gamma)
+
+**On the facets the manifold occupies**, for a *bulk* species — a surface quantity, with
+the manifold passed where a surface subdomain normally goes. This is the exchange
+between the bulk and the manifold:
+
+.. code-block:: python
+
+    F.SurfaceFlux(field=c_bulk, surface=gamma)
+    F.TotalSurface(field=c_bulk, surface=gamma)
+
+On an interior manifold, which side the quantity is read on follows from ``field``,
+exactly as it does for the flux boundary conditions above — declare one export per side,
+each naming that side's species. The sign convention is the ordinary one: a positive
+flux leaves the subdomain the species lives on, so an exchange from the left bulk to the
+right one through the manifold reads positive on the left and negative on the right.
+
+Asking for a manifold's *own* species on its own facets raises: it has no flux across
+the manifold, and the quantity meant is the volume one above.
+
+**On the boundary of the manifold** — a surface quantity whose ``surface`` is the
+codim-2 :class:`festim.SurfaceSubdomain` described in the previous section. This is what
+gives the outlet flux of the pipe example:
+
+.. code-block:: python
+
+    outlet = F.SurfaceSubdomain(id=4, dim=0,
+                                locator=lambda x: np.isclose(x[0], L))
+    ...
+    exports=[F.SurfaceFlux(field=c_fluid, surface=outlet)]
+
+.. note::
+
+    :class:`festim.SurfaceFlux` computes the **diffusive** flux only. On a manifold
+    carrying an :class:`festim.AdvectionTerm` the advective part is not included, and
+    FESTIM warns.
+
 Limitations
 -----------
 
@@ -265,8 +314,12 @@ Limitations
   so the exchange with it would not be well posed.
 * Boundary conditions on the boundary of a manifold are limited to
   :class:`festim.FixedConcentrationBC`.
-* Dedicated exports on a manifold or on its boundary are not available;
-  :class:`festim.VTXSpeciesExport` on a manifold works.
+* Exports on and around a manifold are limited to the integral-based derived quantities
+  (:class:`festim.SurfaceFlux`, :class:`festim.TotalSurface`,
+  :class:`festim.AverageSurface`, :class:`festim.TotalVolume`,
+  :class:`festim.AverageVolume`) and :class:`festim.VTXSpeciesExport`. The minimum and
+  maximum quantities are not available in
+  :class:`festim.HydrogenTransportProblemDiscontinuous` at all, manifold or not.
 * Cartesian coordinates only.
 
 ----------
--- a/docs/source/userguide/subdomains.rst
+++ b/docs/source/userguide/subdomains.rst
@@ -318,7 +318,7 @@ Limitations
   (:class:`festim.SurfaceFlux`, :class:`festim.TotalSurface`,
   :class:`festim.AverageSurface`, :class:`festim.TotalVolume`,
   :class:`festim.AverageVolume`) and :class:`festim.VTXSpeciesExport`. The minimum and
-  maximum quantities are not available in
+  maximum quantities and :class:`festim.CustomQuantity` are not available in
   :class:`festim.HydrogenTransportProblemDiscontinuous` at all, manifold or not.
 * Cartesian coordinates only.
 
--- a/src/festim/exports/surface_quantity.py
+++ b/src/festim/exports/surface_quantity.py
@@ -58,11 +58,10 @@ class SurfaceQuantity(DerivedQuantity):
     @surface.setter
     def surface(self, value):
         # a manifold is a codim-1 VolumeSubdomain and is passed directly wherever a
-        # surface is expected. Whether it really is codim 1 needs the mesh, so it is
-        # checked at initialisation instead: HydrogenTransportProblemDiscontinuous in
-        # export_surface_context, and every other problem class in
-        # HydrogenTransportProblem.initialise_exports, which rejects volume subdomains
-        # outright since only that one class supports manifolds
+        # surface is expected. Checking whether it really is codim 1 needs the mesh, 
+        # so it's checked at initialisation: via export_surface_context in
+        # HydrogenTransportProblemDiscontinuous, and rejected outright in 
+        # HydrogenTransportProblem and its children since they don't support codim
         accepted = int | SurfaceSubdomain | VolumeSubdomain
         if not isinstance(value, accepted) or isinstance(value, bool):
             raise TypeError(
--- a/src/festim/exports/surface_quantity.py
+++ b/src/festim/exports/surface_quantity.py
@@ -3,6 +3,7 @@ from abc import abstractmethod
 from festim.exports.derived_quantity import DerivedQuantity
 from festim.species import Species
 from festim.subdomain.surface_subdomain import SurfaceSubdomain
+from festim.subdomain.volume_subdomain import VolumeSubdomain
 
 
 class SurfaceQuantity(DerivedQuantity):
@@ -10,7 +11,10 @@ class SurfaceQuantity(DerivedQuantity):
 
     Args:
         field: species for which the surface flux is computed
-        surface: surface subdomain
+        surface: surface subdomain. A codim-1 ``VolumeSubdomain`` (a manifold) may be
+            given instead, to compute the quantity on the facets it occupies -- for a
+            *bulk* species, since a manifold's own species has no flux across it. A
+            codim-2 ``SurfaceSubdomain`` computes it on the boundary of a manifold.
         filename: name of the file to which the surface flux is exported
 
     Attributes:
@@ -22,7 +26,7 @@ class SurfaceQuantity(DerivedQuantity):
     """
 
     field: Species
-    surface: SurfaceSubdomain
+    surface: SurfaceSubdomain | VolumeSubdomain
     filename: str | None
 
     t: list[float]
@@ -31,7 +35,7 @@ class SurfaceQuantity(DerivedQuantity):
     def __init__(
         self,
         field: Species | str,
-        surface: SurfaceSubdomain | int,
+        surface: SurfaceSubdomain | VolumeSubdomain | int,
         filename: str | None = None,
     ) -> None:
         super().__init__(filename=filename)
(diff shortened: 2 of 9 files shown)


docs/source/userguide/subdomains.rst
```diff
@@ -265,8 +314,12 @@ Limitations
   so the exchange with it would not be well posed.
 * Boundary conditions on the boundary of a manifold are limited to
   :class:`festim.FixedConcentrationBC`.
-* Dedicated exports on a manifold or on its boundary are not available;
-  :class:`festim.VTXSpeciesExport` on a manifold works.
+* Exports on and around a manifold are limited to the integral-based derived quantities
+  (:class:`festim.SurfaceFlux`, :class:`festim.TotalSurface`,
+  :class:`festim.AverageSurface`, :class:`festim.TotalVolume`,
+  :class:`festim.AverageVolume`) and :class:`festim.VTXSpeciesExport`. The minimum and
```
ee-nn, line 320: So would #1226 be merged after the codim stuff is integrated?

src/festim/helpers.py
```diff
@@ -120,6 +120,29 @@ def solution_on(species: "Species", subdomain: "VolumeSubdomain"):
...
+
+    Interior facet integrals need every discontinuous terminal restricted to one side
+    of the facet. Restricting the expression as a whole rather than each terminal keeps
+    a compound cross-mesh expression -- ``k * (c_bulk - c_manifold)``, or a flux
+    ``-D * grad(c) . n`` -- on a single side, which also picks the outward normal of
+    that side.
+
+    Args:
+        expression: the ufl expression to restrict
+        restriction: ``"+"``, ``"-"``, or ``None`` for an exterior facet integral, where
```
ee-nn, line 134: Perhaps this can make it more explicit that `None` is for an exterior facet integral only: `restriction: ``"+"`` or ``"-"``. ``None`` can also be set for an exterior facet integral`

docs/source/userguide/subdomains.rst
ee-nn: Might also be worth mentioning that `CustomQuantity` is not yet supported (maybe per your philosophy of small PRs this can be saved for a later date)

test/system_tests/test_codim1_exports.py
```diff
@@ -0,0 +1,392 @@
...
+            )
+        )
+        errors.append(abs(weighted - D_O / 2))
+
+    rates = [np.log(e0 / e1) / np.log(2) for e0, e1 in pairwise(errors)]
+    assert all(r > 1.8 for r in rates), rates
+
+
+@pytest.mark.skipif(MPI.COMM_WORLD.size > 1, reason="serial only for now")
+def test_flux_on_an_interior_manifold_is_not_silently_zero():
```
ee-nn, line 177: This test is very similar to the one needed for #1228, perhaps we could modify this test to ensure the correct flux both before and after applying exports?

test/system_tests/test_codim1_exports.py
```diff
@@ -0,0 +1,392 @@
...
+        ],
+        exports=list(exports),
+        temperature=500,
+        settings=F.Settings(atol=1e-12, rtol=1e-12, transient=False),
+    )
+    return model, (omega, gamma, right), (H_om, H_gam)
+
+
+def by_title(model):
+    return {e.title: e.data[-1] for e in model.exports}
```
ee-nn, line 108: Since this is a single line of code used in 2 places I don't really think it needs its own function

The same logic in is applied `test_interior_manifold_flux_balance()` without invoking this helper

test/system_tests/test_codim1_exports.py
```diff
@@ -0,0 +1,392 @@
...
+    from .test_codim1_interface import build
+
+    model, (_left, _right, gamma), (H_l, H_r, H_g) = build(n=20)
+    model.exports = [
+        F.SurfaceFlux(field=H_l, surface=gamma),
+        F.SurfaceFlux(field=H_r, surface=gamma),
+        F.TotalSurface(field=H_l, surface=gamma),
+        F.TotalSurface(field=H_r, surface=gamma),
+        F.TotalVolume(field=H_g, volume=gamma),
+    ]
```
ee-nn, line 201: Might be good to include `AverageSurface` here too

src/festim/exports/surface_quantity.py
```diff
@@ -53,8 +57,18 @@ def surface(self):
...
     @surface.setter
     def surface(self, value):
-        if not isinstance(value, int | SurfaceSubdomain) or isinstance(value, bool):
-            raise TypeError("surface should be an int or F.SurfaceSubdomain")
+        # a manifold is a codim-1 VolumeSubdomain and is passed directly wherever a
+        # surface is expected. Whether it really is codim 1 needs the mesh, so it is
+        # checked at initialisation instead: HydrogenTransportProblemDiscontinuous in
+        # export_surface_context, and every other problem class in
+        # HydrogenTransportProblem.initialise_exports, which rejects volume subdomains
+        # outright since only that one class supports manifolds
```
ee-nn:
```suggestion
        # surface is expected. Checking whether it really is codim 1 needs the mesh, 
        # so it's checked at initialisation: via export_surface_context in
        # HydrogenTransportProblemDiscontinuous, and rejected outright in 
        # HydrogenTransportProblem and its children since they don't support codim
```

src/festim/hydrogen_transport_problem.py
```diff
@@ -503,6 +506,21 @@ def initialise_exports(self):
...
+                # up in the facet tags, silently reporting some other surface's value
+                if isinstance(export, exports.SurfaceQuantity) and isinstance(
+                    export.surface, _subdomain.VolumeSubdomain
+                ):
+                    raise TypeError(
+                        f"volume subdomain {export.surface.id} was given as the "
+                        f"surface of a {type(export).__name__}. Only a codim-1 volume "
+                        "subdomain can be, and only in "
+                        "HydrogenTransportProblemDiscontinuous, the sole problem class "
+                        "supporting manifolds."
```
ee-nn:
```suggestion
                        f"volume subdomain {export.surface.id} was given as the "
                        f"surface of a {type(export).__name__}. Co-dim volume subomains"
                        "are only supported by HydrogenTransportProblemDiscontinuous"
```

ee-nn (commented):
First time writing a code review! Most of the changes were fairly straightforward to me also, most of the comments I have are cosmetic

docs/source/userguide/subdomains.rst
RemDelaporteMathurin: Yes probably, it is a separate issue. 

> `CustomQuantity` is not yet supported

Yes I am aware of this current limitation. I have a couple ideas on how to fix it. Yes it's better to keep it for a separate PR

src/festim/helpers.py
RemDelaporteMathurin: The current already reads "`None` for an exterior facet integral`", is that not explicit enough you think?

test/system_tests/test_codim1_exports.py
RemDelaporteMathurin: That's what I thought too! I don't see how before/after applying exports would change the flux though

src/festim/helpers.py
ee-nn: Upon re-reading I think it's fine as-is, when I initially read and made that comment I read it as "all 3 of these conditions are for the case of an exterior facet integral"

test/system_tests/test_codim1_exports.py
ee-nn: If exports are updated or modified incorrectly I could see it potentially cropping up, for example if the wrong expression or measure is passed perhaps?

ee-nn (approved):
Nice! Thanks for giving me the opportunity to learn how to code review!

PR #1231, opened 2026-08-13
Drift terms

### LINKED ISSUE #973: Soret effect

### LINKED ISSUE #1080: Correct advection term

RemDelaporteMathurin:
## Description

To be merged after #1229 (stack).
Replaces #1081

### Summary

Adds **drift terms**: transport driven by a gradient other than the species' own. All of
them add a velocity to the flux, and differ only in what sets it:

```
J = -D grad(c) + c v
```

| term | drift velocity | driving field |
|---|---|---|
| `F.AdvectionTerm` *(existing)* | given directly | a moving fluid |
| **`F.SoretTerm`** | `-D Q*/(k_B T²) grad(T)` | temperature |
| **`F.ElectromigrationTerm`** | `-z D/(k_B T) grad(phi)` | electric potential |

They share one assembly, in **divergence form** `div(c v)`, and one new list on the
problem, `drift_terms`. `AdvectionTerm` is folded onto the same base class.

Also here, because the divergence form makes them necessary:

- **`F.OutflowBC`** — marks a surface the flow leaves through, so drift carries the
  species out of the domain. This is the `+ (v·n) u w ds` term of #1081, arrived at from
  the other direction; see *Relationship to #1081* below.
- **`F.SurfaceFlux` now reports the total flux** `-D grad(c)·n + c v·n`. The
  "advection terms are not accounted for" warning is gone.

```python
model.drift_terms = [
    F.SoretTerm(species=H, Q_star=0.2, subdomain=vol),
    F.ElectromigrationTerm(species=OD, charge=1, potential=phi, subdomain=membrane),
]
model.boundary_conditions = [
    F.FixedConcentrationBC(subdomain=inlet, value=1.0, species=H),
    F.OutflowBC(subdomain=outlet, species=H),
]
```

### Related Issues

Stacked on #1229. Replaces #1081.

Fixes #973, #1080 

### Motivation and Context

**Soret is a v1 → v2 regression.** `docs/source/theory.rst` documents the effect with its
flux equation, and `docs/source/userguide/settings.rst` listed `soret` as a settings
option — but nothing in `src/festim/` implemented it (`grep -i soret src/` returned
nothing, and `Settings` has no such parameter). FESTIM 1 had it, including for
cylindrical and spherical coordinates. Thermal gradients in plasma-facing components are
large enough to redistribute tritium, and TMAP7/TMAP8 both have Soret, so any
thermal-gradient benchmark was out of reach.

**Electromigration was unreachable**, which rules out proton-conducting-ceramic
electrochemistry (e.g. TMAP8's val-2l case), electrochemical hydrogen pumping, and
Devanathan–Stachurski permeation cells — the standard experimental method for measuring
H diffusivity in metals.

**The existing `AdvectionTerm` was non-conservative and volume-only.** It assembled
`v·grad(c)`, which equals `div(c v)` only where `div(v) = 0`, and it contributed nothing
at the boundary — so a flux BC constrained only the diffusive part.

Doing all three as one term is what makes this a small change rather than three: the same
weak form covers stress-driven diffusion later, which is the remaining member of the
family.

### Relationship to #1081

#1081 identified exactly the same missing term: *"an additional term may be required from
integration by parts: + (v·n) u w ds"*, and added `boundary_conditions/outflow_bc.py`
under the same name. This PR arrives at it from the conservative-form direction — the
term is required precisely because `div(c v)` is integrated by parts — and adds the tests
and docs that PR was missing, plus support for the boundary of a codim-1 manifold.

One deliberate difference: #1081 added a separate `exports/advection_flux.py`. Here the
advective contribution is folded into `SurfaceFlux` instead, so that one export reports
the flux the model actually conserves. Happy to revisit if a separate export is wanted.

## Type of Change

- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] 📝 Documentation update
- [x] ✅ Test update (adding missing tests or correcting existing tests)

## Implementation

- **`src/festim/drift.py`** (new) — `DriftTermBase`, `SoretTerm`, `ElectromigrationTerm`,
  and the shared `drift_form`. Coefficients (`Q_star`, `potential`) go through
  `helpers.Value` with `up_to_ufl_expr=True`, so they get the usual `x`/`t`/`T`
  dependence for free.
- **`advection.py`** — `AdvectionTerm` subclasses `DriftTermBase` and keeps
  `VelocityField`. The new terms cannot reuse it: their velocities are ufl expressions
  built from `T`, `D` and a gradient, and `VelocityField` accepts only a `fem.Function`
  or a callable of `t`.
- **The metric factors are new code.** FESTIM multiplies the equation by `m` (`r`
  cylindrical, `r²` spherical) and tests with `w/m`; those cancel in `v·grad(c)` but not
  in the divergence form, which needs `-m c v·grad(w/m)`.
- Wired into all three problem classes, including
  `HydrogenTransportProblemDiscontinuousChangeVar`, **which ignored `advection_terms`
  entirely and silently** before this.
- A drift term whose driving gradient vanishes is skipped rather than assembled — a form
  built from an all-zero integrand has no integration domain for FFCx — and **warns**,
  naming the input at fault. `temperature=lambda t: ...` is the case this exists for: it
  varies in time, not in space.

## Testing

- [x] All existing tests pass locally (`pytest`) — **1021 passed, 5 skipped**

44 new tests: 30 unit (`test/test_drift.py`), 11 system (`test/system_tests/test_drift.py`),
3 in a worked example (`test/system_tests/test_counterflow.py`). The physical ones assert
closed forms, not regression baselines.

| test | asserts | result |
|---|---|---|
| Soret equilibrium | `c = A exp(Q*/k_B T)` on a no-flux slab | 6e-7 relative |
| Electromigration equilibrium | the Boltzmann profile `c = A exp(-z phi/k_B T)` | 1e-5 relative |
| Inventory conservation | nothing leaves a domain with no BCs at all | 1e-9 |
| `OutflowBC` semantics | wall gives `c = e^(2x)` = 7.389, outflow gives `c = 1` | 7.389087 vs `e²` = 7.389056 |
| Counterflow mass balance | the four channel-end fluxes sum to zero | **1.8e-15** |
| `SurfaceFlux` includes drift | total flux → 0 at equilibrium while the diffusive part is O(0.4) | ratio 1e-3, first-order convergent |

**The discriminating test** is `test_advection_mms_with_a_non_solenoidal_velocity`: an MMS
whose source is built from `div(c v)` with `div(v) = 1`. It converges at rate 2 under the
divergence form and stalls under `v·grad(c)`. Worth knowing that **every pre-existing
advection test uses a divergence-free velocity with Dirichlet data on the whole
boundary**, so all of them pass under either form and none would have caught a regression
here. `test_drift_form_is_the_divergence_form` pins the assembled form directly as a
cheap unit-level guard.

**Worked example** (`test_counterflow.py`): a counter-current membrane contactor — a 2D
solid separator with a 1D fluid channel running up each face as a codim-1 manifold, one
flowing up and one down, exchanging through `J = k(c_fluid − c_wall)`. The four
channel-end fluxes balance to 1.8e-15, and counter-current recovers 3.4% more into the
sweep than co-current (mesh-converged: 3.371 / 3.370 / 3.369 % at n = 60/120/240).

## Code Quality Checklist

- [x] My code passes linting checks (`ruff check .`) — clean on every file touched

## Documentation

New `docs/source/userguide/drift.rst`; the Soret section of `theory.rst` generalised to
drift with electromigration added; `festim.drift` added to the API index. `soret` removed
from the settings list in `settings.rst`, where it had been advertised as an option that
does not exist in v2. As in #1229 I could not run `sphinx-build` in a full environment
(the docs env here lacks `io4dolfinx`), so the HTML builds clean and the RST is otherwise
unverified beyond review.

## Breaking Changes

**1. `AdvectionTerm` is now assembled in divergence form.** Results are unchanged if the
velocity field is divergence-free (as an incompressible flow is) **and** every boundary
the flow crosses carries a boundary condition. If the flow leaves through an untagged
boundary, that boundary is now a closed end and the species backs up against it — add an
`F.OutflowBC` there. Documented with a warning in both the user guide and the theory
guide.

This bit two tests in this repo, both of which had a manifold outlet with no condition on
it; both now declare one. `test_multi_material_with_advection` also changes answers — it
has untagged boundary either side with a horizontal velocity — but it asserts nothing, so
it passes either way. Worth deciding whether to give it an `OutflowBC` and some
assertions.

**2. `advection_terms` is deprecated** in favour of `drift_terms`, with a
`DeprecationWarning`. It is deliberately **not** a plain alias: it appends to
`drift_terms` rather than replacing it, so that

```python
model = F.HydrogenTransportProblem(drift_terms=[soret])
model.advection_terms = [advection]   # soret survives
```

does not silently drop a Soret term and leave the model solving different physics.
Re-assigning replaces the advection entries rather than accumulating them, and the getter
returns just that subset. The five test files that used the old name have been migrated,
so the shim is exercised deliberately rather than incidentally.

## Additional Notes

**Known limitations:**

- **No stabilisation.** FESTIM has no SUPG, so at a large cell Péclet number the Galerkin
  form oscillates. Documented; a runtime warning on cell Péclet would be a reasonable
  follow-up.
- **Stress-driven diffusion is deferred** — it is the same term with
  `v = D V_H/(RT) grad(sigma_h)`, but needs a hydrostatic-stress field API FESTIM has no
  story for. `DriftTermBase` is shaped so it drops in later.
- **The electric potential is prescribed, not solved.** Enough for the PCC cases, which
  impose a linear potential across the membrane, but there is no Poisson/current
  conservation solve.
- Drift is not added to `CustomQuantity`, which #1229 already documents as broken in the
  discontinuous problem for unrelated reasons.
- The `ChangeVar` wiring is included because that class silently ignored drift terms, but
  it is low-value given the plan to remove the class — easy to drop in review.

### DIFF
--- a/src/festim/advection.py
+++ b/src/festim/advection.py
@@ -4,13 +4,23 @@ import basix
 import ufl
 from dolfinx import fem
 
+from festim.drift import DriftTermBase
 from festim.helpers import Value, nmm_interpolate
 from festim.species import Species
 from festim.subdomain import VolumeSubdomain
 
 
-class AdvectionTerm:
+class AdvectionTerm(DriftTermBase):
     """Advection term class.
+
+    Transport by a velocity field given directly, as opposed to one built from a driving
+    gradient (see :class:`festim.SoretTerm`, :class:`festim.ElectromigrationTerm`).
+
+    Assembled in divergence form, ``-div(c v)``, like every drift term. This differs
+    from the ``v . grad(c)`` FESTIM assembled before: the two agree in the interior
+    wherever ``div(v) == 0``, which holds for an incompressible flow, but only the
+    divergence form carries species out through a boundary that has no flux condition on
+    it -- an outlet, typically. See :func:`festim.drift.drift_form`.
 
     args:
         velocity: the velocity field or function
@@ -33,9 +43,8 @@ class AdvectionTerm:
         subdomain: VolumeSubdomain,
         species: Species,
     ):
+        super().__init__(subdomain=subdomain, species=species)
         self.velocity = velocity
-        self.subdomain = subdomain
-        self.species = species
 
     @property
     def velocity(self):
@@ -58,37 +67,14 @@ class AdvectionTerm:
         else:
             raise TypeError(err_message)
 
-    @property
-    def subdomain(self):
-        return self._subdomain
+    def convert_inputs(self, function_space, t=None, temperature=None):
+        self.velocity.convert_input_value(function_space=function_space, t=t)
 
-    @subdomain.setter
-    def subdomain(self, value):
-        if value is None:
-            self._subdomain = value
-        elif isinstance(value, VolumeSubdomain):
-            self._subdomain = value
-        else:
-            raise TypeError(
-                f"Subdomain must be a festim.Subdomain object, not {type(value)}"
-            )
+    def time_dependent_inputs(self):
+        return [self.velocity]
 
-    @property
-    def species(self) -> list[Species]:
-        return self._species
-
-    @species.setter
-    def species(self, value):
-        if not isinstance(value, list):
-            value = [value]
-        # check that all species are of type festim.Species
-        for spe in value:
-            if not isinstance(spe, Species):
-                raise TypeError(
-                    f"elements of species must be of type festim.Species not "
-                    f"{type(spe)}"
-                )
-        self._species = value
+    def drift_velocity(self, D, temperature):
+        return self.velocity.fenics_object
 
 
 class VelocityField(Value):
--- a/src/festim/advection.py
+++ b/src/festim/advection.py
@@ -149,11 +149,14 @@ class VelocityField(Value):
                 )
 
         # create vector function space and function
+        # NOTE: the shape is the *geometric* dimension, not the topological one: on a
+        # codim-1 subdomain the mesh is a manifold (eg. a line in 2D) whose cells are
+        # 1D but whose points, and therefore velocities and gradients, are ambient
         v_cg = basix.ufl.element(
             "Lagrange",
             function_space.mesh.topology.cell_name(),
             1,
-            shape=(function_space.mesh.topology.dim,),
+            shape=(function_space.mesh.geometry.dim,),
         )
         self.vector_function_space = fem.functionspace(function_space.mesh, v_cg)
         self.fenics_object = fem.Function(self.vector_function_space)
--- a/src/festim/boundary_conditions/outflow_bc.py
+++ b/src/festim/boundary_conditions/outflow_bc.py
@@ -5,22 +5,11 @@ from festim.subdomain.surface_subdomain import SurfaceSubdomain
 class OutflowBC:
     r"""Lets a drift term carry the species out through a boundary.
 
-    Drift terms are assembled in divergence form, so the boundary term they leave
-    behind makes the flux conditions of the problem constrain the **total** flux. A
-    boundary with no condition on it is then a wall: nothing leaves, and the drift is
-    balanced by back-diffusion. That is what you want at an actual wall, and wrong at an
-    outlet, where whatever the flow reaches should be carried out of the domain.
-
-    This boundary condition adds
-
-    .. math::
-
-        + \int_{\partial\Omega} c\, (\mathbf{v} \cdot \mathbf{n})\, w
-
-    which cancels exactly that term, so the natural condition on this surface becomes
-    zero *diffusive* flux, :math:`D \nabla c \cdot \mathbf{n} = 0` -- the standard
-    "do-nothing" outflow condition of advection-diffusion. The species leaves at the
-    rate the drift carries it.
+    Adds :math:`+ \int_{\partial\Omega} c\, (\mathbf{v} \cdot \mathbf{n})\, w`,
+    cancelling the boundary term the divergence form leaves behind, so the natural
+    condition on this surface becomes zero *diffusive* flux -- the standard "do-nothing"
+    outflow of advection-diffusion. Without it an untagged outlet is a closed end and
+    the species backs up against it. See :ref:`outflow` in the user guide.
 
     It is a no-op on a surface where no drift term acts on ``species``.
 
--- a/src/festim/boundary_conditions/outflow_bc.py
+++ b/src/festim/boundary_conditions/outflow_bc.py
@@ -1,0 +1,62 @@
+from festim.species import Species
+from festim.subdomain.surface_subdomain import SurfaceSubdomain
+
+
+class OutflowBC:
+    r"""Lets a drift term carry the species out through a boundary.
+
+    Drift terms are assembled in divergence form, so the boundary term they leave
+    behind makes the flux conditions of the problem constrain the **total** flux. A
+    boundary with no condition on it is then a wall: nothing leaves, and the drift is
+    balanced by back-diffusion. That is what you want at an actual wall, and wrong at an
+    outlet, where whatever the flow reaches should be carried out of the domain.
+
+    This boundary condition adds
+
+    .. math::
+
+        + \int_{\partial\Omega} c\, (\mathbf{v} \cdot \mathbf{n})\, w
+
+    which cancels exactly that term, so the natural condition on this surface becomes
+    zero *diffusive* flux, :math:`D \nabla c \cdot \mathbf{n} = 0` -- the standard
+    "do-nothing" outflow condition of advection-diffusion. The species leaves at the
+    rate the drift carries it.
+
+    It is a no-op on a surface where no drift term acts on ``species``.
+
+    Args:
+        subdomain: the surface subdomain the species flows out through. On a
+            codimensional problem this may be the boundary of a manifold -- the outlet
+            of a 1D fluid, for instance
+        species: the species carried out
+
+    Attributes:
+        subdomain: the surface subdomain the species flows out through
+        species: the species carried out
+
+    Examples:
+        .. code-block:: python
+
+            import festim as F
+
+            F.OutflowBC(subdomain=outlet, species=H)
+    """
+
+    subdomain: SurfaceSubdomain
+    species: Species
+
+    def __init__(self, subdomain: SurfaceSubdomain, species: Species):
+        self.subdomain = subdomain
+        self.species = species
+
+    @property
+    def species(self) -> Species:
+        return self._species
+
+    @species.setter
+    def species(self, value):
+        if not isinstance(value, Species):
+            raise TypeError(
+                f"species must be of type festim.Species, not {type(value)}"
+            )
+        self._species = value
--- a/src/festim/drift.py
+++ b/src/festim/drift.py
@@ -11,6 +11,7 @@ Different physics differ only in what sets ``v``: a fluid carries the species al
 """
 
 import warnings
+from abc import ABC, abstractmethod
 
 import ufl
 from dolfinx import fem
@@ -22,7 +23,7 @@ from festim.species import Species
 from festim.subdomain import VolumeSubdomain
 
 
-class DriftTermBase:
+class DriftTermBase(ABC):
     """Base class for drift terms.
 
     Subclasses supply :meth:`drift_velocity`; everything else -- validation, the
@@ -114,6 +115,7 @@ class DriftTermBase:
         """The user-given coefficients of this term that may depend on time."""
         return []
 
+    @abstractmethod
     def drift_velocity(self, D, temperature):
         """The drift velocity of this term, as a ufl expression.
 
@@ -125,7 +127,6 @@ class DriftTermBase:
             the velocity, a vector-valued ufl expression. ``ufl.zero`` when the driving
             gradient vanishes, in which case the term is skipped.
         """
-        raise NotImplementedError
 
     def inert_reason(self) -> str:
         """Why this term's velocity came out identically zero, for the user.
--- a/src/festim/drift.py
+++ b/src/festim/drift.py
@@ -273,9 +273,15 @@ def is_zero_velocity(velocity) -> bool:
     """Whether a drift velocity is identically zero.
 
     ``ufl.grad`` of a spatially constant field is a ``Zero`` at construction, and that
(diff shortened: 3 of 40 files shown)


src/festim/drift.py
```diff
@@ -0,0 +1,371 @@
+"""Drift terms: transport driven by a gradient other than the species' own.
+
+A drift term adds a velocity to the species flux::
+
+    J = -D grad(c) + c v
+
+Different physics differ only in what sets ``v``: a fluid carries the species along
+(:class:`festim.AdvectionTerm`), a temperature gradient drives thermodiffusion
```
ee-nn, line 8: Should the contents of the advection term file be moved here? I feel having all the drift term classes in a single file would make the most sense, especially since `AdvectionTerm` is already mentioned here as if it is included

src/festim/drift.py
ee-nn, line 1: Might also be nice to include a `CustomDriftTerm` class where the user can simply pass in a function that they would like added to the transport equation, if it doesn't fit within the existing drift terms that are offered

src/festim/drift.py
```diff
@@ -0,0 +1,371 @@
...
+from dolfinx import fem
+
+from festim import k_B
+from festim.helpers import Value
+from festim.mesh import CoordinateSystem
+from festim.species import Species
+from festim.subdomain import VolumeSubdomain
+
+
+class DriftTermBase:
```
ee-nn: Would it make more sense for this class to inherit from ABC? That's how it's done in other parts of the codebase (eg. the `DerivedQuantity` base class). That way `drift_velocity` in this base class could become an `@abstractmethod` and pass instead of raising a `NotImplementedError`

ee-nn (commented):
A few high-level comments

src/festim/drift.py
RemDelaporteMathurin: I guess it should. The reason to keep it separate for now is that there still are some differences with other drift terms. For example, the use of `VelocityField` and the fact that the problem class (for now) has both `advection_terms` and `drift_terms`

src/festim/drift.py
RemDelaporteMathurin: In this case, the idea is that a user would make a class inheriting from `DriftTermBase` and implement the right methods

src/festim/drift.py
RemDelaporteMathurin: Yes I think it could and that's also something @jhdark suggested. I'll look into it

src/festim/advection.py
```diff
@@ -149,11 +135,14 @@ def convert_input_value(
                 )
 
         # create vector function space and function
+        # NOTE: the shape is the *geometric* dimension, not the topological one: on a
+        # codim-1 subdomain the mesh is a manifold (eg. a line in 2D) whose cells are
+        # 1D but whose points, and therefore velocities and gradients, are ambient
         v_cg = basix.ufl.element(
             "Lagrange",
             function_space.mesh.topology.cell_name(),
             1,
```
jhdark, line 144: We should also allow for DG 0 velocity fields for imports from openFOAM

RemDelaporteMathurin: Probably.

1. Since this PR now enforces that we use the conservative form for the advection, the DG0 to CG1 interpolation making the velocity field non divergence-free is no longer a problem
2. But we still want to not modify the data too much! I feel like this is a separate problem and can be tacked by a smaller, more focused PR

What do you think?

jhdark: Agreed on both counts, and a separate focused PR for the DG0 handling works for me.

One concrete piece from #1081 to carry over whenever that happens, since I am closing that PR: `VelocityField.convert_input_value` there short-circuits when the given function already lives on the target mesh

```python
if vel.function_space.mesh == function_space.mesh:
    self.fenics_object = vel
else:
    ...  # build the CG1 vector space and nmm_interpolate
```

As it stands here every velocity is copied into a freshly built CG1 vector space even when the input is already CG1 on that same mesh, which is both an avoidable interpolation and the point at which a DG0 field from foam2dolfinx loses its data. In the discontinuous problem the interpolation onto each subdomain's submesh is genuinely needed, so the short-circuit only buys anything in the continuous case, but that is the common one for an OpenFOAM velocity.

jhdark:
I am closing #1081 in favour of this PR, having checked that everything in it is covered here. One difference is worth a decision rather than a silent drop, so raising it here.

#1081 added a separate `AdvectionFlux` export and left `SurfaceFlux` reporting the diffusive flux only. This PR does the opposite: `SurfaceFlux` reports the total, which I agree is the right default, since it is the quantity the model actually conserves.

How we could fix this:
An opt-out on the export, e.g. `F.SurfaceFlux(..., include_drift=False)`, defaulting to `True` so the conservative behaviour stays the default.

src/festim/boundary_conditions/outflow_bc.py
```diff
@@ -0,0 +1,62 @@
...
+    .. math::
+
+        + \int_{\partial\Omega} c\, (\mathbf{v} \cdot \mathbf{n})\, w
+
+    which cancels exactly that term, so the natural condition on this surface becomes
+    zero *diffusive* flux, :math:`D \nabla c \cdot \mathbf{n} = 0` -- the standard
+    "do-nothing" outflow condition of advection-diffusion. The species leaves at the
+    rate the drift carries it.
+
+    It is a no-op on a surface where no drift term acts on ``species``.
```
RemDelaporteMathurin, line 14: a lot of info - maybe this should live in the docs?

src/festim/drift.py
```diff
@@ -0,0 +1,371 @@
...
+
+    Returns:
+        the ufl form to add to the problem's formulation, or ``None`` when the velocity
+        is identically zero
+
+    Raises:
+        NotImplementedError: for an unknown coordinate system
+    """
+    if is_zero_velocity(velocity):
+        return None
```
RemDelaporteMathurin: This is done at initialise. In the scenario where the velocity is zero at the beginning but increases later on then it will never be added to the form, leading to wrong results!

We should simply add it everytime.

The warning above could stay but it should be more specific ("the drift velocity is zero at t=0")

src/festim/boundary_conditions/outflow_bc.py
RemDelaporteMathurin: or at least have a version of this in the docs

src/festim/hydrogen_transport_problem.py
```diff
@@ -613,6 +664,48 @@ def _get_temperature_field_as_function(self) -> dolfinx.fem.Function:
...
+
+        velocities = []
+        for term in self.drift_terms:
+            if term.subdomain is not volume:
+                continue
+            if field not in term.species:
+                continue
+            D = volume.material.get_diffusion_coefficient(mesh, temperature, field)
+            velocity = term.drift_velocity(D=D, temperature=temperature)
+            if not _drift.is_zero_velocity(velocity):
```
RemDelaporteMathurin, line 702: Same comment as above. This is a bit dangerous if the drift velocity is zero initially but later grows

src/festim/hydrogen_transport_problem.py
```diff
@@ -860,11 +953,17 @@ def create_sources_from_reactions(self):
             self._unpacked_sources += reaction.create_sources()
 
     def convert_advection_term_to_fenics_objects(self):
```
RemDelaporteMathurin: Should we rename this method to make it not focussed on advection?

src/festim/hydrogen_transport_problem.py
```diff
@@ -985,18 +1084,39 @@ def create_formulation(self):
...
+            vol = drift_term.subdomain
+            for species in drift_term.species:
+                D = vol.material.get_diffusion_coefficient(
+                    self.mesh.mesh, self.temperature_fenics, species
+                )
+                velocity = drift_term.drift_velocity(
+                    D=D, temperature=self.temperature_fenics
+                )
+                if _drift.warn_if_no_effect(drift_term, species, velocity):
+                    continue
```
RemDelaporteMathurin: Same comment as above

src/festim/hydrogen_transport_problem.py
```diff
@@ -985,18 +1084,39 @@ def create_formulation(self):
...
+                self.formulation += _drift.drift_form(
+                    concentration=species.solution,
+                    test_function=species.test_function,
+                    velocity=velocity,
+                    dx=self.dx(vol.id),
```
RemDelaporteMathurin, line 1104: Does that work with the codim1 stuff?

src/festim/hydrogen_transport_problem.py
```diff
@@ -985,18 +1084,39 @@ def create_formulation(self):
...
-                advection_term = ufl.inner(ufl.dot(ufl.grad(conc), vel), v) * self.dx(
-                    adv_term.subdomain.id
+        for bc in self.boundary_conditions:
+            if isinstance(bc, boundary_conditions.OutflowBC):
+                velocity = self.drift_velocity_in(
+                    field=bc.species,
+                    volume=self.volume_subdomain_of_surface(bc.subdomain),
                 )
-                self.formulation += advection_term
+                if velocity is not None:
```
RemDelaporteMathurin, line 1115: I think we could remove this (see comments above)

src/festim/hydrogen_transport_problem.py
RemDelaporteMathurin: it's correctly covered in the Discontinuous class no problem

src/festim/hydrogen_transport_problem.py
```diff
@@ -1721,13 +1840,21 @@ def convert_source_input_values_to_fenics_objects(self):
                 )
 
     def convert_advection_term_to_fenics_objects(self):
```
RemDelaporteMathurin: Rename method

RemDelaporteMathurin:
Reviewed with @jhdark and @ee-nn and approved

PR #1232, opened 2026-08-18
Derived quantities revamp

ee-nn:
## Description

This provides an shift in how derived quantities are handled. With the new codim functionality coming with the stack at #1216, the previous clear distinction between `Surface` and `Volume` derived quantities starts to break down. This PR creates a single class for each derived quantity export: `Total`, `Average`, `Minimum`, etc.  

This is currently posted as a draft for visibility and will be updated as the other pending PRs are merged to the code. 

This will include backwards compatibility. 

### Related Issues
Designed to come after #1231 is applied. 

## Type of Change

- [X] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] 🔨 Code refactoring (no functional changes, no API changes)
- [ ] 📝 Documentation update

## Testing

All the tests pass at the moment, but I have yet to go through them in detail to ensure that they all make the most sense. 

## Code Quality Checklist

## Documentation

- [ ] I have updated the documentation accordingly (if applicable) --> needs some review

## Breaking Changes

PR does away with the distinction between flux and surface quantities; a user can pass either a surface, a regular subdomain, or a codimensional subdomain to a single derived quantity operation and is able to handle the differences internally. The legacy derived quantities have been kept, but now give a deprecation warning.


ee-nn:
Now that I think about it, it could also be nice to have a similar set of operations for fluxes; instead of just having `SurfaceFlux` we could have `TotalFlux`, `MinimumFlux`, `MaximumFlux`, etc

Also maybe this is straying too far form the topic of this PR but I was thinking that it could also be nice to  be able to measure and export a `SurfaceFlux` quantity through an "imaginary" surface or edge. For example, you have a unit cube and you are interested in the flux across the middle of the cube. Traditionally you would need to have some sort of subdomain object there to track this, but an imaginary surface could simply specify a geometry and then track a flux through said geometry without needing meshtagging, subdomain assignment, problem formulation, etc.

RemDelaporteMathurin:
> an imaginary surface could simply specify a geometry and then track a flux through said geometry without needing meshtagging, subdomain assignment

In order to compute say a flux in dolfinx you always need: a formulation that you integrate <-- a measure (ds)  <-- a subdomain <-- meshtags

Issue #897, opened 2024-10-25
[BUG] Mobile species defined in only one subdomain, but with a common interface for some other species


jorgensd:
When defining a mobile species that lives only in a subdomain then festim tries to assign a concentration discontinuity interface condition on this species. Causing an error because that species doesn't exist on the other side.

In this case, an homogeneous neuman BC should be applied (no-flux). A simple way to do this is to just check for a given interface that the species lives in both subdomains. If not, then do nothing.

## MWE

Run this code on `fenicsx` 

```python
my_model = F.HTransportProblemDiscontinuous()
my_model.mesh = F.Mesh1D(vertices=np.array([0.0, 1.0, 2.0, 3.0, 4.0]))
my_mat = F.Material(D_0=3, E_D=2, K_S_0=1, E_K_S=0, name="mat")

s0 = F.VolumeSubdomain1D(1, borders=[0.0, 2], material=my_mat)
s1 = F.VolumeSubdomain1D(2, borders=[2, 4], material=my_mat)
l0 = F.SurfaceSubdomain1D(1, x=0.0)
l1 = F.SurfaceSubdomain1D(2, x=4.0)
my_model.interfaces = [F.Interface(6, (s0, s1))]

my_model.temperature = 55
my_model.subdomains = [s0, s1, l0, l1]
my_model.surface_to_volume = {l0: s0, l1: s1}

my_model.species = [
    F.Species("H", subdomains=[s0]),
    F.Species("T", subdomains=[s0, s1], mobile=False),
]

filename = str(tmpdir.join("my_export.txt"))
my_export = F.VTXSpeciesExport(
    filename, field=my_model.species, subdomain=s0)
assert my_export.filename.suffix == ".bp"
my_model.exports = [my_export]
my_model.settings = F.Settings(atol=1, rtol=0.1)
my_model.settings.stepsize = F.Stepsize(initial_value=1)

my_model.initialise()
```

PR #1030, opened 2025-09-17
Fix penalty term

RemDelaporteMathurin:
## Proposed changes

Closes #1026 

## Types of changes

What types of changes does your code introduce to FESTIM?

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code refactoring
- [ ] Documentation Update (if none of the other choices apply)
- [ ] New tests

## Checklist

- [ ] Black formatted
- [ ] Unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

Issue #1026, opened 2025-08-18
Penalty-approach for Sievert/Henry interfaces.


jorgensd:
The current implementation of the Sievert/Henry interface condition with quadratic penalty assumes that we modify the boundary conditions, rather than adding a quadratic penalty.
Following is an MWE, using DOLFINx nightly, that illustrates why we should use quadratic penalty as apposed to the modified boundary condition.

The manufactured solution of this problem is:
Given a unit square, divided into two materials $\Omega_A=[0,d]\times[0,1]$ and $\Omega_B=[d, 1]x[0,1]$
with 

$$
\begin{align*}
    -\nabla \cdot (D_A\nabla u_A)&=f_A&&\text{in } \Omega_A,\\
    -\nabla \cdot (D_B\nabla u_B)&=f_B&&\text{in } \Omega_B,\\
    u_A &= K_A\left(\frac{u_B}{K_B}\right)^n&&\text{on } \Gamma,\\
    D_A\nabla u_A \cdot n_A &=  D_B\nabla u_B\cdot n_A && \text{on }\Gamma
\end{align*}
$$

where $n=\frac{1}{2}$ or $1$, where 0.5 correponds to $\Omega_A$ is Sievert, $\Omega_B$ Henry.

$$
\begin{align*}
u_A&=x\\
u_B&=\frac{n}{d^{\frac{1}{n}-1}}x^{\frac{1}{x^n}}\\
K_A&= 1\\
K_B&= n d^{\frac{n-1}{n}}\\
D_A&=D_B=1
\end{align*}
$$

```python
from mpi4py import MPI
import dolfinx.fem.petsc
import ufl
import numpy as np
import numpy.typing as npt
from enum import Enum
import argparse


class PenaltyMode(Enum):
    BC = 1
    QUADRATIC = 2


def tr_a(v):
    return v("+")


def tr_b(v):
    return v("-")


def left_dofs(x):
    return np.isclose(x[0], 0.0)


def right_dofs(x):
    return np.isclose(x[0], 1.0)


def compute_interface_data(
    cell_tags: dolfinx.mesh.MeshTags, facet_indices: npt.NDArray[np.int32]
) -> npt.NDArray[np.int32]:
    """
    Compute interior facet integrals that are consistently ordered according to the `cell_tags`,
    such that the data `(cell0, facet_idx0, cell1, facet_idx1)` is ordered such that
[...]

PR #1233, opened 2026-08-18
Codim: Fix multiple codim subdomains erroring on a single bulk subdomain

ee-nn:
### Summary
See https://github.com/festim-dev/FESTIM/pull/1216#discussion_r3805332451. Note that this implementation replaces `coupling_measure` and `unindexed_coupling_measure` with a single `facet_measure` method, this means that #1231 will need to be rebased. 

We could keep both methods if rebasing is too much of a headache but they ended up being rather redundant 

### Motivation and Context
Ensure that multiple codimensional couplings can be applied across each subdomain. 

UDPATE: I believe this also applies the fix suggested for the issue in #897 

## Type of Change

- [X] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation

### DIFF
--- a/test/system_tests/test_codim1_interface.py
+++ b/test/system_tests/test_codim1_interface.py
@@ -164,7 +164,7 @@ def test_interior_manifold_conserves_particles():
     c_r = H_r.subdomain_to_solution[right]
     c_g = H_g.subdomain_to_solution[gamma]
     entity_maps = [sd.cell_map for sd in model.volume_subdomains]
-    dS = model.coupling_measure(gamma)
+    dS = model.facet_measure(gamma)(gamma.id)
 
     def assemble(form):
         return model.mesh.mesh.comm.allreduce(
@@ -246,7 +246,7 @@ def test_interior_manifold_inside_a_single_subdomain():
     model.initialise()
 
     assert model.manifold_is_interior(gamma)
-    assert model.coupling_measure(gamma).integral_type() == "interior_facet"
+    assert model.facet_measure(gamma)(gamma.id).integral_type() == "interior_facet"
 
     model.run()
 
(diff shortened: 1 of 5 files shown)


RemDelaporteMathurin (approved):
Thanks for the fix @ee-nn and for the clear comments!

test/system_tests/test_codim1_interface.py
```diff
@@ -515,3 +515,194 @@ def exact(sub, a, b):
...
+        F.ParticleFluxBC(
+            subdomain=gamma,
+            species=bulk_species,
+            value=lambda c_g, c_b: k * (c_g - c_b),
+            species_dependent_value=dependencies,
+        ),
+    )
+
+
+def build_two_junctions(n=12, second_junction="manifold"):
```
RemDelaporteMathurin, line 555: It would be nice to use `second_junction` with something else than manifold to test that we can mix `Interface` and codim1 coupling

ee-nn: Thanks for catching this! This points to a deeper issue actually; trying to run the following MWE doesn't work: 

```python

from mpi4py import MPI

import dolfinx.mesh
import numpy as np

import festim as F

# Parameters
L = 4.0
x_int_left = L / 3
x_int_right = 2 * L / 3
D_left = 2.0
D_middle, K_middle = 1.5, 1.5  # these need solubilities for the interface formulation
D_right, K_right = 1.0, 1.0
D_int = 0.5
# four exchanges -> four (adsorption, desorption) pairs, not three
k1, k2 = 4.0, 1.0  # left   <-> gamma_left
k3, k4 = 3.0, 1.5  # middle <-> gamma_left
c_int_max = 1.0

# 24 cells over [0, 4] puts nodes exactly on L/3 and 2L/3. With 20 the interfaces fall
# between nodes, the facet meshtags come out empty, and the manifolds are rejected for
# being adjacent to no volume subdomain.
mesh = dolfinx.mesh.create_rectangle(
    MPI.COMM_WORLD,
    [np.array([0.0, 0.0]), np.array([L, 1.0])],
    [24, 10],
    cell_type=dolfinx.mesh.CellType.quadrilateral,
)

# Create bulk subdomainds
eps = 1e-10
left = F.VolumeSubdomain(
    id=1,
    material=F.Material(D_0=D_left, E_D=0.0),
    locator=lambda x: x[0] <= x_int_left + eps,
    name="left",
)
middle = F.VolumeSubdomain(
    id=2,
    material=F.Material(D_0=D_middle, E_D=0.0, K_S_0=K_middle, E_K_S=0),
    locator=lambda x: (x[0] >= x_int_left - eps) & (x[0] <= x_int_right + eps),
    name="middle",
)
right = F.VolumeSubdomain(
    id=3,
    material=F.Material(D_0=D_right, E_D=0.0, K_S_0=K_right, E_K_S=0),
    locator=lambda x: x[0] >= x_int_right - eps,
    name="right",
)

# create codim subdomains
gamma_left = F.VolumeSubdomain(
    id=4,
    material=F.Material(D_0=D_int, E_D=0.0),
    dim=mesh.topology.dim - 1,
    locator=lambda x: np.isclose(x[0], x_int_left),
    name="interface left",
)

# This is now an Interface instead of a codim subdomain
gamma_right = F.Interface(id=5, subdomains=[middle, right], penalty_term=10)

# left and right edges
left_boundary = F.SurfaceSubdomain(id=6, locator=lambda x: np.isclose(x[0], 0.0))
right_boundary = F.SurfaceSubdomain(id=7, locator=lambda x: np.isclose(x[0], L))

H_left = F.Species("c_left", subdomains=[left])
H_middle_and_right = F.Species("c_middle", subdomains=[middle, right])
# H_right = F.Species("c_right", subdomains=[right])
H_int_left = F.Species("c_int_left", subdomains=[gamma_left])

# Flux from the external left boundary into the left bulk.
left_boundary_flux = F.ParticleFluxBC(
    subdomain=left_boundary,
    species=H_left,
    value=0.5,
)

# Interface coupling: bulk fluxes into the codim-1 interfaces, plus matching sources
# in the interface equations. J > 0 moves particles from the bulk into the manifold.
# The middle bulk appears twice, once per face.
J_left = lambda c_int_left, c_left: (
    k1 * c_left * (1.0 - c_int_left / c_int_max) - k2 * c_int_left
)
J_middle_left = lambda c_int_left, c_middle_and_right: (
    k3 * c_middle_and_right * (1.0 - c_int_left / c_int_max) - k4 * c_int_left
)


# One source per (manifold, side): an interior manifold is integrated over interior
# facets, where each term is restricted to one side, so a single source may not read
# both bulk concentrations at once. The bulk species named in
# species_dependent_value is what tells FESTIM which side the term belongs to.
interface_sources = [
    F.ParticleSource(
        value=J_left,
        species=H_int_left,
        volume=gamma_left,
        species_dependent_value={"c_int_left": H_int_left, "c_left": H_left},
    ),
    F.ParticleSource(
        value=J_middle_left,
        species=H_int_left,
        volume=gamma_left,
        species_dependent_value={
            "c_int_left": H_int_left,
            "c_middle_and_right": H_middle_and_right,
        },
    ),
]

interface_fluxes = [
    F.ParticleFluxBC(
        subdomain=gamma_left,
        species=H_left,
        value=lambda c_int_left, c_left: -J_left(c_int_left, c_left),
        species_dependent_value={"c_int_left": H_int_left, "c_left": H_left},
    ),
    F.ParticleFluxBC(
        subdomain=gamma_left,
        species=H_middle_and_right,
        value=lambda c_int_left, c_middle_and_right: (
            -J_middle_left(c_int_left, c_middle_and_right)
        ),
        species_dependent_value={
            "c_int_left": H_int_left,
            "c_middle_and_right": H_middle_and_right,
        },
    ),
]

right_dirichlet = F.FixedConcentrationBC(
    subdomain=right_boundary,
    value=0.0,
    species=H_middle_and_right,
)

model = F.HydrogenTransportProblemDiscontinuous(
    mesh=F.Mesh(mesh),
    species=[H_left, H_middle_and_right, H_int_left],
    subdomains=[
        left,
        middle,
        right,
        gamma_left,
        left_boundary,
        right_boundary,
    ],
    sources=[*interface_sources],
    boundary_conditions=[left_boundary_flux, *interface_fluxes, right_dirichlet],
    interfaces=[gamma_right],
    temperature=500,
    settings=F.Settings(atol=1e-10, rtol=1e-10, transient=False),
)

model.initialise()
model.run()



```

Note that we are setting the same species on both volume subdomains (`middle` and `right`) that abut the `Interface` since that's what the interface formulation needs. However, it runs into the `ValueError: species c_middle is not defined on volume subdomain 4 and has a solution on several other subdomains ([2, 3]), so it is ambiguous which one this expression should use.` This is coming from the `solution_on` function: 

```python

solutions = species.subdomain_to_solution
if subdomain in solutions:
    return solutions[subdomain]
if len(solutions) == 1:
    return next(iter(solutions.values()))
others = [s.id for s in solutions]
raise ValueError(
    f"species {species.name} is not defined on volume subdomain {subdomain.id} and "
    f"has a solution on several other subdomains ({others}), so it is ambiguous "
    "which one this expression should use."
)

```

Since `H_middle_and_right` is defined on two different subdomains, this conditional block raises the `ValueError`. However, the new `foreign_species` and `coupling_side` methods should already remove this ambiguity, so their capabilities should now be passed through `solution_on`. 

The other part of this is that the current `Interface` formulation needs *all* species to be on both sides of the interface. So, even though `H_left` never touches the interface region, even with the above fix to `solution_on` it will still fail. So, another helper function is added to help with this.

ee-nn:
Hmm, still a few kinks to iron out, will take another look a little later today

ee-nn:
Ok, so I think that the interface + codim subdomain both adjacent to a single bulk volume should be correct now, but one of the new tests added, `test_interface_and_interior_manifold_bounding_one_subdomain()`, kept failing. I spent a while trying to figure out why, and I don't think that it has to do with anything on the codim side of things. I will create a separate issue tomorrow

test/system_tests/test_codim1_interface.py
RemDelaporteMathurin: I think here the issue is that `Interface` only allows having the same species across the interface.

We should allow `cA / K_SA = cB / K_SB` for instance... and it should be easy to implement

ee-nn: I agree that the issue stems from Interface only allowing the same species across the interface. For now I have it set up such that only species that are assigned to both sides of the interface will have chemical potential continuity enforced on them (I think this should fix #897?)

I'm a bit confused by why `cA / K_SA = cB / K_SB` would work. Is this assuming that A and B are the exact same species (eg. protium), just with different labels based on what domain they are in? If so, I agree, but we'd have to be careful about not accidentally setting the chemical potential of two arbitrary species equal to each other, no?

RemDelaporteMathurin: consider having H on one side and H2 on the other, you would want to set `cH ^2 / K_S^2 = cH2 /K_H` the interface class currently doesn't allow this

ee-nn: I see. Was this already worked on in #1026 and #1030? It looks a bit dated but seems like that could help

RemDelaporteMathurin: I don't think so. Plus I think these PRs are a bit stale

ee-nn:
> Ok, so I think that the interface + codim subdomain both adjacent to a single bulk volume should be correct now, but one of the new tests added, `test_interface_and_interior_manifold_bounding_one_subdomain()`, kept failing. I spent a while trying to figure out why, and I don't think that it has to do with anything on the codim side of things. I will create a separate issue tomorrow

Ok, after some thought issue is posted in detail at #1236

PR #1242, opened 2026-08-21
Io4dolfinx export formats

### LINKED ISSUE #1241: Extend I/O in FESTIM by relying more on `io4dolfinx`
Write now, FESTIM's exports are limited to VTX but if we were to rely more on `io4dolfinx` we could theoretically support any format (eg. VTKHDF)

RemDelaporteMathurin:
## What this does

FESTIM's field exports were effectively VTX-only: `VTXSpeciesExport`, `VTXTemperatureExport` and `CustomFieldExport` all wrote `.bp` through `dolfinx.io.VTXWriter`, and `XDMFExport` was a separate one-off. io4dolfinx was already a hard dependency but only used in two hidden places — the `checkpoint=True` flag and `read_function_from_file`.

This makes the output format a first-class choice and leans on io4dolfinx for it.

```python
F.SpeciesExport("results.vtkhdf", field=[H], subdomain=vol, format="vtkhdf")
F.SpeciesExport("state.bp",       field=[H], subdomain=vol, format="checkpoint")
```

## Why it needed a refactor first

Two things blocked simply adding formats:

1. **`ExportBaseClass` was a pure data holder** with no `write()`/`close()`. Every writer's create/write/close lifecycle was duplicated as `isinstance` branches across the problem classes, so each new format multiplied branches.
2. **The class names hard-coded the format.** `VTXSpeciesExport(checkpoint=True)` didn't write VTX at all — it wrote an ADIOS2 checkpoint.

So the export objects now own their writer:

- **`festim/exports/writers.py`** (new) — a `FieldWriter` strategy with four implementations. VTX and XDMF wrap `dolfinx.io`; vtkhdf and checkpoint go through io4dolfinx. Adding a format is now one small class.
- **`festim/exports/vtx.py` → `festim/exports/field.py`** — `ExportBaseClass` gains `define_writer()` / `update()` / `write()` / `close()` and a format registry. `vtx.py` remains as an import shim.
- **Problem classes** resolve `_export_functions(export) → (functions, names, mesh)` and hand it to `export.define_writer(...)`; `post_processing` collapses to `export.update(); export.write(t)`.

## New capability

| `format` | Extension | ParaView | Notes |
|---|---|---|---|
| `"vtx"` | `.bp` | yes | The default, unchanged |
| `"vtkhdf"` | `.vtkhdf` | yes | **New.** One HDF5 file instead of a directory tree |
| `"xdmf"` | `.xdmf` | yes | Now goes through the same path as the rest |
| `"checkpoint"` | `.bp` / `.h5` | no | Now available for **all** field types, with a `backend` choice |

The headline win is `vtkhdf`: it is a `MultiBlockDataSet`, so exports pointing at the same filename become named blocks of **one file**. A multi-material model no longer scatters one `.bp` directory per subdomain:

```python
my_model.exports = [
    F.SpeciesExport("results.vtkhdf", field=[H], subdomain=vol1, format="vtkhdf"),
    F.SpeciesExport("results.vtkhdf", field=[H], subdomain=vol2, format="vtkhdf"),
]
```

Checkpointing also stops being species-only — `TemperatureExport` and `CustomFieldExport` can be checkpointed now.

## Backwards compatibility

`VTXSpeciesExport`, `VTXTemperatureExport` and `XDMFExport` all still work, as thin subclasses pinning their format. Only `checkpoint=True` warns (it maps to `format="checkpoint"`). Checkpoint dataset **names are unchanged**, so existing `read_function_from_file(name=...)` calls keep working.

Three intentional behaviour changes worth noting:

- **`CustomFieldExport(checkpoint=True)` starts working.** It was accepted and stored but never read by any problem class, so it silently produced a plain VTX file.
- **`XDMFExport` works in `HydrogenTransportProblemDiscontinuous`**, where it previously hit a `NotImplementedError`.
- **`XDMFExport` gained `times=`** — it never forwarded it, so it was always `None`.

Also fixed along the way: `export.times` → stepsize milestones was only registered in `HydrogenTransportProblem`, so in the discontinuous class `times` could silently miss timesteps.

## Fixing the test suite under MPI

While verifying the parallel path I found the suite **hangs indefinitely under `mpirun`** — and it reproduces on a clean `main`, so it predates this PR.

Cause: pytest gives each rank its own temp directory (`pytest-17` on one rank, `pytest-18` on another) while the ADIOS2/HDF5 writes underneath are collective across `COMM_WORLD`. Every rank blocks forever on a file the others never touch.

`test/conftest.py` (new) overrides `tmp_path`/`tmpdir` to broadcast rank 0's path. No test needed changing. The test that hung past 500 s now passes in **1.57 s**, and `test_io.py` passes on 1, 2 and 3 ranks.

## Packaging

The vtkhdf backend needs `h5py`, which io4dolfinx only pulls in as an extra — so the dependency becomes `io4dolfinx[h5py]`. Writing from more than one MPI process additionally needs the MPI build, pinned as `h5py=*=mpi_*` in the conda recipe, docs env and conda CI. FESTIM raises an actionable error rather than letting io4dolfinx fail with a bare `ValueError` from inside a context manager.

## Testing

Full suite green: **1010 passed, 5 skipped** (dolfinx 0.11.0, io4dolfinx 1.2.0). `test_io.py` additionally verified on 2 and 3 MPI ranks.

- `test/unit/test_exports/test_field_export.py` (new) — format validation, extension coercion, and the backwards-compatibility surface including the `checkpoint=` deprecation.
- `test/system_tests/test_io.py` — vtkhdf round trip, vtkhdf multiblock (two subdomains in one file), and a temperature checkpoint round trip.

Note for anyone adding parallel tests here: assert on **every** rank. An assertion inside `if rank == 0` turns a failure into a deadlock, because rank 0 raises while the others walk into the next collective. The new tests broadcast their verdict via `_bcast_from_root`.

## Not included

A `TemperatureExport` with a *constant* temperature writes a data-less file — the field is never written because `temperature_time_dependent` is `False`. That is pre-existing behaviour, preserved here deliberately rather than changed silently, but it looks like a bug worth a follow-up.


src/festim/hydrogen_transport_problem.py
```diff
@@ -55,6 +53,12 @@
 ]
 
 
+def _block_name(export) -> str:
+    """Name of the block an export writes into, for multi-mesh formats like vtkhdf."""
+    subdomain = getattr(export, "subdomain", None)
+    return "mesh" if subdomain is None else f"subdomain_{subdomain.id}"
+
```
RemDelaporteMathurin: feel like this could live somewhere else?

src/festim/exports/field.py
```diff
@@ -0,0 +1,631 @@
...
+        checkpoint: bool = False,
+        times: list[float] | list[int] | None = None,
+    ):
+        super().__init__(
+            filename,
+            field,
+            subdomain=subdomain,
+            format=_resolve_checkpoint_kwarg("vtx", checkpoint),
+            times=times,
+        )
```
jhdark, line 638: Let's not bother having a deprecation warning for the checkpoint bool itself but for the whole class since we want to remove it

src/festim/exports/field.py
```diff
@@ -0,0 +1,631 @@
...
+        return format
+    warnings.warn(
+        "`checkpoint=True` is deprecated, use `format='checkpoint'` instead.",
+        DeprecationWarning,
+        stacklevel=3,
+    )
+    return "checkpoint"
+
+
+class ExportBaseClass:
```
jhdark: Maybe we could call this `FieldExportBase` instead to differenciate from derived quantties?

src/festim/exports/field.py
```diff
@@ -0,0 +1,631 @@
...
+        self.field = field
+        self.subdomain = subdomain
+
+    @property
+    def subdomain(self) -> VolumeSubdomain | None:
+        return self._subdomain
+
+    @subdomain.setter
+    def subdomain(self, value: VolumeSubdomain | None):
+        self._subdomain = value
```
jhdark: Why do we need a setter/getter here? it wasn't there before

src/festim/exports/vtx.py
jhdark, line 1: Maybe we can just remove this

src/festim/exports/xdmf.py
```diff
@@ -1,83 +1,31 @@
...
-            )
-            self._mesh_written = True
-
-        for field in self.field:
-            self._writer.write_function(field.post_processing_solution, t)
-
-    def __del__(self):
-        if self._writer is not None:
-            self._writer.close()
+    def __init__(
```
jhdark, line 27: It makes no sense to maintain this if we have SpeciesExport with format=xdmf so we should add a DeprecationWarning here

src/festim/heat_transfer_problem.py
```diff
@@ -217,22 +216,38 @@ def create_formulation(self):
...
+
+        Returns:
+            the functions to write, the name to store each one under, and the mesh
+            they live on
+        """
+        if isinstance(export, exports.TemperatureExport):
+            return [self.u], [self.u.name], self.u.function_space.mesh
+        raise NotImplementedError(
+            f"{type(export).__name__} is not implemented for heat transfer problems"
+        )
```
jhdark: I don't like that we have a separate function just for this (it's only used once) it could be directly done in initialise_exports

jhdark: We should also be able to support the `CustomFieldExport` in the `HeatTransferProblem` class

src/festim/hydrogen_transport_problem.py
```diff
@@ -428,75 +432,50 @@ def define_temperature(self):
                 )
                 self.temperature_fenics.interpolate(self.temperature_expr)
 
+    def _export_functions(self, export):
```
jhdark:
```suggestion
    def _export_functions(self, export) -> list[fem.function], list[str], dolfinx.mesh.Mesh :
```
or something like this to be clearer on what this function returns

src/festim/hydrogen_transport_problem.py
jhdark: If it's for multi-mesh formats then it only makes sense for the Discontinuous class and the block name can just be `mesh` for the HydrogenTransportProblem class

src/festim/hydrogen_transport_problem.py
jhdark: Maybe we should call this "export_context" instead since it returns more than just the functions

src/festim/problem.py
```diff
@@ -81,6 +81,28 @@ def dt(self):
...
+        Without this the adaptive stepsize can step over the requested times and the
+        export silently never fires.
+
+        Args:
+            export: the export whose `times` should be hit exactly
+        """
+        if not getattr(export, "times", None):
+            return
+        if self.settings is None or self.settings.stepsize is None:
+            return
```
jhdark: Why do we need this? a problem _needs_ settings so if the settings haven't been defined at this stage we should raise instead of simply returning none

src/festim/hydrogen_transport_problem.py
```diff
@@ -2073,50 +2052,68 @@ def create_flux_values_fenics(self):
                     t=self.t,
                 )
 
+    def _export_functions(self, export):
```
jhdark: same comment as above, this should be renamed to something like "export_context"

test/conftest.py
jhdark, line 1: It may be worth how io4dolfinx works around this:

https://github.com/scientificcomputing/io4dolfinx/blob/main/tests/conftest.py

jhdark (changes requested):

src/festim/exports/field.py
RemDelaporteMathurin: Renamed to `FieldExportBase`. `ExportBaseClass` is kept as an alias at the bottom of `field.py` since it was already public API on main, but everything internal now uses the new name.

src/festim/exports/field.py
RemDelaporteMathurin: Good catch, it was a leftover. Removed, `subdomain` is a plain attribute again.

src/festim/exports/vtx.py
RemDelaporteMathurin: Removed `src/festim/exports/vtx.py` (and the test that asserted the shim was importable).

src/festim/exports/xdmf.py
RemDelaporteMathurin: Added a `DeprecationWarning` to `XDMFExport.__init__` pointing at `SpeciesExport(..., format="xdmf")`, plus a `.. deprecated::` note in the docstring and a table of deprecated classes in the user guide.

src/festim/exports/field.py
RemDelaporteMathurin: Done: `VTXSpeciesExport` and `VTXTemperatureExport` now warn about the class itself and `_resolve_checkpoint_kwarg` is no longer called there. It's still used by `CustomFieldExport`/`ReactionRateExport`, which aren't going away, so the `checkpoint` bool still warns in those.

src/festim/heat_transfer_problem.py
RemDelaporteMathurin: Inlined into `initialise_exports`, the helper is gone.

RemDelaporteMathurin: Added. `CustomFieldExport` now gets a function on the problem's own function space and `set_dolfinx_expression(temperature=self.u, time=self.t)`. Two tests in `test_heat_transfer_problem.py`: one checks the exported field equals `2*T`, one checks anything else (e.g. `SpeciesExport`) still raises `NotImplementedError`.

src/festim/hydrogen_transport_problem.py
RemDelaporteMathurin: Annotated as `-> tuple[list[fem.Function], list[str], dolfinx.mesh.Mesh]` in both classes.

RemDelaporteMathurin: Renamed to `_export_context` (kept the leading underscore since it's internal).

src/festim/hydrogen_transport_problem.py
RemDelaporteMathurin: Renamed here too.

src/festim/hydrogen_transport_problem.py
RemDelaporteMathurin: Agreed. `_block_name` is gone: `HydrogenTransportProblem` no longer passes `block_name` at all so it takes the `"mesh"` default, and the two-line subdomain naming now lives inline in `HydrogenTransportProblemDiscontinuous.initialise_exports` where it's the only place several meshes can share a file.

src/festim/problem.py
RemDelaporteMathurin: Changed to raise: `AttributeError` if `settings` is missing, and `ValueError` if the problem is steady state (no stepsize, so `times` is meaningless there). Test added in `test_heat_transfer_problem.py`.

test/conftest.py
RemDelaporteMathurin: io4dolfinx does the same `MPI.COMM_WORLD.bcast(tmp_path, root=0)`, just per test inside its `write_function` fixture. I kept it in `conftest.py` so the whole suite is covered without tests opting in, but simplified it to match: broadcasting the path directly and dropping the extra `Barrier` (`bcast` already returns only after rank 0 reached it, which is after pytest created the directory). Docstring now points at their conftest.

PR #1247, opened 2026-08-31
changes to allow for SurfaceFlux exports for spherical and cylindrical coordinates

ey44:
### Summary
This is only a small change but it allows SurfaceFlux to take cylindrical and spherical coordinates. SurfaceFlux for spherical is just atom/m2 while the cylindrical (I think) is per unit length). Currently a NotImplementedError is raised so this support these computed properties for these other coordinate systems. This does not cover partial solid angles, only full. Some support from Claude Sonnet 4.8 was utilised.

### Related Issues
PR #689 and #688 -- these were found after the write up, and it looks like such changes were rejected (maybe this change will also be the same)

### Motivation and Context
Useful for some spherical pebble modelling! I was doing this anyway in postprocessing. It also coudl be useful to extend it to help the GasEnclosure stuff.

## Type of Change

- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation

## Additional Notes

The surface flux export for the cylindrical I think is per unit axial length. a review on my maths would be appreciated! I also do not have a good test for the cylindrical yet.

### DIFF
--- a/src/festim/exports/surface_flux.py
+++ b/src/festim/exports/surface_flux.py
@@ -1,8 +1,11 @@
+import math
+
 import ufl
 from dolfinx import fem
 from scifem import assemble_scalar
 
 from festim.exports.surface_quantity import SurfaceQuantity
+from festim.mesh import CoordinateSystem, Mesh
 from festim.species import Species
 from festim.subdomain.surface_subdomain import SurfaceSubdomain
 
@@ -22,6 +25,7 @@ class SurfaceFlux(SurfaceQuantity):
     field: Species
     surface: SurfaceSubdomain
     filename: str
+    mesh: Mesh
 
     title: str
     value: float
@@ -51,9 +55,31 @@ class SurfaceFlux(SurfaceQuantity):
         mesh = ds.ufl_domain()
         n = ufl.FacetNormal(mesh)
 
+        match self.mesh.coordinate_system:
+            case CoordinateSystem.CARTESIAN:
+                weight = 1
+            case CoordinateSystem.CYLINDRICAL:
+                assert self.mesh.vdim == 1, (
+                    "SurfaceFlux in cylindrical coordinates is only implemented "
+                    "for 1D radial meshes"
+                )
+                r = ufl.SpatialCoordinate(mesh)[0]
+                weight = 2 * math.pi * r
+            case CoordinateSystem.SPHERICAL:
+                assert self.mesh.vdim == 1, (
+                    "SurfaceFlux in spherical coordinates is only implemented "
+                    "for 1D radial meshes"
+                )
+                r = ufl.SpatialCoordinate(mesh)[0]
+                weight = 4 * math.pi * r**2
+            case _:
+                raise NotImplementedError(
+                    f"Unknown coordinate system {self.mesh.coordinate_system!s}"
+                )
+
         self.value = assemble_scalar(
             fem.form(
-                -self.D * ufl.dot(ufl.grad(u), n) * ds(self.surface.id),
+                -weight * self.D * ufl.dot(ufl.grad(u), n) * ds(self.surface.id),
                 entity_maps=entity_maps,
             )
         )
--- a/src/festim/exports/surface_flux.py
+++ b/src/festim/exports/surface_flux.py
@@ -5,8 +5,8 @@ from dolfinx import fem
 from scifem import assemble_scalar
 
 from festim.exports.surface_quantity import SurfaceQuantity
-from festim.mesh import CoordinateSystem
 from festim.helpers import restrict
+from festim.mesh import CoordinateSystem
 from festim.species import Species
 from festim.subdomain.surface_subdomain import SurfaceSubdomain
 from festim.subdomain.volume_subdomain import VolumeSubdomain
@@ -100,12 +100,16 @@ class SurfaceFlux(SurfaceQuantity):
                     f"Unknown coordinate system {self.coordinate_system!s}"
                 )
 
-        integrand = -weight * self.D * ufl.dot(ufl.grad(u), n)
+        integrand = -self.D * ufl.dot(ufl.grad(u), n)
         if subdomain_id is None:
             subdomain_id = self.surface.id
 
         if self.drift_velocity is not None:
             integrand += u * ufl.dot(self.drift_velocity, n)
+
+        # the weight carries the metric of the coordinate system, so it multiplies the
+        # whole flux density, drift included
+        integrand *= weight
 
         self.value = assemble_scalar(
             fem.form(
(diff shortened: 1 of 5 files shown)


ey44:
I read these wrong PR https://github.com/festim-dev/FESTIM/pull/689 and https://github.com/festim-dev/FESTIM/pull/688 -- so ignore my message about that - I shoudl probably add the partial coverage stuff

src/festim/exports/surface_flux.py
```diff
@@ -51,9 +55,31 @@ def compute(
...
+                assert self.mesh.vdim == 1, (
+                    "SurfaceFlux in spherical coordinates is only implemented "
+                    "for 1D radial meshes"
+                )
+                r = ufl.SpatialCoordinate(mesh)[0]
+                weight = 4 * math.pi * r**2
+            case _:
+                raise NotImplementedError(
+                    f"Unknown coordinate system {self.mesh.coordinate_system!s}"
+                )
```
jhdark, line 101: might not need this

RemDelaporteMathurin, line 101: Is this covered somewhere else?

src/festim/exports/surface_flux.py
```diff
@@ -51,9 +55,31 @@ def compute(
...
         n = ufl.FacetNormal(mesh)
 
+        match self.mesh.coordinate_system:
+            case CoordinateSystem.CARTESIAN:
+                weight = 1
+            case CoordinateSystem.CYLINDRICAL:
+                assert self.mesh.vdim == 1, (
+                    "SurfaceFlux in cylindrical coordinates is only implemented "
+                    "for 1D radial meshes"
+                )
```
RemDelaporteMathurin: Similar comment, isn't this covered elsewhere already?

src/festim/hydrogen_transport_problem.py
```diff
@@ -536,6 +540,7 @@ def initialise_exports(self):
                 # add the global D to the export
                 export.D = self._species_to_D_global.get(export.field)
                 export.D_expr = self._species_to_D_global_expr.get(export.field)
+                export.mesh = self.mesh
```
RemDelaporteMathurin: We don't need to do this if we pass the coordinate system as an attribute of the export.

```suggestion
                export.coordinate_system= self.mesh.coordinate_system
```

src/festim/exports/surface_flux.py
```diff
@@ -51,9 +55,31 @@ def compute(
         mesh = ds.ufl_domain()
         n = ufl.FacetNormal(mesh)
 
+        match self.mesh.coordinate_system:
```
RemDelaporteMathurin: This is confusing because now we have a fenics mesh (`mesh`) and a FESTIM mesh `self.mesh`. 

It would be more straightforward to add `coordinate_system` as an attribute of the export instead

src/festim/hydrogen_transport_problem.py
```diff
@@ -2130,6 +2135,7 @@ def initialise_exports(self):
                 # NOTE: maybe we need to make sure there are no functionspace clashes?
 
                 export.D = D
+                export.mesh = self.mesh
```
RemDelaporteMathurin: Same comment as above

test/system_tests/test_spherical_coordinates.py
```diff
@@ -66,6 +68,67 @@ def u_exact(x):
...
+    )
+
+    my_sim.initialise()
+    my_sim.run()
+
+    A = (c1 - c2) * r1 * r2 / (r2 - r1)
+    expected_flux_magnitude = 4 * math.pi * D * A
+
+    assert np.isclose(flux_left.value, -expected_flux_magnitude, rtol=1e-2)
+    assert np.isclose(flux_right.value, expected_flux_magnitude, rtol=1e-2)
```
RemDelaporteMathurin, line 129: Should we check for an error convergence too?

src/festim/exports/surface_flux.py
```diff
@@ -51,9 +55,31 @@ def compute(
...
+                )
+                r = ufl.SpatialCoordinate(mesh)[0]
+                weight = 2 * math.pi * r
+            case CoordinateSystem.SPHERICAL:
+                assert self.mesh.vdim == 1, (
+                    "SurfaceFlux in spherical coordinates is only implemented "
+                    "for 1D radial meshes"
+                )
+                r = ufl.SpatialCoordinate(mesh)[0]
+                weight = 4 * math.pi * r**2
```
RemDelaporteMathurin:
```suggestion
                coverage = 4 * math.pi  # steradians
                weight = coverage * r**2
```

I would factor the coverage out for now so that we can add a TODO later to expose this parameter

src/festim/exports/surface_flux.py
```diff
@@ -51,9 +55,31 @@ def compute(
...
+        match self.mesh.coordinate_system:
+            case CoordinateSystem.CARTESIAN:
+                weight = 1
+            case CoordinateSystem.CYLINDRICAL:
+                assert self.mesh.vdim == 1, (
+                    "SurfaceFlux in cylindrical coordinates is only implemented "
+                    "for 1D radial meshes"
+                )
+                r = ufl.SpatialCoordinate(mesh)[0]
+                weight = 2 * math.pi * r
```
RemDelaporteMathurin:
```suggestion
                coverage =  2 * math.pi
                weight = coverage * r
```

test/system_tests/test_spherical_coordinates.py
```diff
@@ -66,6 +68,67 @@ def u_exact(x):
     assert L2_error < 1e-6
 
 
+def test_surface_flux_spherical():
```
RemDelaporteMathurin, line 71: You could write a test for cylindrical + for cylindrical ti would be nice to test in 1D and in 2D also in a case where the surface doesn't align with the axes

test/test_surface_quantity.py
```diff
@@ -43,6 +43,7 @@ def test_surface_flux_export_compute():
         surface=dummy_surface,
     )
     my_export.D = D
+    my_export.mesh = my_mesh
```
RemDelaporteMathurin: see above

RemDelaporteMathurin (changes requested):
Thanks so much @ey44 for this! A few comments

test/system_tests/test_spherical_coordinates.py
ey44: added for 1D, added TODO RE the alignment that is not with aligned with the axis

test/test_surface_quantity.py
ey44: chaged it to follow the above idea

src/festim/exports/surface_flux.py
ey44: done as suggested

RemDelaporteMathurin:
@ey44 I just added a test to check cylindrical coordinates on an oblique surface

RemDelaporteMathurin (approved):
Thanks @ey44 for this first contribution!

PR #1260, opened 2026-09-09
Anisotropic diffusion coefficient

RemDelaporteMathurin:
A homogenised polycrystal, a rolled or columnar microstructure, and any non-cubic lattice conduct differently along different directions, and the property that describes that is a second-rank tensor. ``D_0`` may now be given as a square matrix:

    F.Material(D_0=[[9e-11, 2e-12], [2e-12, 2.8e-11]], E_D=0.1)

``E_D`` stays a scalar -- one activation energy shared by every direction, the prefactor carrying the anisotropy. For per-direction activation energies, which a matrix ``D_0`` deliberately cannot express, ``Material.D`` now also accepts a ``fem.Constant``, a ufl expression or a plain matrix as well as a ``fem.Function``, so a tensor built from the problem's temperature stays temperature dependent.

Four places read D as a scalar and had to be written with the matrix product. Each reduces to the previous expression when D is a scalar:

* ``SurfaceFlux``: ``-dot(D grad(u), n)`` rather than ``-D dot(grad(u), n)``, which for a tensor cannot even be assembled.
* Nitsche: the consistency and symmetry terms take the matrix product, and the penalty scales on the normal conductance ``n.D.n``, which is the conductance the boundary actually sees.
* ``define_D_global`` builds tensor-valued DG spaces when the material is anisotropic.
* ``as_fenics_constant`` builds tensor constants.

``ChangeVar`` and the drift terms already used ``D * grad(...)`` and are unchanged.

Also fixes a latent bug this feature would otherwise trip over: ``if self.D_0 and self.D`` in ``Material.__init__`` raises "truth value of an array is ambiguous" for any array input, as does ``if material.D`` in ``define_D_global``. Both now test against None.

Tests: a manufactured solution ``sin(pi x) sin(pi y)`` on the unit square, run with a *rotated* tensor as well as a diagonal one -- with ``D_xy == 0`` the off-diagonal term of the source vanishes and a formulation that mishandled it would still pass. Second-order convergence for identity, diagonal, rotated and strongly rotated tensors, the Arrhenius factor applied to the tensor, the surface flux of a slab against its analytical value and against the same integral assembled by hand, and weak Dirichlet enforcement against strong.

## Type of Change

- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation

### DIFF
--- a/src/festim/helpers.py
+++ b/src/festim/helpers.py
@@ -14,9 +14,14 @@ if TYPE_CHECKING:
 
 
 def as_fenics_constant(
-    value: float | int | fem.Constant, mesh: dolfinx.mesh.Mesh
+    value: float | int | np.ndarray | list | tuple | fem.Constant,
+    mesh: dolfinx.mesh.Mesh,
 ) -> fem.Constant:
     """Converts a value to a dolfinx.Constant.
+
+    Array-like values become tensor-valued constants, which is how an anisotropic
+    material property is supplied: a ``(dim, dim)`` nested list or array gives a
+    constant that multiplies a gradient as a matrix rather than as a scalar.
 
     Args:
         value: the value to convert
@@ -26,15 +31,22 @@ def as_fenics_constant(
         The converted value
 
     Raises:
-        TypeError: if the value is not a float, an int or a dolfinx.Constant
+        TypeError: if the value is not a float, an int, an array-like or a
+            dolfinx.Constant
     """
+    if isinstance(value, bool):
+        raise TypeError(f"Value must not be a bool, not {type(value)}")
     if isinstance(value, float | int):
         return fem.Constant(mesh, dolfinx.default_scalar_type(float(value)))
     elif isinstance(value, fem.Constant):
         return value
+    elif isinstance(value, np.ndarray | list | tuple):
+        array = np.asarray(value, dtype=dolfinx.default_scalar_type)
+        return fem.Constant(mesh, array)
     else:
         raise TypeError(
-            f"Value must be a float, an int or a dolfinx.Constant, not {type(value)}"
+            "Value must be a float, an int, an array-like or a dolfinx.Constant, "
+            f"not {type(value)}"
         )
 
 
--- a/src/festim/material.py
+++ b/src/festim/material.py
@@ -1,5 +1,6 @@
 from enum import Enum
 
+import numpy as np
 import ufl
 from dolfinx import fem
 
@@ -110,7 +111,7 @@ class Material:
         self.solubility_law = solubility_law
         self.D = D
 
-        if self.D_0 and self.D:
+        if self.D_0 is not None and self.D is not None:
             raise ValueError(
                 "D_0 and D cannot be set at the same time. Please set only one of them."
             )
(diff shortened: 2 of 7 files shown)


src/festim/helpers.py
```diff
@@ -14,10 +14,15 @@
 
 
 def as_fenics_constant(
-    value: float | int | fem.Constant, mesh: dolfinx.mesh.Mesh
+    value: float | int | np.ndarray | list | tuple | fem.Constant,
+    mesh: dolfinx.mesh.Mesh,
 ) -> fem.Constant:
     """Converts a value to a dolfinx.Constant.
 
+    Array-like values become tensor-valued constants, which is how an anisotropic
```
ee-nn, line 22:
```suggestion
    Array-like values become tensor-valued `fem.Constant` objects, which is how an anisotropic
```

src/festim/helpers.py
```diff
@@ -26,15 +31,22 @@ def as_fenics_constant(
...
         return value
+    elif isinstance(value, np.ndarray | list | tuple):
+        array = np.asarray(value, dtype=dolfinx.default_scalar_type)
+        return fem.Constant(mesh, array)
     else:
         raise TypeError(
-            f"Value must be a float, an int or a dolfinx.Constant, not {type(value)}"
+            "Value must be a float, an int, an array-like or a dolfinx.Constant, "
+            f"not {type(value)}"
         )
```
ee-nn, line 50: The first `isinstance` check appears redundant with the `else` statement at the end of the conditional block

The `TypeError` statement itself is also confusing (..._not_ be a bool, _not_ type()...). If kept, it should be changed

src/festim/material.py
```diff
@@ -140,10 +141,50 @@ def D(self):
...
+            value, fem.Function | fem.Constant | ufl.core.expr.Expr
+        ) or self._is_square_matrix(value):
+            # array-like values are held as given and turned into a tensor constant
+            # in get_diffusion_coefficient, which is the first point a mesh is known
             self._D = value
         else:
-            raise TypeError("D must be of type fem.Function")
+            raise TypeError(
+                "D must be a fem.Function, a fem.Constant, a ufl expression or an "
+                f"array-like (for an anisotropic tensor), not {type(value)}"
```
ee-nn, line 153:
```suggestion
                "D must be a fem.Function, fem.Constant, ufl expression or a "
                f"square array-like (for an anisotropic tensor), not {type(value)}"
```

src/festim/material.py
```diff
@@ -140,10 +141,50 @@ def D(self):
...
+        shape = getattr(value, "ufl_shape", None)
+        return bool(shape) and len(shape) == 2
+
+    def is_anisotropic(self, species=None) -> bool:
+        """Whether this material's diffusivity is a tensor rather than a scalar.
+
+        Anything reading ``D`` as a scalar -- a surface flux, a Nitsche penalty --
+        has to ask, because for a tensor the same expression has to be written
+        with the matrix product instead.
+        """
```
ee-nn, line 181: Just because a diffusion coefficient is passed as a tensor doesn't necessarily mean that it's anisotropic, e.g. $D = \left[\matrix{D_0 & 0 \\\ 0 & D_0}\right]$. Perhaps update the method name?

test/system_tests/test_anisotropic_diffusion.py
```diff
@@ -0,0 +1,270 @@
+"""Anisotropic diffusion: a diffusivity given as a tensor rather than a scalar.
+
+A homogenised polycrystal, a rolled or columnar microstructure, or any hcp
+lattice conducts differently along different directions, and the material
+property that describes that is a second-rank tensor. ``D_0`` may therefore be
```
ee-nn, line 5: In practice allowing the energy to vary with direction could be helpful, for example the energy to migrate across a defect could be different from the ideal lattice. 

One could get away with factoring this into `D_0` but only at constant temperature, I believe the workaround to this is to allow the user to pass an array of ufl expressions

test/test_material.py
```diff
@@ -246,3 +254,46 @@ def test_error_raised_when_D_and_D_0_both_None():
         match=r"D_0 and D cannot both be None. Please set one of them.",
     ):
         F.Material()
+
+
+class TestAnisotropicDiffusion:
```
ee-nn, line 259: Is there a reason these are in their own class?

PR #1264, opened 2026-09-12
Add support for non-cartesian manifolds in codim problems

### LINKED ISSUE #1263: Allow non-cartesian meshes with manifolds
This will be useful for doing codim-1 in 3D domains, e.g. pipes and spheres. As noted by @RemDelaporteMathurin the co-dimensional formulation is currently cartesian-only

ee-nn:
## Description

`v2.2-rc.2` released co-dimensional subdomains but only for cartesian coordinates. This adds functionality for cylindrical and cartesian coordinates. 

### Related Issues
Fixes  #1263 

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📝 Documentation update

## Testing

## Code Quality Checklist

## Documentation

## Screenshots/Examples
Here is a small mwe: 

```python

from mpi4py import MPI

import dolfinx
import numpy as np

import festim as F


def run(name, mesh):
    bulk = F.VolumeSubdomain(id=1, material=F.Material(D_0=1, E_D=0))
    surface = F.VolumeSubdomain(  # the co-dim part 
        id=2,
        dim=mesh.vdim - 1,
        locator=lambda x: np.isclose(x[0], 1.0),
        material=F.Material(D_0=0.1, E_D=0),
    )
    c_bulk = F.Species("bulk", subdomains=[bulk])
    c_surface = F.Species("surface", subdomains=[surface])
    dependencies = {"cb": c_bulk, "cs": c_surface}

    # Hydrogen initially in the bulk transfers to the initially empty surface.
    # The pipe starts with an axial gradient so its surface profile is nonuniform.
    model = F.HydrogenTransportProblemDiscontinuous(
        mesh=mesh,
        subdomains=[bulk, surface],
        species=[c_bulk, c_surface],
        temperature=500,
        initial_conditions=[
            F.InitialConcentration(
                value=(lambda x: 1 + x[1]) if name == "pipe" else 1.0,
                species=c_bulk,
                volume=bulk,
            ),
        ],
        boundary_conditions=[
            F.ParticleFluxBC(
                subdomain=surface,
                species=c_bulk,
                value=lambda cb, cs: cs - cb,  # assume k = 1
                species_dependent_value=dependencies,
            ),
        ],
        sources=[
            F.ParticleSource(
                volume=surface,
                species=c_surface,
                value=lambda cb, cs: cb - cs,
                species_dependent_value=dependencies,
            ),
        ],
        # A codim manifold carries its own equation, so use a volume quantity.
        exports=[
            F.AverageVolume(
                field=c_surface,
                volume=surface,
                filename=f"{name}_surface.csv",
            ),
        ],
        settings=F.Settings(
            transient=True,
            stepsize=0.01,
            final_time=0.2,
            atol=1e-10,
            rtol=1e-10,
        ),
    )
    model.initialise()
    model.run()


# Pipe: inner radius 0.5, outer radius 1, length 2; outer wall has dim=1.
pipe = dolfinx.mesh.create_rectangle(
    MPI.COMM_WORLD, [np.array([0.5, 0.0]), np.array([1.0, 2.0])], [8, 24]
)
run("pipe", F.Mesh(pipe, coordinate_system="cylindrical"))

# Ball: radius 1; outer spherical surface has dim=0 in the radial mesh.
ball = dolfinx.mesh.create_interval(MPI.COMM_WORLD, 32, [0.0, 1.0])
run("ball", F.Mesh(ball, coordinate_system="spherical"))

```

### DIFF
--- a/src/festim/advection.py
+++ b/src/festim/advection.py
@@ -141,7 +141,7 @@ class VelocityField(Value):
         v_cg = basix.ufl.element(
             "Lagrange",
             function_space.mesh.topology.cell_name(),
-            1,
+            0 if function_space.mesh.topology.dim == 0 else 1,
             shape=(function_space.mesh.geometry.dim,),
         )
         self.vector_function_space = fem.functionspace(function_space.mesh, v_cg)
--- a/src/festim/helpers.py
+++ b/src/festim/helpers.py
@@ -13,6 +13,30 @@ if TYPE_CHECKING:
     from festim.subdomain.volume_subdomain import VolumeSubdomain
 
 
+def spatial_coordinate(mesh):
+    """Coordinates on an integration mesh, including a zero-dimensional submesh.
+
+    FFCx 0.11 cannot compile SpatialCoordinate on a vertex cell (its coordinate
+    table is identically one). A P0 coefficient stores the exact coordinates at
+    each point, also when a point subdomain contains several disconnected points.
+    """
+    domain = mesh if isinstance(mesh, ufl.Mesh) else mesh.ufl_domain()
+    # UFL exposed this as a method before the release bundled with DOLFINx 0.11.
+    tdim = domain.topological_dimension
+    if callable(tdim):
+        tdim = tdim()
+    if tdim > 0:
+        return ufl.SpatialCoordinate(domain)
+    if isinstance(mesh, ufl.Mesh):
+        mesh = dolfinx.mesh.Mesh(domain.ufl_cargo(), domain)
+    gdim = mesh.geometry.dim
+    V = fem.functionspace(mesh, ("P", 0, (gdim,)))
+    coordinate = fem.Function(V)
+    coordinate.interpolate(lambda x: x[:gdim])
+    coordinate.x.scatter_forward()
+    return coordinate
+
+
 def as_fenics_constant(
     value: float | int | fem.Constant, mesh: dolfinx.mesh.Mesh
 ) -> fem.Constant:
@@ -71,15 +95,12 @@ def as_mapped_function(
     kwargs = {}
     if "t" in arguments:
         kwargs["t"] = t
-    if "x" in arguments:
-        x = ufl.SpatialCoordinate(function_space.mesh)
-        kwargs["x"] = x
     if "T" in arguments:
         kwargs["T"] = temperature
     if "x" in arguments:
         # the spatial coordinate must come from the mesh the integral is assembled
         # over, which is not always the mesh the unknown lives on
-        kwargs["x"] = ufl.SpatialCoordinate(mesh or function_space.mesh)
+        kwargs["x"] = spatial_coordinate(mesh or function_space.mesh)
 
     for name, species in (species_dependent_value or {}).items():
         # only pass the species the callable actually declares, as done above for
--- a/src/festim/hydrogen_transport_problem.py
+++ b/src/festim/hydrogen_transport_problem.py
@@ -1567,7 +1567,7 @@ class HydrogenTransportProblemDiscontinuous(HydrogenTransportProblem):
                 element_CG = basix.ufl.element(
                     basix.ElementFamily.P,
                     subdomain.submesh.basix_cell(),
-                    1,  # could expose?
+                    0 if subdomain.submesh.topology.dim == 0 else 1,
                     basix.LagrangeVariant.equispaced,
                 )
                 V = dolfinx.fem.functionspace(subdomain.submesh, element_CG)
@@ -1718,7 +1718,9 @@ class HydrogenTransportProblemDiscontinuous(HydrogenTransportProblem):
             else:
                 V = condition.species.subdomain_to_function_space[condition.volume]
 
-                if isinstance(self.temperature_fenics, fem.Function):
+                if condition.volume in self.manifold_subdomains or isinstance(
+                    self.temperature_fenics, fem.Function
+                ):
                     temperature = condition.volume.sub_T
                 else:
                     temperature = self.temperature_fenics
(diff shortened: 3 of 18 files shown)


src/festim/mesh/mesh.py
```diff
@@ -14,6 +17,19 @@ class CoordinateSystem(Enum):
...
+
+        Use the integration domain's coordinates, including when it is a manifold
+        submesh. Cylindrical 1D quantities are per unit axial length.
+        """
+        if self == CoordinateSystem.CARTESIAN:
+            return 1
+        r = spatial_coordinate(mesh)[0]
+        if self == CoordinateSystem.CYLINDRICAL:
+            return 2 * math.pi * r
+        return 4 * math.pi * r**2
```
RemDelaporteMathurin, line 31: Good idea to store this in the `CoordinateSystem` class.

src/festim/mesh/mesh_1d.py
```diff
@@ -152,7 +152,7 @@ def check_borders(self, volume_subdomains):
 
     def define_meshtags(self, surface_subdomains, volume_subdomains, interfaces=None):
         # check if all borders are defined
-        self.check_borders(volume_subdomains)
+        self.check_borders([v for v in volume_subdomains if v.codim(self.vdim) == 0])
```
RemDelaporteMathurin, line 155: Good catch. Is there a test catching this bug?

src/festim/advection.py
```diff
@@ -141,7 +141,7 @@ def convert_input_value(
         v_cg = basix.ufl.element(
             "Lagrange",
             function_space.mesh.topology.cell_name(),
-            1,
+            0 if function_space.mesh.topology.dim == 0 else 1,
```
RemDelaporteMathurin, line 144: - advection in 0D?
- I don't even think basix allows ("Lagrange", 0), is this tested?

src/festim/helpers.py
```diff
@@ -13,6 +13,30 @@
...
+    if tdim > 0:
+        return ufl.SpatialCoordinate(domain)
+    if isinstance(mesh, ufl.Mesh):
+        mesh = dolfinx.mesh.Mesh(domain.ufl_cargo(), domain)
+    gdim = mesh.geometry.dim
+    V = fem.functionspace(mesh, ("P", 0, (gdim,)))
+    coordinate = fem.Function(V)
+    coordinate.interpolate(lambda x: x[:gdim])
+    coordinate.x.scatter_forward()
+    return coordinate
```
RemDelaporteMathurin, line 37: So this is only to catch the situation where you have a vertex cell (a point mesh)?

In that case, could we not have a try/except instead?

```python
try:
    return ufl.SpatialCoordinate(domain)
except [INSERT ERROR WITH 0D MESH]:
   tdim = ....
   V = ....
    coordinate = ...
```

src/festim/hydrogen_transport_problem.py
```diff
@@ -1768,6 +1770,9 @@ def define_function_spaces(
                 unique_species.append(species)
         nb_species = len(unique_species)
 
+        # A radial surface is a point in a 1D mesh. Basix only supports P0 there.
+        if subdomain.submesh.topology.dim == 0:
+            element_degree = 0
```
RemDelaporteMathurin, line 1775: Do we agree that this is not necessarily related to non-cartesian manifolds but 0D manifolds in general yes?

src/festim/hydrogen_transport_problem.py
```diff
@@ -2441,6 +2447,9 @@ def create_subdomain_formulation(self, subdomain: _subdomain.VolumeSubdomain):
         for drift_term in self.drift_terms:
             if drift_term.subdomain != subdomain:
                 continue
+            if subdomain.submesh.topology.dim == 0:
+                # A point has no tangential direction in which a species can drift.
+                continue
```
RemDelaporteMathurin, line 2452: Should we raise a warning or maybe even an error here?

ee-nn: Yeah I agree, I think a warning might be good

src/festim/hydrogen_transport_problem.py
ee-nn: Yes that's my understanding as well

src/festim/mesh/mesh_1d.py
ee-nn: I don't believe so, I can add one

src/festim/advection.py
ee-nn: Sorry I glossed over this initially! You're right I don't think this makes sense, perhaps we can replace this with a warning/error, similar to your comment further down. 

I did double-check the syntax and `basix.ufl.element("Lagrange", "point", 0)` does work. "point" should already be what is set from `dolfinx.mesh.create_submesh`.

src/festim/helpers.py
ee-nn: Yes we could do something like that instead, however after a bit of testing I realize that we'd need to wrap `ufl.SpatialCoordinate` with `dolfinx.fem.form()` because it only fails at form compilation

src/festim/mesh/mesh_1d.py
ee-nn: Added a small direct unit test for this just to be safe, although several of the system tests added do fail if this bug is present

src/festim/helpers.py
ee-nn: I left it as-is for now since trying to return `ufl.SpatialCoordinate(domain)` won't directly trip a try/except, but let me know if you think it would still be good to have a try/except block here

ee-nn:
@RemDelaporteMathurin Let me know if this looks okay now, thanks for all the feedback so far! Once it's merged it should make #1271 easier

RemDelaporteMathurin:
@ee-nn i'm traveling today so I won't have time to review it again. @hhy2022  Could you base this other PR off this branch instead?

hhy2022:
Sure! I will base my PR on this branch when I start working on #1271. @ee-nn @RemDelaporteMathurin

PR #1265, opened 2026-09-15
Skip MUMPS-specific test when unavailable

shimwell:
The `test_petsc_options` test explicitly selects MUMPS:

```python
"pc_factor_mat_solver_type": "mumps",
```

but currently assumes every PETSc build includes it. PETSc builds without MUMPS therefore fail before the option-handling behavior can be tested.

This uses `PETSc.Sys().hasExternalPackage("mumps")` to skip that MUMPS-specific test when the external package is unavailable. This is the same capability check used in [FEniCS/dolfinx#4505](https://github.com/FEniCS/dolfinx/pull/4505).

On the existing FESTIM CI PETSc build, which includes MUMPS, the test continues to run unchanged.

## Verification

- `ruff check test/system_tests/test_misc.py`
- `ruff format --check test/system_tests/test_misc.py`
- Focused test against the Fortran-free PETSc 3.26 development wheel with SuperLU_DIST but no MUMPS: cleanly skipped with `MUMPS is required to test MUMPS-specific PETSc options`


shimwell:
The failing DOLFINx nightly job is unrelated to this change. I reproduced one of its failures from an archive of unmodified FESTIM `main` using the same current `dolfinx/dolfinx:nightly` image and CI installation sequence:

```
test_custom_quantity_volume_integration
ValueError: Mesh of function-space is not the same as the one used for the meshtags
```

The nightly matrix currently reports 286 failures and 3 errors across mesh/subdomain APIs, including changed `create_index_sets()` and `compute_integration_domains()` signatures. This PR only changes `test_petsc_options`; lint and benchmark pass, and the focused test cleanly skips against PETSc without MUMPS. The last `main` Docker run was green, so the moving nightly image appears to have advanced since then.

RemDelaporteMathurin (approved):

PR #1266, opened 2026-09-15
Add PyVista to test dependencies

shimwell:
Some FESTIM tests exercise the PyVista backend in io4dolfinx, but the `test` extra does not install PyVista. The existing Docker and Conda workflows install it separately, which masks the missing test dependency for other environments.

Add `pyvista` to the `test` extra so `pip install .[test]` is sufficient to run these tests.

## Verification

In a clean Python 3.12 container with the wheel-based FEniCSx stack:

- `pip install .[test]` installed PyVista without a separate command;
- `test_vtkhdf_export_roundtrip` passed;
- `test_vtkhdf_multiblock_single_file` passed;
- Ruff check and format checks passed.


shimwell:
I think the failing CI is due to the changes in dolfinx/dolfinx:nightly

RemDelaporteMathurin (approved):

